PR #1, opened 2019-04-19
Add ISO week of encounter to incidence model observation view

tsibley:
Time binning is useful for the modeling and viz efforts, so it's nice to
harmonize that early in the data flow rather than separately several
times.  Weeks are the standard time bin used in epidemiology.

Note that the WHO's epi weeks, synonymous with ISO weeks, are slightly
different from CDC's epi weeks, as described most succinctly¹

> The US CDC defines epidemiological week (known as MMWR week) as seven
> days beginning with Sunday and ending with Saturday. The WHO defines
> epidemiological week, based on ISO week, as seven days beginning with
> Monday and ending with Sunday. In either case, the end of the first
> epidemiological week of the year by definition must fall at least four
> days into the year. Most years have 52 epidemiological weeks, but some
> have 53.

It's my understanding that WHO / ISO weeks are used more widely around
the world, and just like we use the metric system in science, it seems
we should use international datetime standards as well.

If we find ourselves wanting CDC epi weeks in the future, we can add
them as well.

¹ https://epiweeks.readthedocs.io

### DIFF
--- a/schema/deploy/shipping/views@2019-04-19.sql
+++ b/schema/deploy/shipping/views@2019-04-19.sql
@@ -14,14 +14,11 @@ begin;
 -- there needs to be a lag between view development and consumers being
 -- updated, copy the view definition into v2 and make changes there.
 
-drop view shipping.incidence_model_observation_v1;
-
-create view shipping.incidence_model_observation_v1 as
+create or replace view shipping.incidence_model_observation_v1 as
 
     select encounter.identifier as encounter,
 
            (encountered at time zone 'US/Pacific')::date as encountered_date,
-           to_char((encountered at time zone 'US/Pacific')::date, 'IYYY-"W"IW') as encountered_week,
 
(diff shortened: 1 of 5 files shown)


tsibley:
This will support https://github.com/seattleflu/incidence-mapper/issues/26.

tsibley:
[ISO week in incidence model output](https://trello.com/c/52fIanAZ/8-iso-week-in-incidence-model-output)

famulare:
Thanks! Similarly, do we have a set of agreed-upon age bins? Something in years like `["0-4","5-9","10-14","15-19","20-25", ...]`.  Or something more epi-appropriate, like I can imagine we'd want to block by school ages like `["0-4","5-10","11-13","14-17","18-24","25-34","35-44", ...]`.  Bins of uniform width are much much nicer for the modeling, so I'd strongly prefer that. But thoughts, @trvrb ?

PR #3, opened 2019-05-01
Add age bins to incidence model view

joverlee521:
@tsibley The way I'm doing the joins for the new age range tables seem a little verbose to me, not sure if there is a better way to do this.


joverlee521:
[Add age bins to incidence modeling view](https://trello.com/c/NnMu50MF/48-add-age-bins-to-incidence-modeling-view)

schema/deploy/shipping/views.sql
```diff
@@ -55,7 +58,9 @@ create view shipping.incidence_model_observation_v1 as
 
       from warehouse.encounter
       join warehouse.individual using (individual_id)
-      join warehouse.site using (site_id),
+      join warehouse.site using (site_id)
+      join shipping.age_range_fine on age_range_fine.age_range @> cast(encounter.details->'age'->>'value' as int)
+      join shipping.age_range_coarse on age_range_coarse.age_range @> cast(encounter.details->'age'->>'value' as int),
```
tsibley: You asked about the verbosity of this join. It's a fair question, and good to consider! Until we properly lift up age out of the encounter details, this is about as concise as it gets.

Note that one reason it's important to make sure that an age can only fall into a single bin (using the exclusion constraint mentioned in another comment) is that otherwise this join may produce multiple age bin rows for each encounter row, which has the effect of seemingly duplicating the encounter. There are other ways to defend against this possibility, but I think the exclusion constraint is the best.

tsibley:
Thanks! This looks good. I appreciate the attention to the commit messages too!

PR #4, opened 2019-05-01
identifier labels: Make labels for existing batches of identifiers

tsibley:
This is useful if you forget to pass --labels to `identifier mint`, you
lose the original PDF, or if the identifier generation succeeds but
label-making fails for some reason.  I found myself in the last scenario
just now, so whipped this up instead of creating a one-off script to do
it.

Eventually the SQL queries in this command may want to be wrapped up
into a separate business logic layer (e.g. similar to
seattleflu.api.datastore vs. seattleflu.db.DatabaseSession).


tsibley:
[Remake identifier labels](https://trello.com/c/hSq8byfi/67-remake-identifier-labels)

PR #5, opened 2019-05-02
Add presence-absence-processor role

kairstenfay:


schema/deploy/roles/presence-absence-processor.sql
```diff
@@ -0,0 +1,31 @@
...
+   to presence_absence_processor;
+
+grant select
+   on receiving.presence_absence
+   to presence_absence_processor;
+
+grant update (processing_log)
+   on receiving.presence_absence
+   to presence_absence_processor;
+
```
kairstenfay: @tsibley : Jover and I were curious to know why this role does not receive `UPDATE` privileges on `receiving.presence_absence` despite the fact we're granting permissions here to `processing_log`. 

When we run `select * from information_schema.table_privileges where grantee='presence-absence-processor;` we see that there is only the `SELECT` `privilege_type` for `receiving.presence_absence`.

We modeled this role after `enrollment-processor` and see the same `privilege_type` for `receiving.enrollment` there, too.

tsibley: You can learn more about `information_schema` here: https://www.postgresql.org/docs/10/information-schema.html

joverlee521:
[Create a presence-absence-processor role for automation](https://trello.com/c/0Lhkh6n4/68-create-a-presence-absence-processor-role-for-automation)

tsibley:
While testing this on my own laptop, I found a privilege that was lacking on `warehouse.identifier` and `identifier_set`. I added a commit to fix that, plus another small tweak. Please take a look and see if those two commits look good to you? If they look good, please fold them into ("fixup") your commit.

joverlee521:
Everything looks good, it should all be under the one commit now.

PR #7, opened 2019-05-03
Update incidence_model_observation_v1

joverlee521:
Change `join` to `left join` for age bins tables to
include encounters where age is unknown/NULL.

Extract upper/lower bounds for age ranges
for easier sorting of encounters.


schema/deploy/shipping/views.sql
```diff
@@ -38,8 +38,10 @@ create view shipping.incidence_model_observation_v1 as
              else (encounter.details->'age'->>'value')::int
            end as age,
 
-           age_bin_fine.range as age_range_fine,
-           age_bin_coarse.range as age_range_coarse,
+           lower(age_bin_fine.range) as age_range_fine_lower,
+           upper(age_bin_fine.range) as age_range_fine_upper,
+           lower(age_bin_coarse.range) as age_range_coarse_lower,
+           upper(age_bin_coarse.range) as age_range_coarse_upper,
```
tsibley: Let's leave the existing range fields too? This makes it clear if the lower/upper bounds are inclusive or exclusive.

tsibley:
Deployed!

PR #9, opened 2019-05-16
etl enrollments: Handle flu@home documents

tsibley:
Audere made some small, backwards-compatible tweaks to the document
schema.  Our version check caught the version bump and bailed for human
intervention.  I verified the changes are ok.  In testing, I discovered
that some of the "chosenOptions" arrays have string elements.  This
violates the type defined in Audere's hutchProtocol.ts which requires
integer elements, so I guess the documents aren't actually type-checked
before being sent to us.  Handle it by forcing integers, which will fail
on unconvertible values.


tsibley:
This is required to process the big push of flu@home documents from yesterday afternoon that's piled up: https://backoffice.seattleflu.org/metabase/dashboard/2

PR #11, opened 2019-05-17
etl presence-absence: Handle another kind of container in the document

tsibley:
Matthew from NWGC pushed a different kind of document that nests items
inside a top-level "Update" key instead of within "store" → "items".
This document structure will be used by Samplify to send us updates to
previously sent items.  Our ETL process doesn't need that distinction,
so we look for one and then the other.

This will unstick the presence/absence ETL once deployed.


tsibley:
[Handle new presence/absence format](https://trello.com/c/rRdsgHK5/79-handle-new-presence-absence-format)

PR #17, opened 2019-05-28
Census tract as string

kairstenfay:
Previously pandas was converting `census_tract` on read to `float`. 
Now we're declaring `census_tract` as a `string` on pandas read_(csv/excel).

This PR also includes some minor refactoring for consistency between UW and SCH preprocess and 
punting null values mapping from clinical back to preprocess (by using `na_values` on pandas read).

### DIFF
--- a/lib/seattleflu/db/cli/command/clinical.py
+++ b/lib/seattleflu/db/cli/command/clinical.py
@@ -91,6 +91,12 @@ def parse_uw(uw_filename, uw_nwh_file, hmc_sch_file, output):
 
     barcode_quality_control(clinical_records, output)
 
+    # Convert dtypes
+    clinical_records["encountered"] = pd.to_datetime(clinical_records["encountered"])
+    # Age must be converted to Int64 dtype because pandas does not support NaNs
+    # with normal type 'int'
+    clinical_records["age"] = clinical_records["age"].astype(pd.Int64Dtype())
+
     # Subset df to drop missing barcodes
     clinical_records = clinical_records.loc[clinical_records['barcode'].notnull()]
 
@@ -117,10 +123,6 @@ def load_data(uw_filename: str, uw_nwh_file: str, hmc_sch_file: str):
     clinical_records = load_uw_metadata(uw_filename)
     clinical_records = clinical_records \
                         .rename(columns={'Collection.Date': 'Collection date'})
-    clinical_records['Collection date'] = pd.to_datetime(clinical_records['Collection date'])
-    # Convert datatype in 'age' column because
-    # Pandas does not support NaNs with normal type 'int'
-    clinical_records['Age'] = clinical_records['Age'].astype(pd.Int64Dtype())
 
     uw_manifest = load_manifest_data(uw_nwh_file, 'UWMC')
     uw_manifest = uw_manifest.rename(columns={'Barcode ID (Sample ID)': 'Barcode ID',
@@ -140,11 +142,12 @@ def load_uw_metadata(uw_filename: str) -> pd.DataFrame:
     Given a filename *uw_filename*, returns a pandas DataFrame containing
     clinical metadata.
     """
+    dates = ['Collection.Date']
     na_values = ['Unknown']
     if uw_filename.endswith('.csv'):
-        df = pd.read_csv(uw_filename, na_values=na_values)
+        df = pd.read_csv(uw_filename, na_values=na_values, parse_dates=dates)
     else:
-        df = pd.read_excel(uw_filename, na_values=na_values)
+        df = pd.read_excel(uw_filename, na_values=na_values, parse_dates=dates)
 
     df = add_metadata(df, uw_filename)
 
@@ -285,6 +288,8 @@ def parse_sch(sch_filename, output):
 
     # Drop unnecessary columns
     clinical_records = clinical_records[column_map.values()]
+
+    # Convert dtypes
     clinical_records["encountered"] = pd.to_datetime(clinical_records["encountered"])
 
     # Insert static value columns
--- a/lib/seattleflu/db/cli/command/clinical.py
+++ b/lib/seattleflu/db/cli/command/clinical.py
@@ -142,12 +142,16 @@ def load_uw_metadata(uw_filename: str) -> pd.DataFrame:
     Given a filename *uw_filename*, returns a pandas DataFrame containing
     clinical metadata.
     """
+    dtypes = {'census_tract': 'str'}
     dates = ['Collection.Date']
     na_values = ['Unknown']
+
     if uw_filename.endswith('.csv'):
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Investigate: Make sure census_tract string instead of float in db.](https://trello.com/c/45HmBAra/89-investigate-make-sure-censustract-string-instead-of-float-in-db)

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -361,16 +361,18 @@ def update_sample(db: DatabaseSession,
 
     if sample.encounter_id:
         assert sample.encounter_id == encounter_id, \
-            f"Sample {sample.id} already linked to another encounter {sample.encounter_id}"
-        return 
+            f"Sample already linked to another encounter {sample.encounter_id}"
+        return
+
+    LOG.info(f"Updating sample {sample.id}")
```
tsibley: This is a functional change in a whitespace-only commit. I suspect rebasing `census_tract_as_string` onto `sch_clinical_updates` introduced this without you realizing. One quick way to check that a whitespace-only commit contains only whitespace is to examine the output of `git show --ignore-all-space COMMIT`.

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -263,9 +261,7 @@ def hispanic_latino(ethnic_group: str) -> list:
 
     ethnic_map = {
         "Not Hispanic or Latino": "no",
-        "Hispanic or Latino": "yes",
-        "Unknown": None,
-        "NULL": None
+        "Hispanic or Latino": "yes"
```
tsibley: For the future, leaving a trailing comma on the last element of a list or dictionary is good practice in Python, as it results in smaller diffs.

lib/seattleflu/db/cli/command/preprocess.py
```diff
@@ -141,8 +141,9 @@ def load_uw_metadata(uw_filename: str) -> pd.DataFrame:
     Given a filename *uw_filename*, returns a pandas DataFrame containing
     clinical metadata.
     """
+    na_values = ['Unknown']
```
tsibley: Is it not necessary to also name `NULL` here, since it was in the previous value maps in the etl routine?

lib/seattleflu/db/cli/command/preprocess.py
```diff
@@ -141,8 +141,9 @@ def load_uw_metadata(uw_filename: str) -> pd.DataFrame:
...
     if uw_filename.endswith('.csv'):
-        df = pd.read_csv(uw_filename)
+        df = pd.read_csv(uw_filename, na_values=na_values)
     else:
         df = pd.read_excel(uw_filename)
```
tsibley: Should `na_values` also be passed to `read_excel`?

lib/seattleflu/db/cli/command/etl/clinical.py
kairstenfay: Good catch!

lib/seattleflu/db/cli/command/preprocess.py
kairstenfay: Nope. So long as the default na_values are kept in pandas (keep_default_na=True (by default)), 'NULL' will be converted to null.

lib/seattleflu/db/cli/command/preprocess.py
kairstenfay: Nope. So long as the default na_values are kept in pandas (keep_default_na=True (by default)), 'NULL' will be converted to null.

lib/seattleflu/db/cli/command/etl/clinical.py
kairstenfay: I see that now, too.

lib/seattleflu/db/cli/command/preprocess.py
kairstenfay: It gets added later just not part of this commit.

kairstenfay:
>     * Seeing the data type conversion in context, I realize now that it might be best done after reading in the files and after renaming the columns to our standard names. The data type conversions would then not be specific to the input sources. The way this works in practice is you read in all columns as strings (so as-is, with no conversion), and then later convert any you want to be ints, dates, bools, etc to those data types.

What about with `census_tract`? Since this is a direct output of our geocoding library, it should theoretically not change in name. 
Otherwise, we would have to read in (pandas converts to float), convert to int, then convert to string to get the desired value we want. 

I'm OK with type casting the dtypes of Age and Date post column-renaming, though.

kairstenfay:
With respect to Date casting, it appears that the the clinical records' date column gets read in as an object, but the Date column for the manifest data is automatically `dateTime`. This is creating a merge error before I can rename columns...

lib/seattleflu/db/cli/command/preprocess.py
tsibley: Ah, thanks!

tsibley:
> Otherwise, we would have to read in (pandas converts to float), convert to int, then convert to string to get the desired value we want.

If pandas is left to its defaults, yes. I wasn't clear enough, but what I was suggesting was to read in _all_ columns as strings (so as-is, with no conversion) and then only later explicitly convert specific columns to ints, dates, bools, etc. The `census_tract` column would be left as a string from the beginning.

kairstenfay:
> > Otherwise, we would have to read in (pandas converts to float), convert to int, then convert to string to get the desired value we want.
> 
> If pandas is left to its defaults, yes. I wasn't clear enough, but what I was suggesting was to read in _all_ columns as strings (so as-is, with no conversion) and then only later explicitly convert specific columns to ints, dates, bools, etc. The `census_tract` column would be left as a string from the beginning.

Given the state of the data, this will not work on encounter date for UW data.
UW clinical data as a string is formatted as 'MM/DD/YYYY 0:00', but manifest data is formatted as 'MM/DD/YYY'.

To get a string-wise merge to work, we would have to trim the ' 0:00' off of UW clinical data. IMO that would put us in a worse spot than we currently are in.

I think your idea of reading all dtypes in as `str` and converting post-renaming is interesting. However I am extremely hesitant to do this without enforcing some data quality standards on the clinical side.

kairstenfay:
I'm finally done rebasing this PR.

kairstenfay:
_rebasing intensifies_
Now updated to reflect the latest changes to master.

tsibley:
Just checking in: this is blocked on new data from Izzy, right?

kairstenfay:
> Just checking in: this is blocked on new data from Izzy, right?

Yes, she said she would pass along the new data this Friday.

tsibley:
Rebased this on top of current master to resolve for conflicts. Will review again now.

PR #18, opened 2019-05-29
Create identifier-minter role

kairstenfay:
Creates an identifier-minter role in the database via sqitch.

### DIFF
--- a/schema/deploy/roles/identifier-minter/grants.sql
+++ b/schema/deploy/roles/identifier-minter/grants.sql
@@ -1,0 +1,24 @@
+-- Deploy seattleflu/schema:roles/identifier-minter/grants to pg
+-- requires: warehouse/identifier
+-- requires: roles/identifier-minter/create
+
+begin;
+
+-- This change is designed to be sqitch rework-able to make it easier to update
+-- the grants for this role.
+
+grant connect on database :"DBNAME" to "identifier-minter";
+
+grant usage
+   on schema warehouse
+   to "identifier-minter";
+
+grant select
+   on warehouse.identifier_set
+   to "identifier-minter";
+
+grant select, insert
+   on warehouse.identifier
+   to "identifier-minter";
+
+commit;
--- a/schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
+++ b/schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
@@ -8,7 +8,9 @@ begin;
 --
 --    cual-id create ids --existing-ids ...
 --
-create function warehouse.identifier_barcode_distance_check() returns trigger as $$
+-- CAUTION ADVISED when changing implementation of this this function, as it
+-- will run as the owner, not the current user.
+create or replace function warehouse.identifier_barcode_distance_check() returns trigger as $$
     declare
         minimum_distance integer := 3;
         conflicting_barcodes citext[];
@@ -57,7 +59,14 @@ create function warehouse.identifier_barcode_distance_check() returns trigger as
             return NEW;
         end if;
     end
-$$ language plpgsql;
+$$ language plpgsql
+   security definer
+   -- Explicitly restrict which schemas the code inside the function can find
+   -- other tables, functions, etc. without qualification
+   SET search_path = pg_catalog, public, pg_temp;
+
+drop trigger identifier_barcode_distance_check_before_insert on warehouse.identifier;
+drop trigger identifier_barcode_distance_check_before_update on warehouse.identifier;
 
 create trigger identifier_barcode_distance_check_before_insert
     before insert on warehouse.identifier
--- a/schema/deploy/roles/identifier-minter/create.sql
+++ b/schema/deploy/roles/identifier-minter/create.sql
@@ -1,0 +1,9 @@
+-- Deploy seattleflu/schema:roles/identifier-minter/create to pg
+
+begin;
+
+create role "identifier-minter";
+
+comment on role "identifier-minter" is 'For minting identifiers and making barcode labels';
+
+commit;
--- a/schema/deploy/warehouse/identifier/triggers/barcode-distance-check@2019-05-31-identifier-minter.sql
+++ b/schema/deploy/warehouse/identifier/triggers/barcode-distance-check@2019-05-31-identifier-minter.sql
@@ -8,7 +8,7 @@ begin;
 --
 --    cual-id create ids --existing-ids ...
 --
-create or replace function warehouse.identifier_barcode_distance_check() returns trigger as $$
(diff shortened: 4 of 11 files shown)


kairstenfay:
[Add "identifier-minter" role](https://trello.com/c/nR0hgAwU/102-add-identifier-minter-role)

schema/deploy/roles/identifier-minter/grants.sql
```diff
@@ -0,0 +1,21 @@
...
+-- This change is designed to be sqitch rework-able to make it easier to update
+-- the grants for this role.
+
+grant connect on database :"DBNAME" to "identifier-minter";
+
+grant usage
+   on schema warehouse
+   to "identifier-minter";
+
+grant select
```
joverlee521: The identifier-minter needs to be able to insert into warehouse.identifier

kairstenfay: Thanks, I thought so but I couldn't explicitly find that in the SQL code when I was making this yesterday. I'm going to test the role now.

kairstenfay:
As discussed with @tsibley, I ran into an issue where the `identifier-minter` role needs UPDATE, DELETE, or TRUNCATE permissions on `warehouse.identifier` in order to run the trigger function `warehouse.identifier_barcode_distance_check()`. 

As Thomas predicted, my initial idea of granting `UPDATE` on the table but immediately revoking it thereafter for each individual column did not work. 

@tsibley , you suggested that we make the trigger function be run by its owner. I'm unsure how to proceed on this.

tsibley:
One thought I had was to grant update to the role, but then effectively block those updates using a [row-level security policy](https://www.postgresql.org/docs/10/ddl-rowsecurity.html). I believe this could be made to work, but would require a lot of unwieldy and easy-to-get-wrong RLS policies because of this behaviour:

> When row security is enabled on a table (with `ALTER TABLE ... ENABLE ROW LEVEL SECURITY`), all normal access to the table for selecting rows or modifying rows must be allowed by a row security policy.

My next thought is to mark the trigger function `security definer` (see the [`create function`](https://www.postgresql.org/docs/10/sql-createfunction.html) docs). However, I'm not sure trigger functions respect that attribute or not; they may always run as the invoking user. Would you give `security definer` a try? If it doesn't work, we have to consider other options. If it does, then we can do the work of auditing the trigger function to make sure marking it `security definer` won't introduce security problems.

Even though this may seem like a lot of work to avoid a simple grant that may be unlikely to be abused or misused accidentally, we do eventually want to open-up identifier minting to more users. Allowing updates (or deletes), in particular, would risk the integrity of our sample tracking.

kairstenfay:
@tsibley I _think_ I added the `security definer` properly according to the postgres docs, but the same permissions error still arises in commit e2111e8623691a89f5b81d21aaf96c7aba64b6c1

tsibley:
> @tsibley I _think_ I added the `security definer` properly according to the postgres docs, but the same permissions error still arises in commit [e2111e8](https://github.com/seattleflu/id3c/commit/e2111e8623691a89f5b81d21aaf96c7aba64b6c1)

Ah, it looks like the syntax isn't quite right. The `security definer` attribute should be part of the `create function` statement. Currently in that commit it's after the statement (after the semi-colon). I would be surprised if that SQL script ran without error…

kairstenfay:
> What's the intent here with setting `search_path`?

See "Writing SECURITY DEFINER Functions Safely" in the linked docs

tsibley:
> See "Writing SECURITY DEFINER Functions Safely" in the linked docs

Nod. I'm glad you read that section! :-) The example in those docs, though, is just an example. We don't have an `admin` schema, for instance.

Like the `security definer` attribute, the `set` attribute also needs to be part of the `create function` clause. I can show you what I mean in person if I'm not explaining this well in text.

kairstenfay:
> Nod. I'm glad you read that section! :-) The example in those docs, though, is just an example. We don't have an `admin` schema, for instance.
> 
> Like the `security definer` attribute, the `set` attribute also needs to be part of the `create function` clause. I can show you what I mean in person if I'm not explaining this well in text.

I understand the `set` attribute needing to be part of `create function`. I missed that semicolon. 
The reason it didn't throw an error is because I wasn't recreating the schema locally but copying it over from prod. Working on it now.

Re: our `admin` schema. You've caught me in my cargo cult programming phase. Should the `search_path` be omitted altogether?

kairstenfay:
FYI when recreating the db schema using `sqitch deploy` and `pg_restore --data-only ...`, I'm getting that same, unexplained error we saw previously. 

```
psql:schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql:63: ERROR:  type "citext[]" does not exist
LINE 4:         conflicting_barcodes citext[];
```

kairstenfay:
@tsibley the latest commit with the syntax error fixed allows `identifier-minter` to run the trigger functions that were previously raising insufficient permissions issues. :tada: 

To run locally, I simply ran 
```
drop trigger if exists identifier_barcode_distance_check_before_insert on warehouse.identifier;
drop trigger if exists identifier_barcode_distance_check_before_update on warehouse.identifier;
drop function if exists warehouse.identifier_barcode_distance_check();
```
before running the file modified in b5965a5af8de1daf26ebd80af23344182459e8ee.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
```diff
@@ -57,7 +57,8 @@ create function warehouse.identifier_barcode_distance_check() returns trigger as
             return NEW;
         end if;
     end
-$$ language plpgsql;
+$$ language plpgsql
+   security definer;
```
joverlee521: I think you need to do a `sqitch rework` to drop and re-create the function/triggers with the security definer rather than edit the original function

tsibley: Jover's correct: the way to update this function is to do a `sqitch rework` of the change.

kairstenfay:
The tag on the rework function will be `@2019-05-20-presence-absence-view` unless I change it. Is that tag ok?

tsibley:
It's not necessary to add a new tag for things to work correctly, but I'd probably `sqitch tag 2019-05-31-identifier-minter` before doing the rework.

kairstenfay:
I interpreted the Description under these docs (https://sqitch.org/docs/manual/sqitch-rework/) to mean that the rework has to have a tag. Using the `rework` function added the last known tag automatically.

kairstenfay:
Now using `sqitch rework` to make changes. Tests pass on my laptop.

tsibley:
Looks great, thanks! Merged.

tsibley:
and deployed.

PR #21, opened 2019-06-03
Encounter age

joverlee521:
This PR pulls age out of the encounter details into a column for the encounter table. 

After much discussion with @tsibley, we decided that the best data type for the age column is [intervals](https://www.postgresql.org/docs/11/datatype-datetime.html#DATATYPE-INTERVAL-INPUT), since it is much more human-readable than large numbers in months or age in floats. 

Public functions are created to convert the intervals to numeric values in both months and years. These are used in the modeling views since the modeling teams have a preference for numeric types.

### DIFF
--- a/lib/seattleflu/db/cli/command/etl/clinical.py
+++ b/lib/seattleflu/db/cli/command/etl/clinical.py
@@ -23,7 +23,7 @@ LOG = logging.getLogger(__name__)
 # clinical records lacking this revision number in their log.  If a
 # change to the ETL routine necessitates re-processing all clinical records,
 # this revision number should be incremented.
-REVISION = 1
+REVISION = 2
 
 
 @etl.command("clinical", help = __doc__)
@@ -111,6 +111,7 @@ def etl_clinical(*, action: str):
                     encountered     = record.document["encountered"],
                     individual_id   = individual.id,
                     site_id         = site.id,
+                    age             = age(record.document),
                     details         = encounter_details(record.document))
 
                 sample = update_sample(db,
@@ -188,13 +189,27 @@ def sex(sex_name) -> str:
 
     return sex_map.get(sex_name, "other")
 
+
+def age(document: dict) -> str:
+    """
+    Given a *document*, retrieve age value and 
+    return as a string to fit the interval format.
+
+    If no value is given for age, then will just return None.
+    """
+    age = document.get("age")
+    if age is None:
+        return None
+    return f"{float(age)} years"
+
+
 def encounter_details(document: dict) -> dict:
     """
     Describe encounter details in a simple data structure designed to be used
     from SQL.
     """
     return {
-            "age": age(document.get("age")),
+            "age": age_to_delete(document.get("age")), # XXX TODO: Remove age from details
             "locations": {
                 "home": {
                     "region": document.get("census_tract"),
@@ -209,15 +224,14 @@ def encounter_details(document: dict) -> dict:
             },
         }
 
-def age(age: float) -> dict:
+def age_to_delete(age: float) -> str:
     """
+    TODO: Delete this function once we remove age from details
     Given an *age*, return a dict containing its 'value' and a boolean for
     'ninetyOrAbove'.
-
     Currently applys math.ceil() to age to match the age from Audere.
     This may change in the future as we push to report age in months for
     participants less than 1 year old.
-
     If no value is given for *age*, then will just retun None.
     """
     if age is None:
@@ -227,6 +241,7 @@ def age(age: float) -> dict:
         "value": min(ceil(age), 90),
         "ninetyOrAbove": ceil(age) >= 90
     }
+
 
 def race(race_name: str) -> list:
     """
--- a/lib/seattleflu/db/cli/command/etl/enrollments.py
+++ b/lib/seattleflu/db/cli/command/etl/enrollments.py
@@ -22,7 +22,7 @@ LOG = logging.getLogger(__name__)
 # lacking this revision number in their log.  If a change to the ETL routine
 # necessitates re-processing all enrollments, this revision number should be
 # incremented.
-REVISION = 3
+REVISION = 4
 
 
 @etl.command("enrollments", help = __doc__)
@@ -113,6 +113,7 @@ def etl_enrollments(*, action: str):
                     encountered     = enrollment.document["startTimestamp"],
                     individual_id   = individual.id,
                     site_id         = site.id,
+                    age             = age(enrollment.document),
                     details         = encounter_details(enrollment.document))
 
                 # Find existing collected samples, or create skeletal sample
@@ -276,6 +277,7 @@ def upsert_encounter(db: DatabaseSession,
                      encountered: str,
                      individual_id: int,
                      site_id: int,
+                     age: str,
                      details: dict) -> Any:
     """
     Upsert encounter by its *identifier*.
@@ -287,6 +289,7 @@ def upsert_encounter(db: DatabaseSession,
         "encountered": encountered,
         "individual_id": individual_id,
         "site_id": site_id,
+        "age": age,
         "details": Json(details),
     }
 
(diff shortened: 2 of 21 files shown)


joverlee521:
[Model age relationally on encounters](https://trello.com/c/JGBt1wsW/78-model-age-relationally-on-encounters)

lib/seattleflu/db/cli/command/etl/enrollments.py
```diff
@@ -320,6 +326,20 @@ def upsert_encounter(db: DatabaseSession,
     return encounter
 
 
+def encounter_age(age_dict: dict) -> str:
+    """
+    Interpret the age of the individual at the time of encounter.
+    Converts age value from int to string to fit interval format. 
+    """
+    if not age_dict:
+        return None
```
kairstenfay: I think using `age_dict.get("value")` should handle lines 334-335.

joverlee521: I kept this line because age_dict.get() errors out if the input is a None value

schema/deploy/functions/age_conversion.sql
```diff
@@ -0,0 +1,39 @@
...
+    immutable
+    parallel safe;
+
+create or replace function public.age_month_to_years(age_in_months integer) returns numeric
+    language sql as $$
+        select round(age_in_months / 12::numeric, 2)
+    $$
+    returns null on null input
+    immutable
+    parallel safe;
```
tsibley: Let's drop this function since we don't plan to use it (since we're not storing months in years).

schema/verify/functions/age_conversion.sql
```diff
@@ -0,0 +1,30 @@
...
+	begin
+
+		with ages (age_as_interval, age_in_months) as (
+		    select make_interval(months => x), x from generate_series(0, 12) as x
+		    union
+		    select make_interval(years => x), x * 12 from generate_series(1, 6) as x
+		)		
+		select ages.age_in_months, age_in_months(age_as_interval), age_in_years(age_as_interval), age_month_to_years(ages.age_in_months) 
+		  into age_in_months, months, years, month_to_years
+		  from ages;
```
tsibley: This only tests a single row of the `ages` table. In PL/pgSQL (the language inside this `do` block), it's good to use the  `select … into strict …` form of `select … into` so that an error is thrown if more than one (or no) rows are returned.

You can loop over the results of a query and assert on each row's values using a for loop: https://www.postgresql.org/docs/10/plpgsql-control-structures.html#PLPGSQL-RECORDS-ITERATING

schema/verify/functions/age_conversion.sql
```diff
@@ -0,0 +1,30 @@
...
+  
+  		assert age_in_months = months;
+		assert years = round(age_in_months / 12::numeric, 2);
+		assert month_to_years = years;
```
tsibley: The test of `years` and `month_to_years` is effectively inlining the implementation of the function, which is not a useful test because they're not independent. It would be better to hardcode the test input values and expected results in a temporary table of tests. An example of this can be found in `schema/verify/functions/hamming_distance.sql`.

schema/sqitch.plan
```diff
@@ -74,3 +74,4 @@ roles/identifier-minter/grants [warehouse/identifier roles/identifier-minter/cre
 warehouse/identifier/triggers/barcode-distance-check [warehouse/identifier/triggers/barcode-distance-check@2019-05-31-identifier-minter warehouse/identifier functions/hamming_distance] 2019-05-31T17:41:38Z Kairsten Fay <kfay@fredhutch.org> # Mark the trigger function
 warehouse/encounter/age [warehouse/encounter] 2019-05-30T23:09:41Z Jover Lee <joverlee@fredhutch.org> # Add age column to encounter
 functions/age_conversion 2019-05-31T20:57:07Z Jover Lee <joverlee@fredhutch.org> # Age conversion functions to change age interval to numeric age
+functions/interval_range 2019-06-03T16:03:03Z Jover Lee <joverlee@fredhutch.org> # Create type intervalrange
```
tsibley: This is a data type, not function. (The function definition is only in support of the data type.) How about renaming the change to types/intervalrange?

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -209,24 +209,20 @@ def encounter_details(document: dict) -> dict:
...
+    year = int(age)
+    month = round((age % 1) * 12)
+    return f"{year} years {month} months"
 
-    return {
-        "value": min(ceil(age), 90),
-        "ninetyOrAbove": ceil(age) >= 90
-    }
 
 def race(race_name: str) -> list:
```
tsibley: I can't comment on a line below which is out of context, but when I try running the new clinical ETL on the production data, I run into the `UnknownRaceError`. Were you able to successfully run this on your computer, or do you find the same error?

kairstenfay: Re: the UnknownRaceError. That made me think the production data have the race 'Unknown' instead of NULL (as newly defined in `preprocess.py` using pandas `na_values` on read).

tsibley: Yep, I believe it does because we loaded the data before that change to preprocess was made.

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -209,24 +209,20 @@ def encounter_details(document: dict) -> dict:
...
-    If no value is given for *age*, then will just retun None.
+    If no value is given for *age*, then will just return None.
     """
     if age is None:
         return None
+    if age > 90:
+        return "90 years"
+    year = int(age)
+    month = round((age % 1) * 12)
+    return f"{year} years {month} months"
```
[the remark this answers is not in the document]
joverlee521: Yay! Fractional years work 🎉

schema/deploy/shipping/age-bin-v2.sql
```diff
@@ -0,0 +1,65 @@
...
+comment on table shipping.age_bin_fine_v2 is 
+    'Version 2 of fine-grained age bins that uses intervalrange type';
+comment on column shipping.age_bin_fine_v2.range is
+    'The range of each age bin defined as intervalrange';
+
+insert into shipping.age_bin_fine_v2 values
+    ('[0 mon,2 mons)'::intervalrange),
+    ('[2 mons,7 mons)'::intervalrange),
+    ('[7 mons,2 years)'::intervalrange),
+    ('[2 years,5 years)'::intervalrange),
```
tsibley: I'm not sure if these first four bins are quite right, based on my interpretation of [this Slack message](https://seattle-flu-study.slack.com/archives/CE9B78CS0/p1556051395043700) and [this other message](https://seattle-flu-study.slack.com/archives/CE9B78CS0/p1555982567033000). I interpreted it as:

```
[0 months, 1 month)   -- first month of life
[1 month, 6 months)   -- before 6 months
[6 months, 1 year)    -- before 1 year
[1 year, 5 years)     -- before 5 years
```

Thoughts? I know I found those two messages I linked to a bit confusing because they leave gaps between suggested ranges and/or don't specify inclusive/exclusive bounds, so maybe we should just ask for clarification.

schema/deploy/shipping/age-bin-v2.sql
```diff
@@ -0,0 +1,65 @@
...
+create index age_bin_coarse_range_idx_v2 on shipping.age_bin_coarse_v2 using gist(range);
+
+comment on table shipping.age_bin_coarse_v2 is
+    'Version 2 of coarse-grained age bins that uses intervalrange type';
+comment on column shipping.age_bin_coarse_v2.range is
+    'The range of each age bin defined as intervalrange';
+
+insert into shipping.age_bin_coarse_v2 values
+    ('[0 mon,7 mons)'::intervalrange),
+    ('[7 mons, 5 years)'::intervalrange),
```
tsibley: If we change the bins above, then ditto here:

```
[0 months, 6 months)
[6 months, 5 years)
```

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -188,13 +189,29 @@ def sex(sex_name) -> str:
...
+    Given a *document*, retrieve age value and 
+    return as a string to fit the interval format.
+
+    If no value is given for age, then will just return None.
+    """
+    age = document.get("age")
+    if age is None:
+        return None
+    if age > 90:
+        return "90 years"
```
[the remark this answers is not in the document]
tsibley: The reason for this is that we want this kind of age masking to only happen outside of id3c. If we do it in the ETL process, then we've sort of hidden from ourselves that the unmasked ages exist in `receiving` because we won't see them in `warehouse`.

joverlee521:
I've updated the age bins and deleted the check for >90 in the etl process. If nothing else needs to be changed this can be merged 😄

tsibley:
Looks good, merged!

PR #23, opened 2019-06-04
Rename preprocess command to clinical

kairstenfay:
Align top-level commands along the data instead of similar actions.
Rename `preprocess` command to `clinical`.
Rename `uw-clinical` subcommand to `uw`.
Rename `sch-clinical` subcommand to `sch`.

PR #24, opened 2019-06-04
Split clinical preprocess into parse and upload steps

kairstenfay:
https://trello.com/b/MFOJfX46/sfs-tasks
https://trello.com/c/oYB01LHh/121-rename-preprocess-command-to-clinical (addressed in PR #23).

This PR lives as a separate one from #23 for consistency with Trello. It felt a little like overkill but I wanted to err on the side of preferring more organization.

### DIFF
--- a/lib/seattleflu/db/cli/command/clinical.py
+++ b/lib/seattleflu/db/cli/command/clinical.py
@@ -1,5 +1,5 @@
 """
-Pre-process clinical data.
+Parse and upload clinical data.
 
 Clinical data will contain PII (personally identifiable information) and
 unnecessary information that does not need to be stored. This process will only
@@ -25,7 +25,7 @@ def clinical():
     pass
 
 # UW Clinical subcommand
-@clinical.command("uw")
+@clinical.command("parse-uw")
 @click.argument("uw_filename", metavar = "<UW Clinical Data filename>")
 @click.argument("uw_nwh_file", metavar="<UW/NWH filename>")
 @click.argument("hmc_sch_file", metavar="<HMC/SCH filename>")
@@ -33,7 +33,7 @@ def clinical():
     help="The filename for the output of missing barcodes")
 
 
-def uw(uw_filename, uw_nwh_file, hmc_sch_file, output):
+def parse_uw(uw_filename, uw_nwh_file, hmc_sch_file, output):
     """
     Process and insert clinical data from UW.
 
@@ -53,6 +53,9 @@ def uw(uw_filename, uw_nwh_file, hmc_sch_file, output):
     <output filename> is the desired filepath of the output CSV of problematic
     barcodes encountered while parsing. If not provided, the problematic
     barcodes print to the log.
+
+    All clinical records parsed are output to stdout as newline-delimited JSON
+    records.  You will likely want to redirect stdout to a file.
     """
     clinical_records, uw_manifest, nwh_manifest, hmc_manifest = load_data(uw_filename,
         uw_nwh_file, hmc_sch_file)
@@ -105,9 +108,7 @@ def uw(uw_filename, uw_nwh_file, hmc_sch_file, output):
     clinical_records['individual'] = clinical_records['individual'].apply(generate_hash)
     clinical_records['identifier'] = clinical_records['identifier'].apply(generate_hash)
 
-    session = DatabaseSession()
-    with session, session.cursor() as cursor:
-        clinical_records.apply(lambda x: insert_clinical(x, cursor), axis=1)
+    print(clinical_records.to_json(orient='records', lines=True))
 
 
 def load_data(uw_filename: str, uw_nwh_file: str, hmc_sch_file: str):
@@ -220,21 +221,6 @@ def generate_hash(identifier: str):
     return new_hash.hexdigest()
 
 
-def insert_clinical(df: pd.DataFrame, cursor):
-    """
-    Given a pandas DataFrame, inserts it as a JSON document into the
-    receiving.clinical table
-    """
-
-    LOG.debug(f"Inserting clinical data for barcode: «{df['barcode']}» ")
-
-    document = df.to_json(date_format='iso')
-
-    cursor.execute(
-        "insert into receiving.clinical (document) values (%s)",
-        (document,)
-    )
-
 def print_problem_barcodes(problem_barcodes: pd.DataFrame, output: str):
     """
     Given a pandas DataFrame of *problem_barcodes*, writes the data to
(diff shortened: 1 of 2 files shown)


kairstenfay:
[Split clinical preprocessing into separate parse and upload steps](https://trello.com/c/cJDbjAwK/122-split-clinical-preprocessing-into-separate-parse-and-upload-steps)

lib/seattleflu/db/cli/command/clinical.py
```diff
@@ -102,9 +108,7 @@ def uw_clinical(uw_filename, uw_nwh_file, hmc_sch_file, output):
     clinical_records['individual'] = clinical_records['individual'].apply(generate_hash)
     clinical_records['identifier'] = clinical_records['identifier'].apply(generate_hash)
 
-    session = DatabaseSession()
-    with session, session.cursor() as cursor:
-        clinical_records.apply(lambda x: insert_clinical(x, cursor), axis=1)
+    print(clinical_records.to_csv(index=False))
```
joverlee521: Do we want to match the manifest command and make this print out ndjson?

kairstenfay: I had a similar thought. My reason for choosing csv over ndjson comes down to an argument for splitting parse & upload Thomas laid out in the related Trello card:
>better inspectability of parsed records (you can parse them without uploading anywhere)

I figured if the original data were in CSV format and someone might want to inspect them after parsing, it would be easier for them to do so if the output was in CSV format, too.

Of course, Thomas is taking manifests from XLSX format and outputting them to ndjson. However, I don't think manifest data are really meant to be read in the same way clinical data are. Perhaps I'm just splitting hairs here, but I appreciate your comment and welcome any feedback.

tsibley: The ability to inspect the output of parse is an important consideration. :-) From my perspective, NDJSON is just as inspectable as CSV, but I realize that is not the case for everyone!

I think the things that swayed and sway my thinking towards NDJSON are:

1. It matches the exact format/document which ends up in the database and doesn't go through an additional, harder-to-inspect layer of transformation by `pd.read_csv` and `df.to_json`. This makes the data more inspectable in some ways (although less accessible as NDJSON, maybe?).

2. It is clear how to represent nested data structures, like the provenance information. This kind of information isn't in the clinical data pipeline yet, but I can imagine wanting to add it sooner than later.

I'm not dead set on NDJSON here or 100% against the current CSV output, so would like to hear any thoughts you both have.

lib/seattleflu/db/cli/command/clinical.py
```diff
@@ -230,12 +233,12 @@ def insert_clinical(df: pd.DataFrame, cursor):
...
     """
     Given a pandas DataFrame of *problem_barcodes*, writes the data to
-    stdout unless a filename *output* is given.
+    the log unless a filename *output* is given.
     """
     if output:
         problem_barcodes.to_csv(output, index=False)
     else:
-        print(problem_barcodes.to_csv(index=False))
+        LOG.info(problem_barcodes.to_csv(index=False))
```
tsibley: ++for outputting to the log. :-) This spits out the CSV as a one honkin' log message, though, and I wonder if it would be nicer to output each row of `problem_barcodes` as a separate, English log message?

I would also suggest the use of `LOG.warning` instead of `info` to imply some greater amount of attention.

lib/seattleflu/db/cli/command/clinical.py
kairstenfay: It sounds like it's two against one :anguished: . 
I think I prefer consistency in the parse/upload process over my preference for reading CSV vs. JSON. I'm working on the changes now.

lib/seattleflu/db/cli/command/clinical.py
```diff
@@ -257,30 +252,63 @@ def sch_clinical(sch_filename):
...
-
-    session = DatabaseSession()
-    with session, session.cursor() as cursor:
-        df.apply(lambda x: insert_clinical(x, cursor), axis=1)
+    clinical_records["FluShot"] = None
+    clinical_records["Race"] = None
+    clinical_records["HispanicLatino"] = None
+    clinical_records["MedicalInsurace"] = None
+
+    print(clinical_records.to_json(orient='records', lines=True))
```
joverlee521: Oh! Does this print out ndjson automatically? Maybe we can refactor the manifest process to do the same?

kairstenfay: Yep, `lines=True` prints a new JSON record per line. 
Oh yes, that is a possibility. For some reason it didn't occur to me that `manifest` also uses pandas.

tsibley: The manifest stuff doesn't use `pandas.DataFrame.to_json`, though, because I wanted streaming output (instead of being all buffered at once) and more explicit control (and less magic) over the JSON produced. For example, the reason the date-as-bigint thing was an issue is because the default is to represent dates as epoch values (seconds since 1970-01-01 00:00:00) instead of ISO 8601 strings.

So the manifest code instead defines a `dump_ndjson(iterable)` function which calls `as_json(dict)` from seattleflu.db.datatypes. This module is shared by some other code as well.

This `dump_ndjson()` could be refactored and shared by the clinical code here too.

kairstenfay: Would you like me to update the `parse` process here to use `dump_ndjson()`? Or should we wait to do so in the refactoring-of-shared-code PR?

lib/seattleflu/db/cli/command/clinical.py
```diff
@@ -230,12 +233,15 @@ def insert_clinical(df: pd.DataFrame, cursor):
...
+    the log unless a filename *output* is given.
     """
     if output:
         problem_barcodes.to_csv(output, index=False)
     else:
-        print(problem_barcodes.to_csv(index=False))
+        problem_barcodes.apply(lambda x: LOG.warning(
+            f"{x['problem']} in MRN {x['MRN']}, Accession {x['Accession']}, "
+            f"barcode {x['barcode']}"
+        ), axis=1)
```
tsibley: This was my fault for suggesting it, but I realize now that we probably don't want to be logging PHI as it'll end up in the system logs. System logs may end up being shipped off-machine too, depending on the system. :grimacing: Can we instead log the problem, barcode (if any), filename, sheet, and row number? I think that will let someone (one of us!) track down problems without revealing any PHI. Sending full details with PHI to the optional CSV output is still fine, as it's obvious to the end user where that data goes.

lib/seattleflu/db/cli/command/clinical.py
tsibley: Whatever is less work for you!

kairstenfay: I'd rather kick that down the road to #27

lib/seattleflu/db/cli/command/clinical.py
kairstenfay: This has been fixed in the latest push

kairstenfay: Commit 553e8cf2cd33b2901b97ef7041f107f15486d5fe

tsibley: Thank you! Sorry for suggesting the wrong thing.

PR #25, opened 2019-06-05
Add new join view

joverlee521:
Adds new shipping view that joins the
incidence_modeling_observation_v2 view with the
presence_absence_results_v1 view.

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -131,7 +131,8 @@ grant select
     on shipping.presence_absence_result_v1
     to "incidence-modeler";
 
-create view shipping.incidence_model_observation_v2 as
+
+create or replace view shipping.incidence_model_observation_v2 as
 
     select encounter.identifier as encounter,
 
@@ -222,5 +223,16 @@ revoke all
 grant select
    on shipping.incidence_model_observation_v2
    to "incidence-modeler";
+
+
+create or replace view shipping.observation_with_presence_absence_result_v1 as
+
+    select target,
+           present,
+           present::int as presence,
+           observation.*
+      from shipping.incidence_model_observation_v2 as observation
+      join shipping.presence_absence_result_v1 using (sample)
+      order by site, encounter, sample, target;
 
 commit;
--- a/schema/deploy/shipping/views@2019-06-04-join-shipping-views.sql
+++ b/schema/deploy/shipping/views@2019-06-04-join-shipping-views.sql
@@ -14,8 +14,6 @@ begin;
 -- there needs to be a lag between view development and consumers being
 -- updated, copy the view definition into v2 and make changes there.
 
-drop view shipping.observation_with_presence_absence_result_v1;
-
 create or replace view shipping.incidence_model_observation_v1 as
 
     select encounter.identifier as encounter,
@@ -133,8 +131,7 @@ grant select
     on shipping.presence_absence_result_v1
     to "incidence-modeler";
 
-
-create or replace view shipping.incidence_model_observation_v2 as
+create view shipping.incidence_model_observation_v2 as
 
     select encounter.identifier as encounter,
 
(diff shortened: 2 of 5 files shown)


schema/deploy/shipping/views.sql
```diff
@@ -223,4 +225,15 @@ grant select
    on shipping.incidence_model_observation_v2
    to "incidence-modeler";
 
+
+create view shipping.observation_with_presence_absence_result as
```
joverlee521: Happy to change the view name if there's something else more appropriate!

schema/deploy/shipping/views.sql
```diff
@@ -223,4 +225,15 @@ grant select
...
+
+    select target,
+           present,
+           present::int as presence,
+           observation.*
+      from shipping.incidence_model_observation_v2 as observation
+      join shipping.presence_absence_result_v1 using (sample)
+      order by site, encounter, sample, target;
```
kairstenfay: Do we want to user an inner `join` here or something else? I know this is how @tsibley originally had it, but I just wanted to know the reasoning behind the choice.

joverlee521: I was under the impression that this view will display only encounters that have presence/absence results, but I may be wrong.

kairstenfay: Yep, I'm just curious to know why we would only want to see only encounters that have presence/absence results rather than showing all encounters and having the user optionally filter to exclude `null` presence/absence.

tsibley: Good question to consider! :-) I believe the reason why I didn't make it a left join was that the only purpose of the view was to join up encounters with presence/absence for convenience. It seemed that nearly all use cases would want to add the filter, so I made it part of the view.

I guess the question is in what cases would we a) want to use this view and b) want to include encounters without p/a results?  Perhaps one way to answer this is to look at how the draft view is currently used in Metabase.

schema/deploy/shipping/views.sql
tsibley: I think it's a fine name. Better than mine! :-)

Should we version this? I don't have particular thoughts here one way or another. I don't know if you considered it or not...

joverlee521: I considered versioning it, but it's confusing that it uses observation v2 and presence/absence v1...

schema/deploy/shipping/views.sql
```diff
@@ -131,6 +131,8 @@ grant select
     on shipping.presence_absence_result_v1
     to "incidence-modeler";
 
+
+drop view shipping.incidence_model_observation_v2;
 create view shipping.incidence_model_observation_v2 as
```
tsibley: I prefer to use `create or replace view` statements when we can get away with them (when we're not re-ordering or re-typing columns), as they avoid some problems with inter-dependent views.

schema/deploy/shipping/views.sql
tsibley: Ah, that makes sense and might be confusing. I guess I'd see its version as starting at v1 and staying there so long as it kept using observation v2 and p/a v1 and the same join criteria.

I'm fine leaving it unversioned though and we can see how it goes.

kairstenfay: I slightly prefer the consistency of versioning this view.

schema/deploy/shipping/views.sql
joverlee521: It looks like all questions using the draft view are looking at encounters that have presence/absence results: https://backoffice.seattleflu.org/metabase/reference/databases/3/tables/66/questions

schema/deploy/shipping/views.sql
tsibley: @joverlee521 I'll leave this decision up to you?

schema/deploy/shipping/views.sql
tsibley: Ok! Let's leave this as an inner join then.

schema/deploy/shipping/views.sql
joverlee521: Latest push added versioning to the view. We're start at v1 😄

PR #28, opened 2019-06-07
Move existing api underneath /v1/receiving

kairstenfay:
From Trello:

> This provides better versioning and namespacing for growth and also means we can share nouns for resource endpoints between the receiving area and warehouse area, e.g.:
> 
> ```
> /v1/receiving/presence-absence
> /v1/warehouse/presence-absence
> ```
> 
> Support old endpoints as deprecated aliases for now, then work on getting consumers of the API moved over.

### DIFF
--- a/lib/seattleflu/api/routes.py
+++ b/lib/seattleflu/api/routes.py
@@ -9,14 +9,17 @@ from .utils.routes import authenticated_datastore_session_required, content_type
 
 LOG = logging.getLogger(__name__)
 
-api = Blueprint("api", __name__)
+api_v1 = Blueprint('api_v1', 'api_v1', url_prefix='/v1')
+api_unversioned = Blueprint('api_unversioned', 'api_unversioned', url_prefix='/')
 
 blueprints = [
-    api,
+    api_v1,
+    api_unversioned,
 ]
 
 
-@api.route("/", methods = ['GET'])
+@api_v1.route("/", methods = ['GET'])
+@api_unversioned.route("/", methods = ['GET'])
 def index():
     """
     Show an index page with documentation.
@@ -24,7 +27,8 @@ def index():
     return send_file("static/index.html", "text/html; charset=UTF-8")
 
 
-@api.route("/enrollment", methods = ['POST'])
+@api_v1.route("/receiving/enrollment", methods = ['POST'])
+@api_unversioned.route("/enrollment", methods = ['POST'])
 @content_types_accepted(["application/json"])
 @check_content_length
 @authenticated_datastore_session_required
@@ -45,7 +49,8 @@ def receive_enrollment(*, session):
     return "", 204
 
 
-@api.route("/presence-absence", methods = ['POST'])
+@api_v1.route("/receiving/presence-absence", methods = ['POST'])
+@api_unversioned.route("/presence-absence", methods = ['POST'])
 @content_types_accepted(["application/json"])
 @check_content_length
 @authenticated_datastore_session_required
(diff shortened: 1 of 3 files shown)


kairstenfay:
[Move existing API underneath the path /v1/receiving/](https://trello.com/c/GX5g1L0T/117-move-existing-api-underneath-the-path-v1-receiving)

Pipfile
```diff
@@ -17,6 +17,7 @@ pandas = "*"
 xlrd = "*"
 pyyaml = "*"
 deepdiff = "*"
+flask-deprecate = "*"
```
joverlee521: I think this also needs to get added to `Pipfile.lock`?

kairstenfay: Ah, for some reason I thought this got .gitignored (like a `node_modules` directory). I'll re-push with those changes.

lib/seattleflu/api/routes.py
```diff
@@ -3,17 +3,21 @@
 """
 import logging
 from flask import Blueprint, request, send_file
+from flask_deprecate import deprecate_blueprint
```
tsibley: I'm not sure flask-deprecate really has much use for us here. The `Warnings` header it adds (which is the only thing it does) is likely to be ignored by most clients. In my experience, it's not a header which sees wide use and is typically only used in caches. Thoughts?

lib/seattleflu/api/routes.py
```diff
@@ -3,17 +3,21 @@
...
 from . import datastore
 from .utils.routes import authenticated_datastore_session_required, content_types_accepted, check_content_length
 
 
 LOG = logging.getLogger(__name__)
 
-api = Blueprint("api", __name__)
+api = Blueprint('new', 'new', url_prefix='/v1')
+old_api = Blueprint('old', 'old', url_prefix='/')
```
tsibley: If we're going to keep this commit (see note above), I think the names `api` and `old_api` would be more future-proof by putting the URL version prefix into the variable name, e.g. `api_v1` and `api_unversioned`. That way when we add `/v2/…` endpoints, we can use `api_v2` and it'll be clear what's what. This also more readily (to me, at least) indicates that the different blueprints are used solely for versioning instead of other common uses like duplicating a set of shared endpoints.

tsibley:
A couple notes from review, also slightly tweaked the curl example you added to the readme and pushed a commit. Thanks for updating the doc! :-)

lib/seattleflu/api/routes.py
kairstenfay: I wasn't sure how else to warn the user of an endpoint deprecation. Do you have other ideas? I also looked at `flask-socketio`.

lib/seattleflu/api/routes.py
kairstenfay: I agree with this naming convention.

lib/seattleflu/api/routes.py
tsibley: Ah, I think we warn them by talking to them in person or via Slack. :-) We can deploy this and get people to shift endpoints later, which might wait until we have a /v2 we want them to use so they go / → /v2 instead of / → /v1 (but same behaviour) → /v2.

(socketio looks unrelated to me; it's about websockets and related protocols.)

tsibley:
Ah, could you also update the index document to use the new versioned and namespaced endpoints?

tsibley:
Looks good! I just pushed one more update to the API doc. Would you take a look? If it looks good to you, go ahead and merge!

PR #29, opened 2019-06-07
Add barcode qc check for sch clinical data

kairstenfay:
Perform quality control on `parse-sch` like we're doing in `parse-uw`.

### DIFF
--- a/lib/seattleflu/db/cli/command/clinical.py
+++ b/lib/seattleflu/db/cli/command/clinical.py
@@ -256,8 +256,10 @@ def print_problem_barcodes(problem_barcodes: pd.DataFrame, output: str):
 
 @clinical.command("parse-sch")
 @click.argument("sch_filename", metavar = "<SCH Clinical Data filename>")
+@click.option("-o", "--output", metavar="<output filename>",
+    help="The filename for the output of missing barcodes")
 
-def parse_sch(sch_filename):
+def parse_sch(sch_filename, output):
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Validate no duplicate barcodes in SCH clinical data](https://trello.com/c/NXeHDVJY/103-validate-no-duplicate-barcodes-in-sch-clinical-data)

lib/seattleflu/db/cli/command/clinical.py
```diff
@@ -191,7 +191,7 @@ def missing_barcode(df: pd.DataFrame) -> pd.DataFrame:
     missing_barcodes = df.loc[df['barcode'].isnull()].copy()
     missing_barcodes['problem'] = 'Missing barcode'
 
-    return missing_barcodes[['MRN', 'Accession', 'barcode', 'problem', '_metadata']]
+    return missing_barcodes
```
tsibley: Can you explain the reasoning for removing the subseting here (and in `duplicated_barcode` below)? My guess is that "more is better" since the full data for each row is only going to a file now?

kairstenfay: Good question! The original subsetting was arbitrary. It was done for the purposes of printing a smaller table to stdout. Now that we're writing de-identified log messages -or- generating a new "problem barcodes" csv, there is no need for subsetting.

PR #35, opened 2019-06-12
Self-test kits

joverlee521:
Support self-test samples from Flu@home kits. 
See commit messages for more details.


joverlee521:
[Support self-test samples with kit barcodes](https://trello.com/c/dODlM6lB/84-support-self-test-samples-with-kit-barcodes)

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -88,48 +90,170 @@ def etl_manifest(*, action: str):
...
+
+                        utm_collection_identifier = check_barcode(db, 
+                            record          = manifest_record,
+                            barcode         = utm_collection_barcode,
+                            identifier_set  = expected_identifier_sets["kit_collections"])
+
+                        # Skip record if no matching UTM collection identifier
+                        # found for the UTM collection barcode    
+                        if not utm_collection_identifier: continue
+
```
kairstenfay: Are these (and line 173 below) LOG-worthy?

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -88,48 +90,170 @@ def etl_manifest(*, action: str):
...
+                            collection_identifier = utm_collection_identifier.uuid,
+                            additional_details    = utm_details)
+
+                        utm_sample_id = utm_sample.id
+
+                    # Convert RDT collection barcode to full identifier and
+                    # check it's from the correct identifier set.
+                    rdt_collection_barcode = manifest_record.document.pop("rdt", None)
+                    rdt_sample_id = None
+                    rdt_status = "None"
```
kairstenfay: Why is `rdt_status` a string here?

lib/seattleflu/db/cli/command/etl/enrollments.py
```diff
@@ -116,57 +117,83 @@ def etl_enrollments(*, action: str):
...
                         continue
 
-                    # XXX TODO: Stop hardcoding this and handle other types.
-                    # ScannedSelfSwab and ManualSelfSwabbed are kit barcodes,
-                    # not collection barcodes.  TestStrip is an identifier
-                    # UUID, not barcode.
-                    #   - trs, 17 May 2019
-                    if sample["type"] != "ClinicSwab":
-                        LOG.warning(f"Skipping collected sample with unknown type {sample['type']}")
+                    if swab["type"] == "ClinicSwab":
```
kairstenfay: What is the motivation for renaming `sample` in the `etl enrollments` code to `swab`?

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -170,9 +294,9 @@ def upsert_sample(db: DatabaseSession,
     """
     Upsert sample by its *identifier* and/or *collection_identifier*.
 
-    An existing sample has its *encounter_id* updated, and the provided
-    *additional_details* are merged (at the top-level only) into the existing
-    sample details, if any.
+    An existing sample has its *identifier* and *collection_identifier* updated, 
+    and the provided *additional_details* are merged (at the top-level only) 
+    into the existing sample details, if any.
```
kairstenfay: Nice catch on the docstring here

lib/seattleflu/db/cli/command/etl/manifest.py
joverlee521: These are logged within `check_barcode` on line 378. Although the log message is not very specific to which barcode is unknown...

lib/seattleflu/db/cli/command/etl/manifest.py
joverlee521: All statuses are strings so I thought it would be more specific for it to say "None" rather than have a null value. But that can be changed if a null value is enough.

lib/seattleflu/db/cli/command/etl/enrollments.py
joverlee521: I couldn't think of a better word to encompass both samples and kits...

kairstenfay: Are they fundamentally different? I haven't considered kits before, but I thought samples have been swabs all along.

lib/seattleflu/db/cli/command/etl/manifest.py
kairstenfay: How many barcodes do these things have? :anguished:

joverlee521: Kits can have up to 5 barcodes (kit, utm collection, utm sample, rdt collection, test strip)

kairstenfay:
I would also love a walk-through of the work you've done here!

lib/seattleflu/db/cli/command/etl/manifest.py
tsibley: I think a null-value would be more appropriate since nulls are the special value intended for representing "unknown" or "missing", which is what we have here. Adding the string-literal `"None"` as one of those special values will require more special-casing in querying code later.

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -88,48 +90,170 @@ def etl_manifest(*, action: str):
...
+                    kit_barcode = manifest_record.document.pop("sample")
+                    kit_identifier = check_barcode(db, 
+                        record          = manifest_record, 
+                        barcode         = kit_barcode, 
+                        identifier_set  = expected_identifier_sets["kits"])
+
+                    # The only time a kit_identifier is not returned is if
+                    # there was no identifier that matches the barcode.
+                    # In this case, the record "skipped".
+                    if not kit_identifier: continue
```
tsibley: Control flow of this outer loop is split between the `check_barcode` function and the `if not kit_identifier: continue` line. The `check_barcode` function has an unobvious side-effect of calling `mark_skipped` which necessitates the conditional `continue`. I think the code would have a clearer flow if instead `check_barcode` didn't call `mark_skipped`. This would leave ownership of the record+loop processing in the loop itself and not in an abstracted function.

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -88,48 +90,170 @@ def etl_manifest(*, action: str):
...
-                # warehouse for each piece of information.
-                sample, status = upsert_sample(db,
-                    identifier            = sample_identifier.uuid,
-                    collection_identifier = collection_identifier.uuid if collection_identifier else None,
-                    additional_details    = manifest_record.document)
-
-                mark_loaded(db, manifest_record.id,
+                # Processing manifest records not involving self-test kits
+                if ("utm" not in manifest_record.document and 
+                        "rdt" not in manifest_record.document):
```
tsibley: I'm a little concerned about splitting the processing into two big conditional branches like this. Large, divergent conditional branches are an immediate red flag for me when writing/reading code. Effectively it means we have two separate ETL routines bundled into one. I think we could make it a unified routine, and if we can't, I think it'd be worth considering two separate routines.

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -88,48 +90,170 @@ def etl_manifest(*, action: str):
...
                     status = status,
                     sample_id = sample.id)
+                
+                else:
+                    # Convert kit barcode to full identifier and check 
+                    # it's from the correct identifier set.
+                    kit_barcode = manifest_record.document.pop("sample")
+                    kit_identifier = check_barcode(db, 
+                        record          = manifest_record, 
+                        barcode         = kit_barcode,
```
tsibley: There is trailing whitespace on some of the lines above and in other added lines as well (there's a "blank" line below which is all trailing whitespace). Trailing whitespace is generally considered undesirable (for various reasons, such as how it affects editor navigation), so it'd be nice to trim that. I believe your editor configuration options should include ways of automatically doing that (or at least showing you when it's present).

We should probably discuss using something automated around this and other code formatting styles.

lib/seattleflu/db/cli/command/etl/enrollments.py
```diff
@@ -116,57 +117,83 @@ def etl_enrollments(*, action: str):
...
+                        # by the processing lab will usually be loaded later 
+                        # and fill in the kit record.
+                        LOG.debug(f"Looking up collected kit code «{barcode}»")
+                        identifier = find_identifier(db, barcode)
+
+                        if not identifier:
+                            LOG.warning(f"Skipping {swab['type']} kit with unknown barcode «{barcode}»")
+                            continue
+                        
+                        if identifier.set_name != 'kits-fluathome.org':
```
tsibley: I suggest making this `if identifier.set_name not in expected_kit_identifier_sets` and defining that new variable up by `expected_collection_identifier_sets`. This will make it easier to find the valid set names for future refactoring to remove the hard-coding (see the comments near the top of this file).

lib/seattleflu/db/cli/command/manifest.py
```diff
@@ -51,7 +51,7 @@ def manifest():
 
 @click.option("--sample-column",
     metavar = "<column>",
-    help = "Name of the single column containing sample barcodes.  "
+    help = "Name of the single column containing sample or kit barcodes.  "
```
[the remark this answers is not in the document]
joverlee521: The thought here was that this `--sample-column` is the only required column option and the only required column for the kit manifest would be the kit barcodes.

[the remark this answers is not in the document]
tsibley: Nod. Thoughts on alternatively adding `--kit-column` and declaring it and `--sample-column` both be optional, but requiring at least one is present at command runtime?

This would also translate to the ETL process looking for a `kit` vs. `sample` key.

lib/seattleflu/db/cli/command/etl/__init__.py
```diff
@@ -140,4 +140,44 @@ def upsert_encounter(db: DatabaseSession,
...
+                            encounter_id: int) -> NamedTuple:
+    """
+    Update a sample by its *sample_id* to link to the given *encounter_id*
+    """
+    data = {
+        "sample_id": sample_id,
+        "encounter_id": encounter_id
+    }
+    
+    LOG.info(f"Linking existing sample {sample_id} to existing encounter {encounter_id}")
```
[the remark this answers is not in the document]
joverlee521: Yes, I was going to pull out the `update_sample` function from the clinical etl into this shared module

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -233,11 +357,142 @@ def upsert_sample(db: DatabaseSession,
...
+    Assess the validity of *barcode* for a given *barcode_name*.
+    Returns matching identifier from *db* if found. 
+    
+    Will mark the *record* as "skipped" if there is no matching identifier
+    found for the given *barcode* and return None.
+
+    Will error out if the returned identifier is not from the given 
+    *identifier_set*
+    """
+    barcode = barcode.lower()
```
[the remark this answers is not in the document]
joverlee521: Oh right barcodes are `citext`. Not necessary then 😄

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -233,11 +357,142 @@ def upsert_sample(db: DatabaseSession,
...
+                    %(utm_sample_id)s, 
+                    %(additional_details)s)
+        
+        on conflict (identifier) do update
+            set rdt_sample_id = %(rdt_sample_id)s,
+                utm_sample_id = %(utm_sample_id)s,
+                details = coalesce(kit.details, '{}') || %(additional_details)s
+        
+        returning kit_id as id, identifier, encounter_id, rdt_sample_id, utm_sample_id,
+            case when xmax::text::int = 0 then 'created' else 'updated' end as status
```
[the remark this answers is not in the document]
joverlee521: This was just an answer I found on [stackoverflow](https://stackoverflow.com/questions/34762732/how-to-find-out-if-an-upsert-was-an-update-with-postgresql-9-5-upsert) on how to determine if an update happened during as upsert. But it may be no longer required because I was thinking of separating things out like you did in `upsert_sample` to check if we are moving around kit-encounter references

joverlee521:
After discussion with @tsibley, we decided to restructure things and create a standalone etl process for kits that would process records from both `receiving.manifest` and `receiving.enrollment`. This way both `etl manifest` and `etl enrollments` can just have the core processes without much change to accommodate kit specific data.

lib/seattleflu/db/cli/command/manifest.py
```diff
@@ -268,6 +310,19 @@ def _parse(*,
         if column_groups[key]:
             manifest[f"{key}s"] = manifest[key].apply(list, axis="columns")
             manifest.drop(columns = manifest[key], inplace=True)
+    
+    # Set of columns names for barcodes
+    barcode_columns = {"sample", "collection", "kit", "test_strip"}
+    # Drop any rows that have duplicated barcodes
+    manifest = barcode_quality_control(manifest, barcode_columns)
+
```
kairstenfay: FYI `barcode_quality_control` doesn't drop duplicated barcodes. It flags them.

lib/seattleflu/db/cli/command/etl/manifest.py
```diff
@@ -98,8 +102,15 @@ def etl_manifest(*, action: str):
...
-                    f"Sample identifier found in set «{sample_identifier.set_name}», not {expected_identifier_sets['samples']}"
+                if (manifest_record.document.get("sample_type") and
+                    manifest_record.document["sample_type"] == "rdt"):
+                    assert sample_identifier.set_name in expected_identifier_sets["rdt"], \
+                        (f"Sample identifier found in set «{sample_identifier.set_name}», " +
+                        f"not {expected_identifier_sets['rdt']}")
+                else:
+                    assert sample_identifier.set_name in expected_identifier_sets["samples"], \
+                        (f"Sample identifier found in set «{sample_identifier.set_name}», " +
+                        f"not {expected_identifier_sets['samples']}")
```
kairstenfay: I think a few formatting changes snuck into this commit (lines 105-113 and 141-142 in `etl/manifest`).

lib/seattleflu/db/cli/command/etl/__init__.py
```diff
@@ -140,4 +140,76 @@ def upsert_encounter(db: DatabaseSession,
...
+
+        utm_sample = find_sample_by_id(db, kit.utm_sample_id)
+        
+        if utm_sample:
+            
+            update_sample(db,
+                sample         = utm_sample,
+                encounter_id   = kit.encounter_id)
+
+
```
kairstenfay: Did you mean to add `update_kit_samples` here only to delete it (specifically move it to `etl/kit` in commit d28106?

joverlee521: Yea, after restructuring the `update_kit_samples` is now only used by the kit etl so it no longer needs to be in the shared module

joverlee521:
I just noticed this today, but with the addition of kit-related samples, all encounters from `self-test` sites get doubled in all `shipping` views.

lib/seattleflu/db/cli/command/etl/kit.py
```diff
@@ -0,0 +1,629 @@
...
+        # Warn if kit is already linked to a different sample!
+        if ((kit.rdt_sample_id and kit.utm_sample_id) and
+            (sample.id not in {kit.rdt_sample_id, kit.utm_sample_id})):
+            LOG.warning(f"Kit «{kit.id}» already linked to another " +
+                        f"{data['sample_type']} «{kit[data['sample_type']]}»")
+
+        kit = db.fetch_row(sql.SQL("""
+            update warehouse.kit
+               set {} = %(sample_id)s,
+                   details = coalesce(details, null) || %(additional_details)s
```
tsibley: coalescing to null doesn't make sense, as the whole reason for coalesce is to use alternate values when some value is null. Appending `||` to null will produce null, so that an existing kit won't get updated details.

lib/seattleflu/db/cli/command/etl/kit.py
```diff
@@ -0,0 +1,629 @@
...
+                        sql.Identifier(sample_type)), data)
+
+    # Found kit → update
+    else:
+        status = "updated"
+        # Warn if kit is already linked to a different sample!
+        if ((kit.rdt_sample_id and kit.utm_sample_id) and
+            (sample.id not in {kit.rdt_sample_id, kit.utm_sample_id})):
+            LOG.warning(f"Kit «{kit.id}» already linked to another " +
+                        f"{data['sample_type']} «{kit[data['sample_type']]}»")
```
tsibley: Should this check require that both samples are present on the kit? It seems like it should check based on `sample.type`. You can do `getattr(kit, sample_type)` to dynamically get, e.g., `kit.utm_sample_id`.

lib/seattleflu/db/cli/command/manifest.py
```diff
@@ -238,16 +275,28 @@ def _parse(*,
...
-                find_one_column(manifest, single_columns[key]): key })
+            map_key = find_one_column(manifest, single_columns[key])
+            # Allow columns to map to multiple keys
+            if map_key in column_map:
+                map_key_copy = map_key + "_copy"
+                manifest[map_key_copy] = manifest[map_key]
+                column_map.update({ map_key_copy: key })
+            else:
+                column_map.update({
+                    find_one_column(manifest, single_columns[key]): key })
```
tsibley: I think there are two issue with this logic:

1. It works for the case of a source column specified twice for two different destination columns, but it doesn't work in the general case for a source column specified three times, for example. The copied column in the column map will be overwritten.

2. There is no check that a column with the `_copy` suffix doesn't already exist, which while unlikely, is definitely possible. Since we don't control the inputs to this code and don't regularly inspect them for changes, we should be as defensive as possible about how we handle them. Breaking in loud ways is better than breaking in silent ways.

I pushed two commits on top of this branch which address both those issues by reworking how the source → destination mapping works. Take a look and see what you think?

joverlee521: Yup! Completely agree with the two issues, I think I funneled myself too deep into the specific case for rdt samples. The source → destination mapping looks great!

joverlee521:
Tested all the parsing, uploading, and etl process on a fresh copy of production and everything worked as expected. 

Just a reminder, if this branch gets merged, it will inflate the number of encounters for the shipping views since the views can't handle multiple samples linked to one encounter.

tsibley:
> Just a reminder, if this branch gets merged, it will inflate the number of encounters for the shipping views since the views can't handle multiple samples linked to one encounter.

Ah, right. I took a look at that issue again and pushed up what I think is an ok fix as #39, although I haven't chatted with the modeling team.

PR #36, opened 2019-06-13
Ingest longitudinal childcare enrollments

kairstenfay:
This was one beast of an ETL process to write given the very wide nature of the underlying datasets. 
It is not yet ready for merging, because we are still waiting on one more dataset from SCH to process. 

I wanted to get additional eyes on this workflow especially around how I'm mapping data to existing values in the `etl longitudinal` command. Please share any thoughts you have.


lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+    if ethnic_group is None:
+        LOG.debug("No ethnic group response found.")
+        return [None]
+
+    ethnic_map = {
+        1: "yes",
+        2: "no",
+        3: "refused",  # TODO: is this what we want, or None instead?
+    }
```
kairstenfay: is there precedence for a value other than yes/no here?

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,278 @@
...
+    lnginal_records = weekly_assessments_wide_to_long(lnginal_records)
+    lnginal_records = select_all_that_apply_wide_to_long(lnginal_records)
+    lnginal_records = duplicate_audere_keys(lnginal_records)
+    lnginal_records = insert_static_values(lnginal_records)
+    lnginal_records = hash_pii(lnginal_records)
+
+    barcode_quality_control(lnginal_records, output)
+
+    print(lnginal_records.to_json(orient='records', lines=True))
+
```
kairstenfay: I'll refactor this to use the new `dump_ndjson` function Thomas recently wrote for clinical records

lib/seattleflu/db/cli/command/etl/__init__.py
```diff
@@ -140,4 +142,106 @@ def upsert_encounter(db: DatabaseSession,
...
+
+    if not sample:
+        LOG.error(f"No sample with identifier «{identifier}» found")
+        return None
+
+    LOG.info(f"Found sample {sample.id} «{sample.identifier}»")
+    return sample
+
+
+def update_sample(db: DatabaseSession,
```
joverlee521: Ah, I totally forgot we had this function in `etl clinical`, I wrote another one very similar in https://github.com/seattleflu/id3c/pull/35

kairstenfay: There's certainly a lot to keep track of, and in the future we may want something better than shoving everything into `__init__.py`. I can refactor the function from #35 if that PR gets merged to master before this one.

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,356 @@
...
+
+    processed_without_error = None
+
+    try:
+        for record in longitudinal:
+            with db.savepoint(f"longitudinal record {record.id}"):
+                LOG.info(f"Processing longitudinal record {record.id}")
+
+                # Check validity of barcode
+                received_sample_identifier = sample_identifier(db,
```
joverlee521: Do we always expect a sample with this longitudinal data? I thought they were doing surveys even if they don't swab the kid?

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,356 @@
...
+
+                # Skip row if no matching identifier found
+                if received_sample_identifier is None:
+                    LOG.info("Skipping due to unknown barcode " + \
+                              f"{record.document['barcode']}")
+                    mark_skipped(db, record.id)
+                    continue
+
+                # Check sample exists in database
+                sample = find_sample(db,
```
joverlee521: I'm a little confused by the order of how longitudinal data is added? When do we expect the samples for these to be created in the database?

lib/seattleflu/db/cli/command/etl/longitudinal.py
joverlee521: I think it's either yes/no or null

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,278 @@
...
+@longitudinal.command("parse")
+@click.argument("baseline_filename",
+    metavar = "<Longitudinal Childcare Data baseline (enrollments) filename>")
+@click.argument("weekly_filename",
+    metavar = "<Longitudinal Childcare Data weekly assessments filename>")
+@click.option("-o", "--output",
+    metavar="<output filename>",
+    help="The filename for the output of missing barcodes")
+
+def parse(baseline_filename, weekly_filename, output):
```
joverlee521: I would love to walk through this in person 😮

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,278 @@
...
+    df["ChildrenNearChildren"] = "yes"
+    df["SchoolType"] = "childcareCenter"
+    df["ChildrenHutchKids"] = np.where(df['enroll_site'] == 1, 'yes', 'no')
+
+    return df
+
+
+def hash_pii(df: pd.DataFrame) -> pd.DataFrame:
+    """ Hashes personally identifiable information in a given DataFrame *df*. """
+    df['study_id'] = df['study_id'].astype('str').apply(generate_hash)
```
joverlee521: I just looked at the data SCH sent us and I'm a little worried that just hashing the `study_id` alone will generate too many repeats since the `study_id` is just a number

lib/seattleflu/db/cli/command/etl/longitudinal.py
kairstenfay: Yes, you're right! I end up with a lot of missing barcodes in the end. In the current process, I'm dropping those records without barcodes. I was operating under the assumption that we only care about encounters with swabs. @tsibley -- is this the case?

lib/seattleflu/db/cli/command/longitudinal.py
kairstenfay: I was also curious if I was being "too secure" by hashing `study_id`.

lib/seattleflu/db/cli/command/longitudinal.py
kairstenfay: It would be my pleasure :smile:

lib/seattleflu/db/cli/command/longitudinal.py
joverlee521: I think we would want to keep things consistent and make the `individual.identifier` still be a 64 character string. We might just want to hash multiple things about the individual so that we don't get repeated identifiers.

kairstenfay:
I'm wavering a bit on our QC checks in the ETL commands here.
For example, I added a commit earlier for `UnknownInsuranceResponseError` in place of using `.get(, None)`. 
Then I reverted it.
The QC checks are largely for us for making sure the data are consistent and unchanging over time. What do you think? Why would we implement one `Unknown...ResponseError` for one response (e.g. race) over another?

lib/seattleflu/db/cli/command/etl/longitudinal.py
kairstenfay: Good question. I don't have a good grasp here of the process.

lib/seattleflu/db/cli/command/longitudinal.py
kairstenfay: What's a good strategy for selecting these columns?
I've reworked this with `study_id` + `race` + `census_tract`.

joverlee521:
> The QC checks are largely for us for making sure the data are consistent and unchanging over time. What do you think? Why would we implement one `Unknown...ResponseError` for one response (e.g. race) over another?

I thought we were implementing this `Unknown..Error` for things we **know** should be within a defined set

kairstenfay:
> > The QC checks are largely for us for making sure the data are consistent and unchanging over time. What do you think? Why would we implement one `Unknown...ResponseError` for one response (e.g. race) over another?
> 
> I thought we were implementing this `Unknown..Error` for things we **know** should be within a defined set

In that case, should we add custom errors for Sex and Insurance? We don't currently have those implemented in clinical.

kairstenfay:
[Ingest longitudinal childcare enrollments](https://trello.com/c/DYO4a8cn/97-ingest-longitudinal-childcare-enrollments)

lib/seattleflu/db/cli/command/etl/longitudinal.py
tsibley: Ah, no, we do care about encounters without swabs. They'll have encounters with details, which may still be useful.

kairstenfay: Ok, that's going to require some changes. I'll get started on those.

lib/seattleflu/db/cli/command/etl/longitudinal.py
tsibley: Yep, I think these samples end up being processed by Helen's lab and recorded in the manifest as normal.

lib/seattleflu/db/cli/command/etl/longitudinal.py
kairstenfay: Done with those changes now. I'm not skipping on missing barcode or missing sample, but only on missing encounter date.

lib/seattleflu/db/cli/command/longitudinal.py
tsibley: I'm not sure we need to hash anything here, although we do need to add additional bits of entropy/information to make it unique in a wider space of values.

Stepping back, the goal is a consistent identifier for the individual, which `study_id` already is. As Jover pointed out, the problem with using that alone is that it is unqualified. We could solve this by namespacing it with static values, for example: `redcap/sch/year-1/childcare/<study_id>`. This is similar (but with a longer namespace prefix) to what we did with the NWGC internal ids we use as identifiers for the presence/absence data (`NWGC_<id>`).

tsibley:
> I thought we were implementing this `Unknown..Error` for things we **know** should be within a defined set

The guideline I use is to limit most custom error classes to things that other code I'm writing wants to catch in a try/except block. If the error is only ever going to propagate upwards until it aborts execution, then I tend to use `raise Exception(…)` or `assert`.

kairstenfay:
FYI I ended up not having to skip null encounters at all. I previously had a bug in my code that was adding too many incomplete rows which is why I was getting null encounters. Now that the  bug is fixed, null encounters no longer happen

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,303 @@
...
+        raise UnknownEthnicGroupError(f"Unknown ethnic group «{ethnic_group}»")
+
+    return [ethnic_map[ethnic_group]]
+
+
+def flu_shot(flu_shot_response: str) -> list:
+    """
+    Given a *flu_shot_response*, returns a 'yes', 'no', or 'doNotKnow' value for
+    FluShot key.
+    """
```
kairstenfay: FYI  I noticed that we don't have consistent coding for some of these variables. For example, values for both 'doNotKnow' and 'dontKnow' exist in the db.

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,355 @@
...
+    logically equivalent values as the original DataFrame.
+    """
+    stubset = df.filter(regex=f"{'|'.join(pid)}|{stub}_")
+    if len(list(stubset)) == len(pid):
+        LOG.warning(f"Stub {stub} not found in data")
+        return None
+
+    long_stubset = pd.wide_to_long(stubset, stub, i=pid, j=f"new_{stub}",
+                                   sep="_", suffix='\\w+'
+                                  ).reset_index()
```
kairstenfay: I don't understand why this `wide_to_long` fails without the provided `suffix` argument. The same function works fine without `suffix` when casting longitudinal columns from `wide_to_long`. I feel like adding a "it works don't touch" inline comment here :grimacing:

lib/seattleflu/db/cli/command/etl/longitudinal.py
tsibley: Yeah, this is because we haven't modeled vaccinations yet in the database but are just using whatever Audere handed us in the enrollment document.

We'll address this inconsistency when we start handling vaccinations.

lib/seattleflu/db/cli/command/etl/__init__.py
tsibley: We definitely want something better in the future, but for now, let's leave these functions in `__init__.py`.

tsibley:
I re-pushed this branch to drop the original first commit that moved `generate_hash` from  `seattleflu.db.cli.command.clinical` into `seattleflu.db.cli.command`, since that function is no longer used outside of the `clinical` command. If the function is refactored in the future, we need to consider if the shared uses then make sense with the use elsewhere (e.g. we shouldn't be using the participant deidentification secret in hashes unrelated to deidentification).

kairstenfay:
FYI @tsibley - I realized that I want to defend against bad `wide_to_long` results in this PR. I plan on adding an assertion that the `week` column after reshaping longitudinal data from wide-to-long can be cast to integer. If this fails, then there was some problem in capturing stubnames. I will add this on Monday or Tuesday of next week.

tsibley:
@kairstenfay Sounds good! I'll have some other review comments posted later today as well, ready for you look at Monday or Tuesday.

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,357 @@
...
+    receive suffixes of '_followup' and '_survey' after merging, respectively.
+    """
+    pid = ['study_id']  # This harcoding may need to be updated in the future
+    merge_method = 'left'
+
+    for combo in combinations([baseline_records, weekly_records, survey_records], 2):
+        duplicated_columns = set(list(combo[0])).intersection(list(combo[1]))
+
+        assert len(duplicated_columns) == len(pid) + 1, \
+            f"You have duplicated columns other than {pid} in your data."
```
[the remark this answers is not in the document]
kairstenfay: I hadn't thought of that edge case. Great idea.

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,357 @@
...
+
+        reshaped_data = reshaped_data.merge(true_values, how='left', on=pid)
+
+    assert len(df) == len(reshaped_data), f"You do not have a 1:1 merge on {pid}"
+
+    return reshaped_data
+
+
+def collapse_wide_stubbed_columns(df: pd.DataFrame, stub: str,
+                                  pid: str) -> pd.DataFrame:
```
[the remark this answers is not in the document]
kairstenfay: Oh right, I changed it mid-way in development.

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,303 @@
...
+                sample = find_sample(db,
+                    identifier = received_sample_identifier)
+
+
+                # Most of the time we expect to see existing sites so a
+                # select-first approach makes the most sense to avoid useless
+                # updates.
+                site = find_or_create_site(db,
+                    identifier = site_identifier(record.document["site"]),
+                    details    = {"type": "childcare"})
```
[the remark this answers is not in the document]
kairstenfay: Right. We're also doing this in clinical. Would you like me to refactor there as well?

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+                # Most of the time we expect to see new individuals and new
+                # encounters, so an insert-first approach makes more sense.
+                # Encounters we see more than once are presumed to be
+                # corrections.
+                individual = upsert_individual(db,
+                    identifier  = record.document["individual"],
+                    sex         = sex(record.document["AssignedSex"]))
```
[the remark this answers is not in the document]
kairstenfay: This is how we've done it in the past with the `clinical` etl pipeline. Would you suggest a refactor there as well?

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,303 @@
...
+                # Check sample exists in database
+                sample = find_sample(db,
+                    identifier = received_sample_identifier)
+
+
+                # Most of the time we expect to see existing sites so a
+                # select-first approach makes the most sense to avoid useless
+                # updates.
+                site = find_or_create_site(db,
+                    identifier = site_identifier(record.document["site"]),
```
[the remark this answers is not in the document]
kairstenfay: Can do. Refactor in `clinical`, too?

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    Given a *race_name*, returns the matching race identifier found in Audere
+    survey data. Converts multiple races to 'other'.
+    """
+    if races is None:
+        LOG.debug("No race response found.")
+        return [None]
+
+    if len(races) > 1:
+        return 'other'
```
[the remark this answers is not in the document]
kairstenfay: I disliked this logic, too. But I wanted to map as closely as possible to `clinical`. There, 'MultipleRaces' gets coded as 'other'.

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,303 @@
...
+                "home": {
+                    "region": document.get("census_tract"),
+                }
+            },
+            "responses": {
+                "Race": race(document.get("Race")),
+                "FluShot": flu_shot(document.get("FluShot")),
+                "AssignedSex": [sex(document.get("AssignedSex"))],
+                "HispanicLatino": hispanic_latino(document.get("HispanicLatino")),
+                "MedicalInsurance": insurance(document.get("MedicalInsurance"))
```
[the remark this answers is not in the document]
kairstenfay: Ditto for should I overhaul `clinical` as well? :slightly_smiling_face:

lib/seattleflu/db/cli/command/etl/longitudinal.py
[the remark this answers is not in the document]
tsibley: Sure, as a separate commit. But it would also be ok to leave `clinical` alone for now.

lib/seattleflu/db/cli/command/etl/longitudinal.py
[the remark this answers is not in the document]
tsibley: Ditto, as above.

lib/seattleflu/db/cli/command/etl/longitudinal.py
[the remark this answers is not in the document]
tsibley: I think `MultipleRaces` → `other` is appropriate because in that case we don't know what they are. But if we have actual multiple race values, then we can map them into the Audere value set.

lib/seattleflu/db/cli/command/etl/longitudinal.py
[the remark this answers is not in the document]
tsibley: Your discretion, as above.

kairstenfay:
Found a bug while incorporating code review requested changes.
Not sure what was causing it, but now the program is no longer dropping `null` encounter dates automatically.

This is what I expected from the beginning, so we're heading in a good direction. 

However, it leads me to this problem:
In development, I saw there was a provided date for both weekly (in-person) and survey (parent-submitted) data. I arbitrarily chose the weekly data over the survey timestamp for our 'encountered' date.

We now have a few records that have no weekly `encountered` date but do have some metadata from the parent survey from the same week.
Thoughts around how to proceed? @tsibley @joverlee521 

I'll be sure to document my choice, but what I'm thinking is filling in the blank `encountered` date with the parent survey timestamp (if it exists). Otherwise, the whole record for that week should be dropped.

kairstenfay:
> Found a bug while incorporating code review requested changes.
> Not sure what was causing it, but now the program is no longer dropping `null` encounter dates automatically.
> 
> This is what I expected from the beginning, so we're heading in a good direction.
> 
> However, it leads me to this problem:
> In development, I saw there was a provided date for both weekly (in-person) and survey (parent-submitted) data. I arbitrarily chose the weekly data over the survey timestamp for our 'encountered' date.
> 
> We now have a few records that have no weekly `encountered` date but do have some metadata from the parent survey from the same week.
> Thoughts around how to proceed? @tsibley @joverlee521
> 
> I'll be sure to document my choice, but what I'm thinking is filling in the blank `encountered` date with the parent survey timestamp (if it exists). Otherwise, the whole record for that week should be dropped.

It seems nearly all of the missing `encountered` dates occur at week 16. There is no follow-up after week 16, so there may be no need to hold onto parent survey metadata. Perhaps all of week 16 could be dropped. Thoughts on this?

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,357 @@
...
+
+def group_true_values_into_list(long_subset: pd.DataFrame, stub: str,
+                                pid: str) -> pd.DataFrame:
+    """
+    Given a long DataFrame *long_subset*, collapses rows with the same *pid*
+    such that every *pid* is represented once in the resulting DataFrame. True
+    values (1) for each category in the given *stub* are collapsed into a
+    human-readable list.
+    """
+    true_subset = long_subset[long_subset[stub] == 1]
```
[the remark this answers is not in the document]
kairstenfay: I tried casting to bool using two different methods. Each provoked a SettingWithCopyWarning from pandas.
`long_subset[stub] = long_subset[stub].apply(lambda x: bool(x))`
`long_subset.loc[:, stub] = long_subset[stub].astype('bool')`

Do you know a good way around this? For some reason, the `trim_whitespace` function doesn't trigger this warning.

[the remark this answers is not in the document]
kairstenfay: Hmm I think it has to do with the `dropna` statement I added directly above it. I'll keep looking into it.

lib/seattleflu/db/cli/command/etl/__init__.py
```diff
@@ -140,6 +141,72 @@ def upsert_encounter(db: DatabaseSession,
...
+    warehouse.identifier.
+
+    Will be sample barcode if from UW and collection barcode if from SCH.
+    """
+    identifier = find_identifier(db, barcode)
+
+    if identifier:
+        assert identifier.set_name == "samples" or \
+            identifier.set_name == "collections-seattleflu.org", \
+            f"Identifier found in set «{identifier.set_name}», not «samples»"
```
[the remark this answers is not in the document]
kairstenfay: It looks like longitudinal records are only from `collections-seattleflu.org`, so this assertion that includes `samples` also should not be shared.

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -0,0 +1,303 @@
...
+                # updates.
+                site = find_or_create_site(db,
+                    identifier = site_identifier(record.document["site"]),
+                    details    = {"type": "childcare"})
+
+
+                # Most of the time we expect to see new individuals and new
+                # encounters, so an insert-first approach makes more sense.
+                # Encounters we see more than once are presumed to be
+                # corrections.
```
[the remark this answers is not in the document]
kairstenfay: technically, no. The wording is wrong. most of the time we expect to see existing individuals and new encounters. 
How would this distinction impact the code? I don't currently see how it makes a difference (after correcting the comment). Individuals still should be created first, right?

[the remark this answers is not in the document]
tsibley: Yep, individuals do need to exist before their encounters. What I was getting at here with insert-first is whether the ETL performs a "select, if not exists insert" (find-or-create) vs. an "insert, on conflict update" (upsert). There are some tradeoffs and side-effects of one way vs. the other, but they may not really matter for us at the moment. Updating the comment is enough for me. :-)

kairstenfay:
I re-did the upsert individual to find-or-create individual. Currently it's adding `null` values to the details column. Is this desired?

tsibley:
Explicit nulls in the details data are fine. This lets someone see the expected structure of the details data even if the record they're examining is missing values.

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,403 @@
...
+    survey_records = rename_stub(survey_records, regex='^household_(?!ill)',
+                                 current_stub='household',
+                                 desired_stub='household_sx')
+
+    return merge_data(baseline_records, weekly_records, survey_records)
+
+
+def load_data(filename: str) -> pd.DataFrame:
+    """
+    Loads longitudinal childcare data from SCH from a given *filename.
```
tsibley: Missing `*` after `*filename`.

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,403 @@
...
+    Given three DataFrames (*baseline_records*, *weekly_records*, and
+    *survey_records*) of longitudinal childcare data, merges them on the unique
+    person identifier (pid).
+
+    Raises a :class:`AssertionError` if the pid columns are missing in any
+    DataFrame or if there are duplicated columns between any two DataFrames
+    other than the pid and '_metadata' columns (added by a previous process).
+
+    The '_metadata' columns in *weekly_records* and *survey_records* receive
+    suffixes of '_followup' and '_survey' after merging, respectively.
```
tsibley: FWIW, the ReStructured Text syntax for a literal bit of code is double backticks, e.g.

```
suffixes of ``_followup`` and ``_survey``
```

lib/seattleflu/db/cli/command/longitudinal.py
```diff
@@ -0,0 +1,403 @@
...
+
+        reshaped_data = reshaped_data.merge(true_values, how='left', on=pid)
+
+    assert len(df) == len(reshaped_data), f"You do not have a 1:1 merge on {pid}"
+
+    return reshaped_data
+
+
+def collapse_wide_stubbed_columns(df: pd.DataFrame, stub: str,
+                                  pid: [str]) -> pd.DataFrame:
```
tsibley: FYI, I think you want `List[str]` here, but again, I know this isn't clear because we're not validating these type annotations yet.

schema/sqitch.plan
```diff
@@ -82,6 +82,7 @@ roles/dumper/create [receiving/schema warehouse/schema shipping/schema] 2019-06-
 roles/dumper/grants [roles/dumper/create] 2019-06-04T20:59:22Z Thomas Sibley <tsibley@fredhutch.org> # Grants to dumper role
 @2019-06-04-join-shipping-views 2019-06-04T23:35:07Z Jover Lee <joverlee@fredhutch.org> # Schema as of 04 June 2019
 shipping/views [shipping/views@2019-06-04-join-shipping-views] 2019-06-04T23:36:16Z Jover Lee <joverlee@fredhutch.org> # Create new view to join modeling and presence/absence views
+receiving/longitudinal [receiving/schema] 2019-06-11T22:46:10Z Kairsten Fay <kfay@fredhutch.org> # Receiving table for longitudinal data
```
tsibley: This line should move to the bottom of the plan so it deploys cleanly.

PR #37, opened 2019-06-17
Strip whitespace from values when parsing

kairstenfay:
See https://trello.com/c/ZIHacQh9/133-strip-leading-trailing-whitespace-from-values-when-parsing-fields


kairstenfay:
[Strip leading/trailing whitespace from values when parsing fields](https://trello.com/c/ZIHacQh9/133-strip-leading-trailing-whitespace-from-values-when-parsing-fields)

tsibley:
@kairstenfay Would you rebase the sole commit for this PR onto master and re-push so it's not attached to your longer `ingest_longitudinal_childcare_enrollments` branch? You can then rebase that branch onto this one (instead of vice versa, like now). It's fine if you don't want to do this now, but I'll then wait to review this until your other branch is settled and merged. Ok by you?

kairstenfay:
@tsibley done. I'll just need to add the whitespace trimming to the longitudinal process once that PR is merged.

lib/seattleflu/db/cli/command/__init__.py
```diff
@@ -9,3 +10,12 @@
     "clinical",
     "user",
 ]
+
+
+def trim_whitespace(df: pd.DataFrame) -> pd.DataFrame:
+    """ Trims leading and trailing whitespace from strings in *df* """
+    str_dtypes = df.select_dtypes(include='object')
+
+    df[str_dtypes.columns] = str_dtypes.apply(lambda column: column.str.strip())
```
tsibley: The `object` dtype also catches things which aren't strings, such as if we had sets or lists or other objects in a column. When `column.str.strip()` is applied to a object-but-non-string column, NaNs are returned. For example:

```python
>>> d = pd.DataFrame({"a":["one","two  ", "  three"], "b":[{1,2,3},{3,4,5},{7,8,9}], "c":[1,2,3]})
>>> d
         a          b  c
0      one  {1, 2, 3}  1
1    two    {3, 4, 5}  2
2    three  {8, 9, 7}  3
>>> d.dtypes
a    object
b    object
c     int64
dtype: object
>>> d.select_dtypes(include = 'object').apply(lambda c: c.str.strip())
       a   b
0    one NaN
1    two NaN
2  three NaN
```

It would be more robust and more defensive to filter the selected columns to just those which are a subclass of `str`. I think something like the following would do it:

```python
str_columns = df.columns[df.applymap(lambda x: isinstance(x, str) or pd.isna(x)).all()]

df[str_columns] = df[str_columns].apply(lambda column: column.str.strip())
```

kairstenfay: Good catch! This was something I noticed in the docs and ran into when implementing, but I (dangerously) assumed we were 'safe' from having to worry about this.

lib/seattleflu/db/cli/command/clinical.py
```diff
@@ -120,27 +120,26 @@ def load_data(uw_filename: str, uw_nwh_file: str, hmc_sch_file: str):
...
-
-    hmc_manifest = load_manifest_data(hmc_sch_file, 'HMC')
-    hmc_manifest = hmc_manifest.rename(columns={'Barcode ID (Sample ID)': 'Barcode ID'})
+    uw_manifest = load_manifest_data(uw_nwh_file, 'UWMC',
+                                     'Barcode ID (Sample ID)',
+                                     'Collection Date')
+    nwh_manifest = load_manifest_data(uw_nwh_file, 'NWH',
+                                      'Barcode ID (Sample ID)',
+                                      'Collection Date (per tube)')
+    hmc_manifest = load_manifest_data(hmc_sch_file, 'HMC', 'Barcode ID',
```
tsibley: I believe `'Barcode ID'` should remain `'Barcode ID (Sample ID)'`.

lib/seattleflu/db/cli/command/__init__.py
```diff
@@ -9,3 +10,18 @@
...
+
+    return df
+
+
+def every_value_is_str_or_na(df):
+    """
+    Evaluates whether every value in the columns of a given DataFrame *df* is
+    either a string or NA.
+    """
+    return df.applymap(lambda col: isinstance(col, str) or pd.isna(col)).all()
```
tsibley: I would prefer to keep these functions local to the code using them until we have a need to use them elsewhere.

kairstenfay:
Good point! I'm adding those changes now.

PR #39, opened 2019-07-12
Incidence model views: Correct handling of encounters with multiple (or no) samples

tsibley:
The lateral subquery had two flaws which only became apparent with
Jover's self-test kit work that greatly increases the number of
encounters with multiple samples (RDT and UTM).

 1. A misunderstanding of lateral subqueries meant that encounters
    without a sample were being omitted because the join was inner
    not left.

 2. A bug in the use of the first_value() window function meant that
    encounters with multiple samples produced wholly duplicate rows in
    the view (one row per sample).  The intent, as noted in my original
    commit message, was to return a _single_ row for each encounter,
    with an arbitrarily-picked "first" sample.

Since the modeling team was already having to deal with multiple rows
for some encounters, correct the flawed queries in both v1 and v2 views
to report the left join of encounters with samples.  This adds in the
previously missing encounters and reports the actual sample identifier
for the previously duplicated encounters.


tsibley:
[Incidence model view needs to handle encounters with multiple samples](https://trello.com/c/mcVGO2Vw/155-incidence-model-view-needs-to-handle-encounters-with-multiple-samples)

tsibley:
Yep. And all the samples are reported in the presence/absence view, so the join of those results with the encounters/observations there will leave only the relevant samples.

joverlee521:
Okay, my one thought was whether or not it would somehow interrupt augur-build, but just confirmed with Cassia and Miguel that they specifically look at the sample anyways.

tsibley:
Ah, yes, good call on checking that.

tsibley:
@famulare @devclinton FYI, this bug fix I just deployed affects the views you're using. The important changes are:

1. Encounters without a sample that were previously omitted are now included
2. Encounters with multiple samples which were previously wholly duplicated now report each individual sample identifier.

tsibley:
I don't think this should fundamentally break anything for you, but let us know if you have problems.

famulare:
Thanks for the heads up. It looks like everything will run fine, and I don't yet have the diagnostics in place to see if it changed anything important, so I suppose we're good with it.

PR #42, opened 2019-07-16
etl: move `SampleNotFoundError` to shared module

kairstenfay:
Move the `SampleNotFoundError` from the `presence_absence` ETL routine
to a shared module. The `clinical` and `longitudinal` ETL routines were
previously importing it from `presence_absence`; this new, shared
location is more sensible.

### DIFF
--- a/lib/seattleflu/db/cli/command/etl/__init__.py
+++ b/lib/seattleflu/db/cli/command/etl/__init__.py
@@ -243,6 +243,13 @@ def find_sample(db: DatabaseSession, identifier: str) -> Any:
     return sample
 
 
+class SampleNotFoundError(ValueError):
+    """
+    Raised when a function is unable to find an existing sample with the given
+    identifier.
+    """
+    pass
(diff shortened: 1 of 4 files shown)


lib/seattleflu/db/cli/command/etl/__init__.py
```diff
@@ -243,6 +243,13 @@ def find_sample(db: DatabaseSession, identifier: str) -> Any:
     return sample
 
 
+class SampleNotFoundError(ValueError):
+    """
+    Raised when :func:``update_sample`` is unable to find an existing
```
joverlee521: Might want to change doc string? There is an `update_sample` function within `__init__.py` that is does not raise this error

kairstenfay: Good catch

kairstenfay:
Updated the docstring for `SampleNotFoundError`

PR #43, opened 2019-07-17
Genomic data

joverlee521:
Creates 3 different tables within warehouse to hold genomic data:
1. `sequence_read_set`: urls of external datastore for a set of FASTQ files for a sample from a sequencing run. This table has 3 triggers to ensure that the urls are not null, unique within a set, and unique only to one set. See commit messages for further details.

1. `genome`: an explicit representation of a whole genome of an organism that was extracted from one sequence read set of a sample. If the genome is from an external source with an unknown origin, then the sequence read set can be left null.

1. `genomic_sequence`: a sequence that belongs to a genome. There can be multiple sequences per genome for segmented genomes (e.g. Influenza).


trvrb:
I don't have concrete things to suggest here, but somethings to think about:

* Might be good semantically to explicitly call this `consensus_genome`. We'd ideally eventually have another concept of `single_nucleotide_variant` that is also attached to a sample. This may require a grouping of `single_nucleotide_variant_set`. I don't know exactly how to do this, but it will eventually be relevant.

* If we want to record something like `single_nucleotide_variant`, we need a Genbank reference (like https://github.com/seattleflu/augur-build/blob/master/config/reference_h3n2_ha.gb) to maintain a reference frame. The concept of a reference genome is more specific than the concept of an organism, ie organism could be `Influenzavirus A`, while reference could be `U26830` (unique Genbank accession).

joverlee521:
[Add sequence read sets and genomes to schema](https://trello.com/c/lsVpaTiT/143-add-sequence-read-sets-and-genomes-to-schema)

tsibley:

PR #44, opened 2019-07-17
Create kit processor role for automation

joverlee521:
Creates kit process role in the database via sqitch
to allow the automation of kit etl jobs. 

There was no need to create a `kit uploader` role
because the kit etl pulls data from `receiving.manifest` 
and `receiving.enrollment` which already have their
own uploaders.

### DIFF
--- a/schema/deploy/roles/kit-processor/create.sql
+++ b/schema/deploy/roles/kit-processor/create.sql
@@ -1,0 +1,9 @@
+-- Deploy seattleflu/schema:roles/kit-processor/create to pg
+
+begin;
+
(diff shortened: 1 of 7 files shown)


joverlee521:
[Create automation user for self-test kits](https://trello.com/c/xlyky1CA/168-create-automation-user-for-self-test-kits)

kairstenfay:
Looks good! Did you test this role locally?

joverlee521:
@kairstenfay Yup! Everything ran as it should with a test user granted the `kit-processing` role.

PR #46, opened 2019-07-18
Web api endpoint to receive red cap data entry triggers

kairstenfay:
Created `receiving.redcap_det`
Created POST endpoing `/v1/receiving/redcap-det`

Unsure if I need to create a DB role for this (seems likely).


kairstenfay:
[Web API endpoint to receive REDCap data entry triggers](https://trello.com/c/z9XAiDMN/164-web-api-endpoint-to-receive-redcap-data-entry-triggers)

kairstenfay:
I added a `redcap-det-uploader` role. To fully incorporate this into production, we'll need to create a user for REDCap on the production db and grant it this role. We will then have to update the REDCap DET web endpoint on the server, including the new REDCap username & password in basic authentication format.

tsibley:
Database and web API changes are deployed. I haven't setup a production user or modified our production REDCap projects. That will be a separate task to do later when we're closer to having those finalized.

PR #48, opened 2019-07-23
Create id3c web api endpoint for consensus genomes

kairstenfay:
Created a receiving table, uploader role, and web API endpoint for assembly. See seattleflu/assembly/pull/7. 

Created off of @joverlee521's https://github.com/seattleflu/id3c/tree/genomic_data branch

PR #49, opened 2019-07-25
Consensus genome etl

kairstenfay:
Continues the work from
#43  and #48 

to create a Consensus Genome ETL process.


kairstenfay:
[ETL routine for received consensus genomes from assembly](https://trello.com/c/MNKd7kqX/174-etl-routine-for-received-consensus-genomes-from-assembly)

joverlee521:
I also had a thought that we need to somehow update the `sequence_read_set.details` if the assembly job fails. Not really sure how to do this, would love to brainstorm ideas.

lib/seattleflu/db/cli/command/etl/consensus_genome.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def lineage(db: DatabaseSession, document: dict) -> str:
+    """
+    Given a *document* containing a reference organism, returns its matching
+    lineage.
+    """
+    organism_name = document.get("reference_organism")
+
+    organism_map = {
```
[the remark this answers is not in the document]
kairstenfay: Brilliant

kairstenfay:
> I also had a thought that we need to somehow update the `sequence_read_set.details` if the assembly job fails. Not really sure how to do this, would love to brainstorm ideas.
> 
> This may need to add to [seattleflu/assembly#7](https://github.com/seattleflu/assembly/pull/7) to upload to `receiving.consensus_genome` or a different receiving table if the job had failed at a specific step in the pipeline.

Interesting idea, and plausible, but potentially tricky with the Snakemake rules posting to ID3C being dependent on the output of the consensus genome + summary stats. Will you make this into a Trello card @joverlee521 ?

lib/seattleflu/db/cli/command/etl/consensus_genome.py
```diff
@@ -0,0 +1,363 @@
...
+    # Postgres allows array columns to be defined as unique, but the ordering of
+    # the arrays must be the same for the arrays to match. We are only
+    # interested in matching on array content, not array ordering. This prevents
+    # us from being able to use the built-in unique constraint on URLs, thereby
+    # preventing `ON CONFLICT` updates to a table when URLs is the only unique
+    # column in the table.
+    with db.cursor() as cursor:
+        cursor.execute("""
+            update warehouse.sequence_read_set
+              set details = %(log_entry)s
```
joverlee521: Might want to change this to `set details = coalesce(details, '{}') || %(log_entry)s` so that the previous details do not get overwritten.

kairstenfay: Assuming the log entry has the same keys, because the coalescing happens at the top level of the Json object, won't the items with the same keys get overwritten anyway?

joverlee521: Hmm that's true, then maybe we want to make `sequence_read_set.details` a nested json somehow?

lib/seattleflu/db/cli/command/etl/consensus_genome.py
```diff
@@ -0,0 +1,363 @@
...
+
+        data = {
+            "sample_id": sample.id,
+            "urls": urls,
+        }
+
+        sequence_read_set = db.fetch_row("""
+            insert into warehouse.sequence_read_set (sample_id, urls)
+                values (%(sample_id)s, %(urls)s)
+            returning sequence_read_set_id as id
```
joverlee521: Also needs to return `sample_id` or else func:`upsert_genome` errors out

lib/seattleflu/db/cli/command/etl/consensus_genome.py
```diff
@@ -0,0 +1,363 @@
...
+                # unique constraints on array columns. Therefore, perform a
+                # find-or-create followed by an update to the details column to
+                # avoid conflict.
+                sequence_read_set = find_or_create_sequence_read_set(db,
+                    record.document, sample)
+
+                mark_received(db,
+                    sequence_read_set.id,
+                    {   "received": record.received,
+                        "consensus_genome_id": record.id    })
```
joverlee521: "consensus_genome_id" might be a little misleading here? One might think it's the `genome_id` is `warehouse.genome`

kairstenfay: I thought it might be useful to tie in the consensus genome id from `receiving.consensus_genome`. Do you have a suggestion for a better name, or should this idea be abandoned?

lib/seattleflu/db/cli/command/etl/consensus_genome.py
kairstenfay: What are you interested in tracking in the details column? I think if I better understand your intended use case, I can design something that makes more sense.

lib/seattleflu/db/cli/command/etl/consensus_genome.py
kairstenfay: Thank you, forgot to test that.

joverlee521:
Something like:
{
      "<target/reference tested>": { 
           "status":"complete/failed/notmapped"
       }
}

lib/seattleflu/db/cli/command/etl/consensus_genome.py
kairstenfay: I updated the `details` column for sequence read set to your suggestion. Now that we have different top-level keys depending on the target organism, I changed the SQL to use coalesce.

kairstenfay:
Right now, if `etl consensus-genome` is run, the sequence read set details column status for a target/reference is set to `success`. This is currently the only status that is set.

joverlee521:
The consensus genome etl can handle both completed assembly jobs (including consensus genome and sequences) and failed assembly jobs(i.e. not mapped sample/reference pairs).

tsibley:

PR #50, opened 2019-07-26
Backfill genomic data

joverlee521:
This PR includes #43 and #48 

Parse and upload sequence read sets to `warehouse.sequence_read_set`.
- This should be a temporary solution of filling in sequence read sets, until we can have the sequence read sets be sent to the API endpoint into `receiving.sequence_read_set` then run an "etl" process that would transfer them from Globus to Rhino and then create sequence read sets within `warehouse.sequence_read_set` with the appropriate Rhino file paths as the urls. 

Upload consensus genomes/summary data to `receiving.consensus_genome`.
- This requires the ndjson file generated from steps listed in https://github.com/seattleflu/assembly/pull/11


lib/seattleflu/db/cli/command/consensus_genome.py
```diff
@@ -0,0 +1,37 @@
...
+import json
+from seattleflu.db.cli import cli
+from seattleflu.db.session import DatabaseSession
+
+
+LOG = logging.getLogger(__name__)
+
+
+@cli.command("upload-consensus-genome", help=__doc__)
+@click.argument("consensus_genome-file",
```
kairstenfay: Naming nitpick: I would suggest changing this to `consensus-genome upload` to match the command naming conventions used for `clinical`, `longitudinal`, etc.

tsibley:

PR #52, opened 2019-08-08
db.session: Assert fetch_row() only finds one row

tsibley:
This is a typical sanity check since the query is expected to return at
most 1 row.  More rows indicate a likely programming error or unexpected
non-uniqueness condition.

I'm not sure why I forgot this check when I wrote the function in the
first place.  It's been a standard practice of mine for years.


tsibley:
I applied this commit to master.

PR #53, opened 2019-08-13
Kit results

joverlee521:
* Parse kit results from the Self-test manifest and print as ndjson
* Upload to `kit-result.ndjson` to `receiving.kit_result`
* Add etl kit-result to add kit results to `warehouse.presence_absence`

I'm not super confident in the parsing step. The results column in the self-test manifest all have the same name regardless of if it was from the test-strip, RDT, or UTM sample, so I had to work around that. Would love to know if there is a better way to do this!


joverlee521:
[Load Cepheid presence/absence results for Flu@Home samples](https://trello.com/c/hRGwBKgC/57-load-cepheid-presence-absence-results-for-fluhome-samples)

joverlee521:
With approval from Helen's lab, we renamed the results column within the specimen manifest on OneDrive so that we can differentiate between the different sample results. The kit-result parse step has now been updated to reflect these changes.

One thought I had was we need some way to record/diff the kit results just as we do for specimen manifests. I think we can use `manifest diff` on kit-results if we also upload kit_result.ndjson to a github repo (this can be a new repo or just the speciment-manifest repo).

lib/seattleflu/db/cli/command/kit_result.py
```diff
@@ -0,0 +1,206 @@
...
+    Will drop any records that have NA or duplicated values for the given *barcode_columns*.
+    """
+    replacements = {
+        pd.np.nan: None,
+        "": None,
+        "NA": None,
+        "N/A": None,
+        "?": None
+    }
+
```
kairstenfay: I'm curious here: why did you want to convert pandas `NaN` to `None`?

lib/seattleflu/db/cli/command/kit_result.py
```diff
@@ -0,0 +1,206 @@
...
+        r"Cepheid Test date":  "test_date",
+        r"RDT Barcode":        "barcode",
+        r"UTM Sample Barcode":    "barcode",
+        r"Strip Barcode":      "barcode",
+    }
+
+    column_map = {}
+
+    for column in columns.keys():
+        pattern = re.compile(column, re.IGNORECASE)
```
kairstenfay: Oh nice, I didn't know about `re.IGNORECASE`

lib/seattleflu/db/cli/command/etl/kit_result.py
```diff
@@ -0,0 +1,349 @@
...
+                        assert sample.id in {kit.rdt_sample_id, kit.utm_sample_id}, \
+                            (f"Found sample {sample.id} does not match samples " +
+                             f"linked to kit: {kit.rdt_sample_id, kit.utm_sample_id}")
+
+                        # Loop through test results
+                        for test_result in kit_result.document:
+
+                            target_result = get_target_result(kit_result.document[test_result])
+
+                            if target_result is None:
```
kairstenfay: At this point we are now three conditionals deep. I think there may be a way to rewrite the code to reduce indentation a bit. 

For example, on line 123, after we mark the kit_result as skipped if it doesn't have a sample_identifier, then we can `continue` (like you use in lines 139-141).

lib/seattleflu/db/cli/command/kit_result.py
joverlee521: I think `NaN` gets converted to a string "NaN" at some point and would not be dropped with the other null values.

lib/seattleflu/db/cli/command/__init__.py
```diff
@@ -104,3 +105,35 @@ def dump_ndjson(df):
...
+            LOG.warning(f"Found duplicate barcodes in column «{column}»")
+            dup_barcodes = duplicates.unique().tolist()
+            LOG.warning(f"Duplicated barcodes: {dup_barcodes}")
+            LOG.warning(f"Dropping records with duplicate barcodes")
+            deduplicated_df = df[(~df[column].duplicated(keep=False)) \
+                                | (df[column].isnull())][column].to_frame()
+            common_idx = deduplicated.index.intersection(deduplicated_df.index)
+            deduplicated = deduplicated.loc[common_idx]
+
+    return deduplicated
```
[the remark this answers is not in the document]
joverlee521: I believe `barcode_quality_control` does not actually drop the duplicated barcodes from the original Dataframe. `qc_barcodes` drops the barcodes and warns about it, but does not raise an assertion error.

lib/seattleflu/db/cli/command/kit_result.py
```diff
@@ -0,0 +1,206 @@
...
+    kit_result records with known identifiers and existing kits/samples.
+    """
+    db = DatabaseSession()
+
+    try:
+        LOG.info(f"Copying kit result records from {kit_result_file.name}")
+
+        row_count = db.copy_from_ndjson(("receiving", "kit_result", "document"), kit_result_file)
+
+        LOG.info(f"Received {row_count:,} kit result records")
```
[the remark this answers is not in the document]
tsibley: It formats the number with commas as the thousands separator.

Relevant Python documentation is a little hard to find and dense, but these two links contain all the gory details:

* https://docs.python.org/3/library/string.html#format-string-syntax
* https://docs.python.org/3/library/string.html#formatspec

joverlee521:
Closing since we've never been asked to ingest kit results from Year 1.

PR #54, opened 2019-08-20
sequence-read-set parse: Take a list of FASTQs instead of a directory

tsibley:
Pushing the globbing into the hands of the user running the command is
more flexible and robust to unexpected situations.  For example, if we
have uncompressed FASTQs, the previous glob wouldn't have found them.
As another example, one command invocation can now be run on FASTQs in
separate directories.

### DIFF
--- a/lib/seattleflu/db/cli/command/sequence_read_set.py
+++ b/lib/seattleflu/db/cli/command/sequence_read_set.py
@@ -25,9 +25,9 @@ def sequence_read_set():
 
 @sequence_read_set.command("parse")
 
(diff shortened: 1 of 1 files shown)


tsibley:
That's fair! Using shell globbing instead of globbing in the command does seem like it would make us more prone to making that mistake of collapsing multiple sets for the same sample. I wonder about passing the command multiple directories instead and having it know to keep them separate, but we don't need to do that right now.

Let's close this PR without merging. We can add a `--file-format` option later if we need it. Thanks for your feedback!

PR #57, opened 2019-08-27
Shipping view for flu assembly jobs

joverlee521:
Creates shipping view for all Flu related assembly jobs that have not been completed. Eventually this will be used to automate the assembly pipeline. For now it can be used to check which jobs haven't been completed manually.

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -121,7 +121,7 @@ comment on view shipping.presence_absence_result_v1 is
 revoke all
     on shipping.presence_absence_result_v1
   from "incidence-modeler";
-  
+
 grant select
     on shipping.presence_absence_result_v1
     to "incidence-modeler";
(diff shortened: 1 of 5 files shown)


schema/deploy/shipping/views.sql
```diff
@@ -225,4 +225,33 @@ create or replace view shipping.observation_with_presence_absence_result_v1 as
...
+             else o1.lineage
+           end as assembly_job_organism,
+           sequence_read_set.details
+
+      from warehouse.sequence_read_set
+      join warehouse.sample using (sample_id)
+      join warehouse.presence_absence using (sample_id)
+      join warehouse.target using (target_id)
+      join warehouse.organism o1 using (organism_id)
+      left join warehouse.organism o2 on (o2.lineage ~ concat(o1.lineage::text, '.*{1}')::lquery)
```
tsibley: I'm not sure I understand why we want this join condition.

schema/deploy/shipping/views.sql
```diff
@@ -225,4 +225,33 @@ create or replace view shipping.observation_with_presence_absence_result_v1 as
...
+
+    where target.identifier like 'Flu%'
```
tsibley: I think it'd be more robust to filter on the target's lineage since that will be more stable than the target identifier, which may not always start with `Flu` in the future.

schema/deploy/shipping/views.sql
tsibley: Jover and I chatted in person, and this makes sense now: for pan-subtype targets like `Flu_A_pan` which detects H1 and H3, we want to spread out to the more specific lineages for assembly jobs.

It'd be great to add a comment to the query explaining why the `o2` join happens and what the join condition is doing, plus how we use the `o2` lineage in the sequence read set details.

schema/deploy/shipping/views.sql
```diff
@@ -225,4 +225,33 @@ create or replace view shipping.observation_with_presence_absence_result_v1 as
...
+    select sample.identifier as sfs_uuid,
+           sample.details ->> 'nwgc_id' as nwgc_id,
+           sequence_read_set.urls,
+           target.identifier as target,
+           o1.lineage as target_linked_organism,
+           case (o2.lineage) is not null
+             when true
+             then o2.lineage
+             else o1.lineage
+           end as assembly_job_organism,
```
tsibley: This can simplify to a `coalesce(o2.lineage, o1.lineage)`.

joverlee521:
I will incorporate changes suggested and move this to the customization repo.

tsibley:
Thanks @joverlee521!

PR #60, opened 2019-09-06
Augur build api

joverlee521:
This depends on the changes made in https://github.com/seattleflu/id3c-customizations/pull/5

Adds two new endpoints for exporting metadata and consensus genome sequences to SFS augur-build. 
1. `/v1/shipping/augur-build-metadata` 
1. `/v1/shipping/genomic-data/{lineage}/{segment}`

The genomic sequence endpoint requires `lineage` and `segment` to subset the data of the shipping view. This was designed to fit easily into the Snakefile of the SFS augur-build. Let me know if you think there's a better way to do this.


lib/seattleflu/api/datastore.py
```diff
@@ -156,6 +156,42 @@ def store_redcap_det(session: DatabaseSession, document: str) -> None:
...
+@catch_permission_denied
+def export_genomic_data(session: DatabaseSession,
+                        lineage: str,
+                        segment: str) -> dict:
+    """
+    Export sample identifier and sequence from shipping view based on the
+    provided *lineage* and *segment*
+    """
+    if lineage == 'h1n1pdm':
+        lineage = 'h1n1'
```
tsibley: This seems like a detail that only augur-build should know about, not this API.

lib/seattleflu/api/datastore.py
```diff
@@ -156,6 +156,42 @@ def store_redcap_det(session: DatabaseSession, document: str) -> None:
...
+
+    # Change lineage to lquery string
+    lineage = '*.' + lineage + '@'
```
tsibley: Ditto with this. I think augur-build should be specifying full lineages.

lib/seattleflu/api/datastore.py
```diff
@@ -156,6 +156,42 @@ def store_redcap_det(session: DatabaseSession, document: str) -> None:
...
+    Export metadata for augur build from shipping view
+    """
+    with session, session.cursor() as cursor:
+        cursor.execute("select * from shipping.metadata_for_augur_build_v1")
+        return list(cursor)
+
+
+@export
+@catch_permission_denied
+def export_genomic_data(session: DatabaseSession,
```
tsibley: I think it's nice to avoid the very generic "data" and use a more precise noun when possible. Something like `fetch_genomic_sequences` maybe?

lib/seattleflu/api/datastore.py
```diff
@@ -156,6 +156,42 @@ def store_redcap_det(session: DatabaseSession, document: str) -> None:
             raise BadRequestDatabaseError(error) from None
 
 
+@export
+@catch_permission_denied
+def export_metadata_for_augur_build(session: DatabaseSession) -> dict:
```
tsibley: I'd prefer `fetch` over `export`. The rest of the functions in this file are named `store_*` and `fetch` is the clearer complement of `store` to me. (or maybe `get`, but `fetch` aligns with existing functions in other database routines of ours)

lib/seattleflu/api/static/index.html
```diff
@@ -74,8 +74,15 @@ <h3 class="code">POST /v1/receiving/consensus-genome</h3>
     <p>Stores the request data as a consensus genome document in a receiving
     area of the study database.
 
+    <h3 class="code">Get /v1/shipping/augur-build-metadata</h3>
+    <p>Exports consensus genome metadata necessay for SFS augur-build.
+
+    <h3 class="code">Get /v1/shipping/genomic-data/{lineage}/{segment}</h3>
```
tsibley: `Get`→`GET` to match standard HTTP method casing (e.g. as `POST` is in the endpoints above)

lib/seattleflu/api/routes.py
```diff
@@ -127,3 +127,30 @@ def receive_redcap_det(*, session):
...
+@authenticated_datastore_session_required
+def get_metadata(session):
+    """
+    Export metadata needed for SFS augur build
+    """
+    LOG.debug("Exporting metadata for SFS augur build")
+
+    metadata = datastore.export_metadata_for_augur_build(session)
+
+    return jsonify(metadata)
```
tsibley: I think it'd be great to send this (and the other endpoint's response) as NDJSON instead of a JSON array.

In particular, this would let us stream it back from the database to the API client instead of fetching it all into memory before sending the response.

joverlee521: Tried my best with this following the flask streaming [docs](https://flask.palletsprojects.com/en/1.1.x/patterns/streaming/). 😬 
We can talk in person if this isn't what you were thinking.

joverlee521: Updated according to our talk in person. Now should be streaming JSONs from the database to the client.

tsibley: It works like a charm. :-)

Issue #59, opened 2019-09-03
How should we upsert genomic sequences?


joverlee521:
In the current consensus genome etl process we have the function `upsert_genomic_sequences` that will update a genomic sequence when there's a conflict on `genomic_sequence.identifier`. However, this identifier is not unique between sequence read sets, which will lead to consensus genomes without genomic sequences linked to them.

Hypothetical situation:

1.  `sample1`, `organism1`, `sequence-read-set1` creates `consensus-genome1`
            - creates genomic_sequence_1 with identifier `sample1-organism1-segment`

2.  `sample1`, `organism1`, `sequence-read-set2` creates `consensus-genome2`
            - the genomic sequence will have the same identifier `sample1-organism1-segment` so 
            will perform update on genomic_sequence_1, which is still linked to `consensus-genome1`

Writing this all out made me realize that the easiest fix might not be to change the upsert process but to make the genomic sequence identifiers more unique. 

Thoughts @tsibley @kairstenfay ?

tsibley:
I think we should definitely make the identifiers more unique. :-)

tsibley:
Additionally, Jover and I chatted in person about adding an assertion in `upsert_genomic_sequence` to ensure the sequence remains associated with the expected consensus genome. If it doesn't, then that indicates a bug in our ETL process of ensuring unique identifiers.

kairstenfay:
I think making the identifiers longer plus adding an assertion both sound great.

joverlee521:
![](https://github.trello.services/images/mini-trello-icon.png) [Update consensus genome etl to make sequence identifiers more unique](https://trello.com/c/xOAidJ0J/202-update-consensus-genome-etl-to-make-sequence-identifiers-more-unique)

PR #61, opened 2019-09-06
Update consensus genome etl

joverlee521:
This addresses issue #59 

Update the consensus genome etl to create more unique identifiers
for genomic sequences by adding the sequence read set id to the end
of the identifiers.

I have not updated the REVISION number because we don't want the
etl process to double insert the same sequences (there will not be an
identifier conflict with previously created sequences).

The genomic sequences that are already within production will need to
be updated manually. (The genomic sequences will need to be wiped 
anyways to include changes made with https://github.com/seattleflu/assembly/pull/13)

### DIFF
--- a/lib/seattleflu/db/cli/command/etl/consensus_genome.py
+++ b/lib/seattleflu/db/cli/command/etl/consensus_genome.py
@@ -322,10 +322,12 @@ def upsert_genomic_sequence(db: DatabaseSession, genome: NamedTuple, masked_cons
     Upsert genomic sequence given a *genome* record and some information from a
     given *masked_consensus*.
     """
-    LOG.info(f"Upserting genomic sequence «{masked_consensus['sequence_identifier']}»")
+    sequence_identifier = "".join(
+        [masked_consensus['sequence_identifier'], "-", str(genome.sequence_read_set_id)])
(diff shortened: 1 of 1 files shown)


lib/seattleflu/db/cli/command/etl/consensus_genome.py
```diff
@@ -317,15 +318,20 @@ def upsert_genome(db: DatabaseSession, sequence_read_set: NamedTuple,
...
+                            masked_consensus: dict) -> Any:
     """
-    Upsert genomic sequence given a *genome* record and some information from a
-    given *masked_consensus*.
+    Upsert genomic sequence given a *sequence_read_set* record, *genome* record
+    and some information from a given *masked_consensus*.
     """
-    LOG.info(f"Upserting genomic sequence «{masked_consensus['sequence_identifier']}»")
+    sequence_identifier = "".join(
+        [masked_consensus['sequence_identifier'], "-", str(sequence_read_set.id)])
```
tsibley: I noticed that since we're only using the sequence read set id, we could read this from the genome record as `genome.sequence_read_set_id` and avoid passing in `sequence_read_set` to this function.

PR #62, opened 2019-09-11
Ingest kaiser permanente data

kairstenfay:
This PR adds to the `clinical` and `etl clinical` processes. 

There is no need for schema changes.

### DIFF
--- a/lib/seattleflu/db/cli/command/clinical.py
+++ b/lib/seattleflu/db/cli/command/clinical.py
@@ -123,10 +123,10 @@ def load_data(uw_filename: str, uw_nwh_file: str, hmc_sch_file: str):
     """
     clinical_records = load_uw_metadata(uw_filename, date='Collection.Date')
 
-    uw_manifest = load_manifest_data(uw_nwh_file, 'UWMC', 'Collection Date')
-    nwh_manifest = load_manifest_data(uw_nwh_file, 'NWH',
+    uw_manifest = load_uw_manifest_data(uw_nwh_file, 'UWMC', 'Collection Date')
+    nwh_manifest = load_uw_manifest_data(uw_nwh_file, 'NWH',
                                       'Collection Date (per tube)')
-    hmc_manifest = load_manifest_data(hmc_sch_file, 'HMC', 'Collection date')
+    hmc_manifest = load_uw_manifest_data(hmc_sch_file, 'HMC', 'Collection date')
 
     return clinical_records, uw_manifest, nwh_manifest, hmc_manifest
 
@@ -157,7 +157,7 @@ def load_uw_metadata(uw_filename: str, date: str) -> pd.DataFrame:
     return df
 
 
-def load_manifest_data(filename: str, sheet_name: str, date: str) -> pd.DataFrame:
+def load_uw_manifest_data(filename: str, sheet_name: str, date: str) -> pd.DataFrame:
     """
     Given a *filename* and *sheet_name*, returns a pandas DataFrame containing
     barcode manifest data.
--- a/lib/seattleflu/db/cli/command/clinical.py
+++ b/lib/seattleflu/db/cli/command/clinical.py
@@ -10,12 +10,19 @@ import click
 import hashlib
 import logging
 import os
+import re
 import pandas as pd
 import seattleflu.db as db
 from math import ceil
 from seattleflu.db.session import DatabaseSession
 from seattleflu.db.cli import cli
-from . import add_metadata, barcode_quality_control, dump_ndjson, trim_whitespace
+from . import (
+    add_metadata,
+    barcode_quality_control,
+    dump_ndjson,
+    group_true_values_into_list,
+    trim_whitespace,
+)
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -224,14 +224,16 @@ def encounter_details(document: dict) -> dict:
...
-    survey data.
+    Given one or more *races*, returns the matching race identifier found in
+    Audere survey data.
     """
-    if race_name is None:
+    if races is None:
         LOG.debug("No race response found.")
         return [None]
+        
+    races = list(races)
```
joverlee521: If `races` is a string, `list(races)` turns it into a list of characters.

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -237,11 +238,17 @@ def race(races: str or list) -> list:
 
     race_map = {
         "American Indian or Alaska Native": "americanIndianOrAlaskaNative",
+        "amerind": "americanIndianOrAlaskaNative",
         "Asian": "asian",
+        "asian": "asian",
         "Black or African American": "blackOrAfricanAmerican",
+        "black": "black",
```
joverlee521: Should the value here be "blackOrAfricanAmerican" to match others?

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -237,11 +238,17 @@ def race(races: str or list) -> list:
 
     race_map = {
         "American Indian or Alaska Native": "americanIndianOrAlaskaNative",
+        "amerind": "americanIndianOrAlaskaNative",
         "Asian": "asian",
+        "asian": "asian",
```
joverlee521: This seems a little redundant. Would it be better to check `if race[i] in race_map.values()` then return `race[i]`.

lib/seattleflu/db/cli/command/etl/clinical.py
kairstenfay: Yes thank you

lib/seattleflu/db/cli/command/etl/clinical.py
kairstenfay: Yep I just caught that in local testing. I have fixed the code. Thanks!

lib/seattleflu/db/cli/command/etl/clinical.py
kairstenfay: Fair point. I now check `if races[i] in race_map.values()` and `continue` if true. 

I can't return `race[i]` though because I need to handle multiple races.

lib/seattleflu/db/cli/command/clinical.py
```diff
@@ -275,6 +282,157 @@ def parse_sch(sch_filename, output):
...
+    clinical_records = trim_whitespace(clinical_records)
+    clinical_records = add_metadata(clinical_records, kp_filename)
+    clinical_records = add_kp_manifest_data(clinical_records, kp_specimen_manifest_filename)
+
+    clinical_records = convert_numeric_columns_to_binary(clinical_records)
+    clinical_records = rename_symptoms_columns(clinical_records)
+    clinical_records = collapse_columns(clinical_records, 'symptom')
+    clinical_records = collapse_columns(clinical_records, 'race')
+
+    clinical_records['FluShot'] = clinical_records['fluvaxdt'].isna()
```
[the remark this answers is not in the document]
kairstenfay: Yes...

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -224,28 +225,41 @@ def encounter_details(document: dict) -> dict:
         }
 
 
-def race(race_name: str) -> list:
+def race(races: str or list) -> list:
```
tsibley: This isn't a valid type annotation, although I don't blame you at all for not knowing since our type-checking for this repo is in disrepair. What you want is called a "union" type, which syntactically is `Union[str, list]`. `Union` comes from the core `typing` module.

(I really need to get type-checking sorted out so that we can actually get some automatic benefit from our type annotations, but it always seems so low priority.)

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -224,28 +225,41 @@ def encounter_details(document: dict) -> dict:
...
 
+    if not isinstance(races, list):
+        races = [races]
+
     race_map = {
         "American Indian or Alaska Native": "americanIndianOrAlaskaNative",
+        "amerind": "americanIndianOrAlaskaNative",
         "Asian": "asian",
         "Black or African American": "blackOrAfricanAmerican",
+        "black": "blackOrAfricanAmerican",
```
tsibley: This line is part of _etl clinical: edit `race` to take str or list_ but seems like it should be part of _etl clinical: add mapping for KP variables_.

lib/seattleflu/db/cli/command/etl/clinical.py
```diff
@@ -224,28 +225,41 @@ def encounter_details(document: dict) -> dict:
...
+        if races[i] in race_map.values():
+            continue
+
+        if races[i] not in race_map:
+            raise UnknownRaceError(f"Unknown race name «{races[i]}»")
+
+        races[i] = race_map[races[i]]
 
-    return [race_map[race_name]]
+    return races
```
tsibley: To me, this would be more clearly expressed as a [pure function](https://en.wikipedia.org/wiki/Pure_function) combined with `map`, something like:

```python
def standardize_name(name):
    try:
        return name if name in race_map.values() else race_map[name]
    except KeyError:
        raise UnknownRaceError(f"Unknown race name «{name}»") from None

return list(map(standardize_name, races))
```

lib/seattleflu/db/cli/command/etl/clinical.py
kairstenfay: Thanks for the tip! I would love to learn more about type-checking in python.

tsibley: Happy to chat more about this in person at any time. :-)

tsibley:
Thanks, merged! Would you upload the KP data tomorrow?

kairstenfay:
> Thanks, merged! Would you upload the KP data tomorrow?

Yes, I can upload the KP data to production tomorrow.

kairstenfay:
Uploaded

PR #63, opened 2019-09-17
Fix all outstanding type check errors

tsibley:
Figuring that this wasn't going to ever get priority otherwise, I just plowed through all our outstanding type check errors today. Most changes are to the type annotations and so shouldn't have functional effects on the code, but some functional changes were necessary (and some bugs were caught by mypy! :-). `./dev/mypy` now reports no problems and Travis CI is configured to test this going forward, to keep us honest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,4 @@ install:
 
 script:
   - ./dev/show-help ./bin/id3c
+  - ./dev/mypy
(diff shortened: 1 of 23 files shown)


tsibley:
(Just rebased this after merging #62 since there would be merge conflicts otherwise.)

PR #64, opened 2019-09-20
longitudinal: update ETL to extract symptoms

kairstenfay:
I _believe_ this fix is now incorporating symptoms into the `etl longitudinal` pipeline. 
However, I am unable to restore my local database from a copy of production since @tsibley's `location` changes. 
I'm hoping to debug this with him in person on Monday so I can fully test this PR.

### DIFF
--- a/lib/seattleflu/db/cli/command/etl/longitudinal.py
+++ b/lib/seattleflu/db/cli/command/etl/longitudinal.py
@@ -3,6 +3,7 @@ Process longitudinal documents into the relational warehouse.
 """
 import click
 import logging
+from itertools import chain
 from typing import Any, Mapping, Optional
 from datetime import datetime, timezone
 from seattleflu.db import find_identifier
@@ -380,7 +381,31 @@ def symptoms(document: dict) -> list:
     parent_reported_symptoms = document.get("sx_specific") or []
     rn_reported_symptoms = document.get("rn_sx") or []
 
-    return list(set(parent_reported_symptoms + rn_reported_symptoms))
+    symptoms = parent_reported_symptoms + rn_reported_symptoms
+
+    symptoms_map = {
+        "ear_pain": ["earPainOrDischarge"],
+        "fever": ["feelingFeverish"],
+        "headache": ["headaches"],
+        "myalgia": ["muscleOrBodyAches"],
+        "ndv": ["nauseaOrVomiting", "diarrhea"],
+        "nvd": ["nauseaOrVomiting", "diarrhea"],
+        "runny_nose": ["runnyOrStuffyNose"],
+        "sore_throat": ["soreThroat"],
+        "wob": ["increasedTroubleBreathing"],
+        "cough": ["cough"],
+        "fatigue": ["fatigue"],
+        "rash": ["rash"],
+    }
+
+    def standardize_symptom(symptom):
+        try:
+            return symptoms_map[symptom]
+        except KeyError:
+            raise Exception(f"Unknown symptom name «{symptom}»") from None
+
+    # Deduplicate symptoms at the very end
+    return list(set(chain.from_iterable(map(standardize_symptom, symptoms))))
 
 
 def mark_processed(db, longitudinal_id: int, entry: Mapping) -> None:
--- a/lib/seattleflu/db/cli/command/etl/longitudinal.py
+++ b/lib/seattleflu/db/cli/command/etl/longitudinal.py
@@ -37,7 +37,7 @@ LOG = logging.getLogger(__name__)
 # longitudinal records lacking this revision number in their log.  If a
 # change to the ETL routine necessitates re-processing all longitudinal records,
 # this revision number should be incremented.
-REVISION = 2
+REVISION = 3
 
 
 @etl.command("longitudinal", help = __doc__)
@@ -224,7 +224,8 @@ def encounter_details(document: dict) -> dict:
                 "FluShot": flu_shot(document),
                 "AssignedSex": [sex(document)],
                 "HispanicLatino": hispanic_latino(document),
-                "MedicalInsurance": insurance(document)
+                "MedicalInsurance": insurance(document),
+                "Symptoms": symptoms(document),
             },
         }
 
@@ -369,6 +370,17 @@ def insurance(document: dict) -> list:
     }
 
     return [insurance_map.get(insurance_response, None)]
+
+
+def symptoms(document: dict) -> list:
+    """
+    Given a *document*, combines the unique parent-reported and RN-reported
+    symptoms into a list.
+    """
+    parent_reported_symptoms = document.get("sx_specific") or []
+    rn_reported_symptoms = document.get("rn_sx") or []
+
+    return list(set(parent_reported_symptoms + rn_reported_symptoms))
 
 
 def mark_processed(db, longitudinal_id: int, entry: Mapping) -> None:
--- a/lib/seattleflu/db/cli/command/longitudinal.py
+++ b/lib/seattleflu/db/cli/command/longitudinal.py
@@ -314,7 +314,6 @@ def duplicate_audere_keys(df: pd.DataFrame) -> pd.DataFrame:
         "smoking": "HouseholdSmoke",
         "enroll_site": "site",
         "flu_shot": "FluShot",
-        "sx_specific": "Symptoms",
         "illness_impact": "DailyInterference",
         "travel_type": "ChildrensRecentTravel",
         "sample_id": "barcode",


kairstenfay:
[Extract symptoms in longitudinal childcare data](https://trello.com/c/TStuf1nZ/210-extract-symptoms-in-longitudinal-childcare-data)

kairstenfay:
@tsibley I have tested these changes locally and they fix the reported bug;

Running the updated `etl longitudinal` process extracts symptoms (both parent + RN-reported) and puts the resulting `Symptoms` array in the `responses` object.

tsibley:
@kairstenfay Excellent, thank you. Does the `longitudinal parse` step need to be re-run? I see a small change there.

tsibley:
One issue here is that the symptom values are different. If we don't normalize them, then Mike will have to. Eventually we want to model this data better, but not sure if we should try to normalize it now or not. Thoughts?

New childcare symptoms:
```
tom@[local]/seattleflu=# select jsonb_array_elements_text(details->'responses'->'Symptoms') symptom, count(distinct identifier) from warehouse.encounter where identifier like 'sch/year-1/childcare/%' group by symptom order by symptom;
   symptom   │ count 
─────────────┼───────
 cough       │   560
 ear_pain    │    47
 fatigue     │   213
 fever       │   189
 headache    │    19
 myalgia     │    15
 ndv         │    54
 nvd         │    29
 rash        │    24
 runny_nose  │   869
 sore_throat │    60
 wob         │    43
(12 rows)
```

Existing symptoms:
```
tom@production.db.seattleflu.org/production=> select jsonb_array_elements_text(details->'responses'->'Symptoms') symptom, count(distinct identifier) from warehouse.encounter group by symptom order by symptom;
          symptom          │ count 
───────────────────────────┼───────
 chillsOrSweats            │   598
 cough                     │  3701
 diarrhea                  │   733
 earPainOrDischarge        │   588
 fatigue                   │  3409
 feelingFeverish           │  2357
 headaches                 │  1923
 increasedTroubleBreathing │  1267
 muscleOrBodyAches         │  2535
 nauseaOrVomiting          │   974
 rash                      │   289
 runnyOrStuffyNose         │  3143
 soreThroat                │  2013
(13 rows)
```

kairstenfay:
> @kairstenfay Excellent, thank you. Does the `longitudinal parse` step need to be re-run? I see a small change there.

The small change in `parse` means we're no longer creating a 'Symptoms' key in the document that is copied from parent reported systems. 
If it would cause confusion down the road then it makes sense to re-run `parse`, but in my opinion it isn't necessary because it doesn't impact the ETL.

kairstenfay:
> One issue here is that the symptom values are different. If we don't normalize them, then Mike will have to. Eventually we want to model this data better, but not sure if we should try to normalize it now or not. Thoughts?
> 
> New childcare symptoms:
> 
> ```
> tom@[local]/seattleflu=# select jsonb_array_elements_text(details->'responses'->'Symptoms') symptom, count(distinct identifier) from warehouse.encounter where identifier like 'sch/year-1/childcare/%' group by symptom order by symptom;
>    symptom   │ count 
> ─────────────┼───────
>  cough       │   560
>  ear_pain    │    47
>  fatigue     │   213
>  fever       │   189
>  headache    │    19
>  myalgia     │    15
>  ndv         │    54
>  nvd         │    29
>  rash        │    24
>  runny_nose  │   869
>  sore_throat │    60
>  wob         │    43
> (12 rows)
> ```
> 
> Existing symptoms:
> 
> ```
> tom@production.db.seattleflu.org/production=> select jsonb_array_elements_text(details->'responses'->'Symptoms') symptom, count(distinct identifier) from warehouse.encounter group by symptom order by symptom;
>           symptom          │ count 
> ───────────────────────────┼───────
>  chillsOrSweats            │   598
>  cough                     │  3701
>  diarrhea                  │   733
>  earPainOrDischarge        │   588
>  fatigue                   │  3409
>  feelingFeverish           │  2357
>  headaches                 │  1923
>  increasedTroubleBreathing │  1267
>  muscleOrBodyAches         │  2535
>  nauseaOrVomiting          │   974
>  rash                      │   289
>  runnyOrStuffyNose         │  3143
>  soreThroat                │  2013
> (13 rows)
> ```

I think it would be ideal for us to normalize the symptoms. I will work on a new commit that does this. 

I think the correct mapping from the SCH data dictionary is this (but will confirm):
```
>  cough       │   Cough
>  ear_pain    │    Ear pain or ear discharge
>  fatigue     │   Fatigue (feeling tired)
>  fever       │   Feeling feverish
>  headache    │    Headaches
>  myalgia     │    Muscle or body aches
>  ndv         │    Nausea, diarrhea, or vomiting
>  nvd         │    Nausea, diarrhea, or vomiting
>  rash        │    Rash
>  runny_nose  │   Runny/stuffy nose
>  sore_throat │    Sore throat
>  wob         │    Increased trouble with breathing
```

tsibley:
:+1: for both items. Thanks!

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -374,7 +375,30 @@ def symptoms(document: dict) -> list:
...
+        "myalgia": "muscleOrBodyAches",
+        "ndv": ["nauseaOrVomiting", "diarrhea"],
+        "nvd": ["nauseaOrVomiting", "diarrhea"],
+        "runny_nose": "runnyOrStuffyNose",
+        "sore_throat": "soreThroat",
+        "wob": "increasedTroubleBreathing",
+        "cough": "cough",
+        "fatigue": "fatigue",
+        "rash": "rash",
+    }
```
kairstenfay: I'm curious to know if either of you have a better idea for how to handle "cough" = "cough" -type mapping. I wanted to avoid a db lookup of existing symptoms, and the only other idea I had was to check inside a list `["rash", "fatigue", "cough"]` before mapping. I didn't like that idea very much, though.

kairstenfay:
This PR is now ready for re-review.

lib/seattleflu/db/cli/command/etl/longitudinal.py
kairstenfay: I will update the `standardize_symptom` function depending on what direction we want to go with this.

kairstenfay:
I re-pushed to fix a typing error

lib/seattleflu/db/cli/command/etl/longitudinal.py
tsibley: I think it's fine to have an explicit identity entry (a = a) in the map.

I have some other notes on this which I pushed as a fixup commit.

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -380,38 +381,31 @@ def symptoms(document: dict) -> list:
     parent_reported_symptoms = document.get("sx_specific") or []
     rn_reported_symptoms = document.get("rn_sx") or []
 
-    symptoms = list(set(parent_reported_symptoms + rn_reported_symptoms))
+    symptoms = parent_reported_symptoms + rn_reported_symptoms
```
tsibley: Using `set()` here to de-duplicate is too early now that we standardize names into multiple possible things below, so I moved it to the end.

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -380,38 +381,31 @@ def symptoms(document: dict) -> list:
...
     }
 
-    def flatten(array):
-        for symptom in array:
-            if isinstance(symptom, list):
-                for nested_symptom in flatten(symptom):
-                    yield nested_symptom
-            else:
-                yield symptom
-
```
tsibley: I replaced this `flatten` function with `chain.from_iterable` from the `itertools` standard library. This was made simpler by making all values of `symptom_map` lists instead of polymorphic.

lib/seattleflu/db/cli/command/etl/longitudinal.py
```diff
@@ -380,38 +381,31 @@ def symptoms(document: dict) -> list:
...
     def standardize_symptom(symptom):
         try:
-            return symptom if symptom in symptoms_map.values() else symptoms_map[symptom]
+            return symptoms_map[symptom]
```
tsibley: This conditional seemed odd and unnecessary? I also think it didn't account for the list-valued entries in the map.

kairstenfay: Agreed, I meant to clean that up after deciding how to handle a = a mapping.

lib/seattleflu/db/cli/command/etl/longitudinal.py
kairstenfay: That's very clever. I looked at `chain` initially but it was splitting my strings. Having every symptom as a list makes a lot of sense. Thanks!

tsibley:
Merged and deployed!

PR #66, opened 2019-09-26
reportable-conditions: improve logic for null site

kairstenfay:
Change the `reportable-conditions notify` routine to perform additional
steps when a sample is missing a linked site.

If the site is missing from a sample encounter, look in the manifest
data and relay workbook and sheet name in the Slack notification.

### DIFF
--- a/lib/seattleflu/db/cli/command/reportable_conditions.py
+++ b/lib/seattleflu/db/cli/command/reportable_conditions.py
@@ -94,8 +94,12 @@ def notify(*, action: str):
             with db.savepoint(f"reportable condition presence_absence_id {record.id}"):
                 LOG.info(f"Processing reportable condition, presence_absence_id «{record.id}»")
 
+                if not record.site:
+                    LOG.info(f"No site found for presence_absence_id «{record.id}». " +
+                        "Inferring site from manifest data.")
+
                 url = SLACK_WEBHOOK_REPORTING_GENERAL \
-                    if record.site not in childrens_sites \
+                    if record.site not in childrens_sites and record.sheet != 'SCH' \
                     else SLACK_WEBHOOK_REPORTING_CHILDRENS
 
                 response = send_slack_post_request(record, url)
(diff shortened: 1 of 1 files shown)


kairstenfay:
Right now the logic I'm employing is that manifest data (workbook, sheet) only get added in the event site is null. Thoughts?

lib/seattleflu/db/cli/command/reportable_conditions.py
```diff
@@ -166,6 +170,13 @@ def send_slack_post_request(record: NamedTupleCursor.Record, url: str) -> reques
         "condition": record.lineage
     }
 
+    if not record.site:
+        additional_data = {
```
joverlee521: Do we want a higher level key like `manifest` to make it clear this is manifest data?

tsibley: I think that's a good idea.

lib/seattleflu/db/cli/command/reportable_conditions.py
```diff
@@ -94,8 +94,12 @@ def webhook(suffix):
             with db.savepoint(f"reportable condition presence_absence_id {record.id}"):
                 LOG.info(f"Processing reportable condition, presence_absence_id «{record.id}»")
 
+                if not record.site:
+                    LOG.info(f"No site found for presence_absence_id «{record.id}». " +
+                        "Inferring site from manifest data.")
+
                 url = SLACK_WEBHOOK_REPORTING_GENERAL \
-                    if record.site not in childrens_sites \
+                    if record.site not in childrens_sites or record.sheet != 'SCH' \
```
tsibley: Unless I'm turned around myself, I'm pretty sure this logic is backwards. `or` should be `and`.

tsibley:
I pushed a fixup commit to fix the logic error and put the `workbook` and `sheet` under a `manifest` key. Take a look?

tsibley:
[Rethink Reportable Conditions Slack bot logic](https://trello.com/c/kWKyynNH/213-rethink-reportable-conditions-slack-bot-logic)

lib/seattleflu/db/cli/command/reportable_conditions.py
kairstenfay: Ah, you're right. I tried to make the most minimal change, but negative logic always gets me.

kairstenfay:
It looks good, thanks!

tsibley:
Merged and deployed!

PR #67, opened 2019-10-01
Add target organism lineage to shipping.presenc...

kairstenfay:
e_absence_result_v1

### DIFF
--- a/schema/deploy/shipping/views.sql
+++ b/schema/deploy/shipping/views.sql
@@ -108,11 +108,13 @@ create or replace view shipping.presence_absence_result_v1 as
 
     select sample.identifier as sample,
            target.identifier as target,
-           present
+           present,
+           organism.lineage as organism
 
     from warehouse.sample
     join warehouse.presence_absence using (sample_id)
     join warehouse.target using (target_id)
+    left join warehouse.organism using (organism_id)
     where target.control = false;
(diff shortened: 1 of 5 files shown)


schema/deploy/shipping/views.sql
```diff
@@ -108,11 +108,13 @@ create or replace view shipping.presence_absence_result_v1 as
...
     select sample.identifier as sample,
            target.identifier as target,
-           present
+           present,
+           organism.lineage
 
     from warehouse.sample
     join warehouse.presence_absence using (sample_id)
     join warehouse.target using (target_id)
+    join warehouse.organism using (organism_id)
```
tsibley: This needs to be a left join so we don't exclude results for targets that aren't yet linked to an organism.

kairstenfay: Done.

kairstenfay:
[Add target organism lineage to shipping.presence_absence_result_v1](https://trello.com/c/0qH0FR5n/215-add-target-organism-lineage-to-shippingpresenceabsenceresultv1)

tsibley:
I renamed the view column from `lineage` → `organism` to match our conventions so far (e.g. `target.identifier` is named just `target` in most views).

Merged and deployed.

PR #68, opened 2019-10-02
Fix: update consensus-genome-process grants

joverlee521:
This should stop the error alerts we are getting in slack regarding `etl consensus-genome`.

We added update to segment in the consensus genome etl
but forgot to grant permissions to update segment to the
consensus-genome-processor role.


tsibley:
Merged and deployed. I added a schema tag to your commit before merging.

PR #69, opened 2019-10-04
Add current and modified timestamps to warehouse

kairstenfay:
Create new columns titled 'created' and 'modified' in the following
tables:
* `warehouse.encounters`
* `warehouse.sample`
* `warehouse.presence_absence`
* `warehouse.sequence_read_set`
* `warehouse.genomic_sequence`

These columns each have default values of the current timestamp. The
`modified` column is updated with a new timestamp via a trigger function
every time a record is updated. The trigger function modifies the
`modified` column even if the update changes are identical to the
previous changes.

---------
## TODO: 
~I didn't write any `verify` scripts so far because I wasn't sure what to test/how to test it.~
Currently, the `modified` column is updated on every update _even if_ the update results in no actual differences. Is this desirable?

### DIFF
--- a/schema/deploy/functions/update-modified-timestamp.sql
+++ b/schema/deploy/functions/update-modified-timestamp.sql
@@ -1,0 +1,17 @@
+-- Deploy seattleflu/schema:functions/update-modified-timestamp to pg
+
+begin;
+
+create or replace function warehouse.update_modified_timestamp() returns trigger as $$
+
+    begin
+        NEW.modified = now();
+        return new;
+    end;
+$$ language plpgsql
+    stable;
+
+comment on function warehouse.update_modified_timestamp is
+    'Updates the modified column timestamp with the current time on every update';
+
+commit;
--- a/schema/deploy/warehouse/encounter/triggers/update-modified-timestamp.sql
+++ b/schema/deploy/warehouse/encounter/triggers/update-modified-timestamp.sql
@@ -1,0 +1,12 @@
+-- Deploy seattleflu/schema:warehouse/encounter/triggers/update-modified-timestamp to pg
+-- requires: warehouse/encounter
+-- requires: functions/update-modified-timestamp
+
+begin;
+
+create trigger update_modified_timestamp
+    before update on warehouse.encounter
+    for each row
+        execute procedure warehouse.update_modified_timestamp();
+
+commit;
--- a/schema/sqitch.plan
+++ b/schema/sqitch.plan
@@ -122,3 +122,16 @@ shipping/views [shipping/views@2019-08-26] 2019-10-01T00:49:00Z Kairsten Fay <kf
 @2019-09-30 2019-10-01T01:35:37Z Kairsten Fay <kfay@fredhutch.org> # Add target organism lineage to presence_absence_result_v1
 roles/consensus-genome-processor/grants [roles/consensus-genome-processor/grants@2019-09-30] 2019-10-02T19:12:27Z Jover Lee <joverlee@fredhutch.org> # Rework grants to allow update on segment of warehouse.genomice_sequence
 @2019-10-02 2019-10-02T23:39:34Z Thomas Sibley <tsibley@fredhutch.org> # schema as of 2 October 2019
+
+functions/update-modified-timestamp 2019-10-04T01:14:53Z Kairsten Fay <kfay@fredhutch.org> # Update modified column timestamp
+warehouse/encounter/timestamp [warehouse/encounter] 2019-10-04T00:24:58Z Kairsten Fay <kfay@fredhutch.org> # Timestamp encounter records with date created and modified
+warehouse/encounter/triggers/update-modified-timestamp [warehouse/encounter functions/update-modified-timestamp] 2019-10-04T00:39:59Z Kairsten Fay <kfay@fredhutch.org> # Trigger to update modified timestamp on encounters
+warehouse/sample/timestamp [warehouse/sample] 2019-10-04T01:26:30Z Kairsten Fay <kfay@fredhutch.org> # Timestamp sample records with date created and modified
+warehouse/sample/triggers/update-modified-timestamp [warehouse/sample functions/update-modified-timestamp] 2019-10-04T01:29:39Z Kairsten Fay <kfay@fredhutch.org> # Trigger to update modified timestamp on samples
+warehouse/presence_absence/timestamp [warehouse/presence_absence] 2019-10-04T01:34:39Z Kairsten Fay <kfay@fredhutch.org> # Timestamp presence-absence records with date created and modified
+warehouse/presence_absence/triggers/update-modified-timestamp [warehouse/presence_absence functions/update-modified-timestamp] 2019-10-04T01:36:38Z Kairsten Fay <kfay@fredhutch.org> # Trigger to update modified timestamp on presence-absence tests
+warehouse/sequence-read-set/timestamp [warehouse/sequence-read-set] 2019-10-04T01:38:54Z Kairsten Fay <kfay@fredhutch.org> # Timestamp sequence read set records with date created and modified
+warehouse/sequence-read-set/triggers/update-modified-timestamp [warehouse/sequence-read-set functions/update-modified-timestamp] 2019-10-04T01:40:46Z Kairsten Fay <kfay@fredhutch.org> # Trigger to update modified timestamp on sequence read sets
+warehouse/genomic-sequence/timestamp [warehouse/genomic-sequence] 2019-10-04T01:43:50Z Kairsten Fay <kfay@fredhutch.org> # Timestamp genomic sequence records with date created and modified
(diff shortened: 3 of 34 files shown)


kairstenfay:
[Timestamp of records in id3c warehouse](https://trello.com/c/KYZ8e3sJ/232-timestamp-of-records-in-id3c-warehouse)

joverlee521:
I added a verify script for the shared trigger function, but since all the other triggers use the same function, I don't think we need to verify every single one individually?

kairstenfay:
> I added a verify script for the shared trigger function, but since all the other triggers use the same function, I don't think we need to verify every single one individually?

That makes sense to me.

schema/deploy/functions/update-modified-timestamp.sql
```diff
@@ -0,0 +1,18 @@
...
+begin;
+
+create or replace function warehouse.update_modified_timestamp() returns trigger as $$
+
+    begin
+        NEW.modified = now();
+        return new;
+    end;
+$$ language plpgsql
+    security definer
```
tsibley: `security definer` functions should be used rarely and with explicit reason. This function doesn't need it, so I assume this is a copy-paste thing. It is important to understand all the code you're adding/using, esp. when cribbing from existing code. I'm happy to explain unknown bits of code anytime. :-)

schema/deploy/functions/update-modified-timestamp.sql
```diff
@@ -0,0 +1,18 @@
...
+    set search_path = pg_catalog, public, pg_temp;
```
tsibley: `search_path` is also unnecessary for this function, but there's also no harm in leaving it.

schema/deploy/functions/update-modified-timestamp.sql
```diff
@@ -0,0 +1,18 @@
...
+
+begin;
+
+create or replace function warehouse.update_modified_timestamp() returns trigger as $$
+
+    begin
+        NEW.modified = now();
+        return new;
+    end;
+$$ language plpgsql
```
tsibley: This function should be marked as `stable` instead of the default of `volatile` since it only uses stable functions and doesn't have side-effects. More info in the Pg docs: https://www.postgresql.org/docs/10/xfunc-volatility.html

schema/sqitch.plan
```diff
@@ -122,3 +122,16 @@ shipping/views [shipping/views@2019-08-26] 2019-10-01T00:49:00Z Kairsten Fay <kf
...
+warehouse/presence_absence/triggers/update-modified-timestamp [warehouse/presence_absence] 2019-10-04T01:36:38Z Kairsten Fay <kfay@fredhutch.org> # Trigger to update modified timestamp on presence-absence tests
+warehouse/sequence-read-set/timestamp [warehouse/sequence-read-set] 2019-10-04T01:38:54Z Kairsten Fay <kfay@fredhutch.org> # Timestamp sequence read set records with date created and modified
+warehouse/sequence-read-set/triggers/update-modified-timestamp [warehouse/sequence-read-set] 2019-10-04T01:40:46Z Kairsten Fay <kfay@fredhutch.org> # Trigger to update modified timestamp on sequence read sets
+warehouse/genomic-sequence/timestamp [warehouse/genomic-sequence] 2019-10-04T01:43:50Z Kairsten Fay <kfay@fredhutch.org> # Timestamp genomic sequence records with date created and modified
+warehouse/genomic-sequence/triggers/update-modified-timestamp [warehouse/genomic-sequence] 2019-10-04T01:45:49Z Kairsten Fay <kfay@fredhutch.org> # Trigger to update modified timestamp on genomic sequences
```
tsibley: All of the `*/update-modified-timestamp` changes should declare a dep on `functions/update-modified-timestamp`. You can do this by:

1. `sqitch revert functions/update-modified-timestamp`
2. Edit the `sqitch.plan` file (for the computers) and add the `-- requires: ...` comment to the top of each deploy (for us humans only)
3. `sqitch deploy`

schema/deploy/functions/update-modified-timestamp.sql
kairstenfay: When is adding `search_path` necessary?

schema/deploy/functions/update-modified-timestamp.sql
kairstenfay: I remember adding `security definer` to an existing trigger function at some point, but I can't remember why it was necessary. Is this in earlier trigger functions because we changed the schema owner at some point?

schema/deploy/functions/update-modified-timestamp.sql
kairstenfay: Neat to read about the optimization advantages of marking functions as `stable` or `immutable`.

schema/deploy/functions/update-modified-timestamp.sql
tsibley: Remembering why something was done is exactly why we write good commit messages: https://github.com/seattleflu/id3c/commit/1efa888834e553efcf5374058f9b5dedb2e1dee4 :-D

schema/deploy/functions/update-modified-timestamp.sql
tsibley: It forces the `search_path` to a known value so that unqualified function/table names aren't found in an unexpected schema (like a schema controlled by the unprivileged database user). This is important because a `security definer` function will call other functions with the same privileged user it runs as. Details here: https://www.postgresql.org/docs/10/sql-createfunction.html#SQL-CREATEFUNCTION-SECURITY

tsibley: (Also happy to more concretely sketch out why this matters in person.)

schema/deploy/warehouse/encounter/triggers/update-modified-timestamp.sql
```diff
@@ -0,0 +1,12 @@
+-- Deploy seattleflu/schema:warehouse/encounter/triggers/update-modified-timestamp to pg
+-- requires: warehouse/encounter
+-- requires: warehouse/encounter/triggers/update-modified-timestamp
```
tsibley: Looks like `warehouse/encounter/triggers/update-modified-timestamp` was used by mistake instead of `functions/update-modified-timestamp` both at the top of deploy files and in the plan file.

schema/deploy/functions/update-modified-timestamp.sql
kairstenfay: I think I get it. So because I declare the function as being part of `warehouse`, setting the `search_path` is redundant?

schema/deploy/warehouse/encounter/triggers/update-modified-timestamp.sql
kairstenfay: Yep, sure enough. I fixed it. 

The pains of naming things too similarly....

schema/deploy/functions/update-modified-timestamp.sql
tsibley: Mmm, not quite. I'll explain in person tomorrow?

kairstenfay: That would be great!

tsibley:
Ahh, nothing is ever simple. There's one other catch here which I'm just now realizing:

The combination of `not null` and `default` means that existing records will get a `created` and `modified` of the deploy time of these changes. That's misleading. We should either a) allow the columns to be null and make sure existing records are null (that is, the timestamps are unknown) or b) explicitly set the timestamps if we can reliably discern them for existing records.

kairstenfay:
> Ahh, nothing is ever simple. There's one other catch here which I'm just now realizing:
> 
> The combination of `not null` and `default` means that existing records will get a `created` and `modified` of the deploy time of these changes. That's misleading. We should either a) allow the columns to be null and make sure existing records are null (that is, the timestamps are unknown) or b) explicitly set the timestamps if we can reliably discern them for existing records.
> 
> Sorry this has turned into such another can of worms!

I addressed this in the latest updated commit via:
```
alter table warehouse.genomic_sequence
    add column created timestamp with time zone,
    add column modified timestamp with time zone not null default now();

alter table warehouse.genomic_sequence
    alter column modified set default now();
```
etc.

tsibley:
Why just `created` and not `modified` too?

kairstenfay:
> Why just `created` and not `modified` too?

Because we are modifying these tables by adding extra columns and triggers to them.

But you're right -- it's probably not a significant enough change to track with a timestamp. Would you prefer I do the same with `modified` as I did with `created`?

tsibley:
> Because we are modifying these tables by adding extra columns and triggers to them.

Nod. Changing the schema around is a different sort of thing from modifying the _records_ in the tables, I think.

> But you're right -- it's probably not a significant enough change to track with a timestamp. Would you prefer I do the same with modified as I did with created?

That'd be great, thanks!

kairstenfay:
> > But you're right -- it's probably not a significant enough change to track with a timestamp. Would you prefer I do the same with modified as I did with created?
> 
> That'd be great, thanks!

It is done.

tsibley:
Merged and deployed. :pray: Thanks for your patience.

PR #71, opened 2019-10-08
Add `redcap-det` commands

joverlee521:
Get REDCap records that have been created/modified since a provided
start date. Parse the records into "fake" DET records and print
as NDJSON. Upload created DET records to receiving.redcap_det

This creates "fake" DET records that mimics REDCap DETs but is missing `project_url` and `username`. I don't really see a way to parse these from the record data returned from REDCap, so if we do want to include them, I think they just need to be additional options in the `redcap-det parse` command. I don't think they are crucial to the ETL process so omitted for now.

### DIFF
--- a/lib/seattleflu/db/cli/command/redcap_det.py
+++ b/lib/seattleflu/db/cli/command/redcap_det.py
@@ -16,7 +16,7 @@ import click
 import logging
 import re
 import requests
-from typing import List
+from typing import Any, List
 from seattleflu.db.cli import cli
 from seattleflu.db.session import DatabaseSession
 from seattleflu.db.datatypes import as_json
@@ -28,10 +28,6 @@ def redcap_det():
     pass
 
 @redcap_det.command("generate")
-@click.option("--project-id",
-    metavar = "<project-id>",
-    help = "The project ID for the REDCap project",
-    required = True)
 @click.option("--token-name",
     metavar = "<token-name>",
     help = "The name of the environment variable that holds the API token",
@@ -41,7 +37,7 @@ def redcap_det():
     help = "Limit to REDCap records that have been created/modified since the given date. " +
            "Format must be YYYY-MM-DD HH:MM:SS (e.g. '2019-01-01 00:00:00')")
 
-def generate(project_id: str, token_name: str, since_date: str):
+def generate(token_name: str, since_date: str):
     """
     Generate DET notifications for REDCap records.
 
@@ -60,6 +56,10 @@ def generate(project_id: str, token_name: str, since_date: str):
     # trailing 'api' from the API URL
     redcap_url = re.sub(r'api/?$', '', api_url)
 
+    project = get_project(api_url, api_token)
+
+    LOG.info(f"REDCap project #{project['project_id']}: {project['project_title']}")
+
     if since_date:
         LOG.debug(f"Getting all records that have been created/modified since {since_date}")
     else:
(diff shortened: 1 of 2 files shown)


joverlee521:
> Can you please describe when/why or how often we would need to run the `redcap-det`?

The Kiosk Enrollment project uses the mobile app to fill in records, but REDCap explicitly says that data imported through the mobile app don't fire DETs. The `redcap-det` command will be used to fill in this gap, so we will probably have to set up a cron job to run it hourly so that we have the most up-to-date encounter data from kiosks.

kairstenfay:
> The Kiosk Enrollment project uses the mobile app to fill in records, but REDCap explicitly says that data imported through the mobile app don't fire DETs. The `redcap-det` command will be used to fill in this gap, so we will probably have to set up a cron job to run it hourly so that we have the most up-to-date encounter data from kiosks.

Does this replace the DETs we receive directly from REDCap, or is it in addition to?

joverlee521:
> > The Kiosk Enrollment project uses the mobile app to fill in records, but REDCap explicitly says that data imported through the mobile app don't fire DETs. The `redcap-det` command will be used to fill in this gap, so we will probably have to set up a cron job to run it hourly so that we have the most up-to-date encounter data from kiosks.
> 
> Does this replace the DETs we receive directly from REDCap, or is it in addition to them?

This should be in addition to them. We should still be receiving DETs directly from REDCap for all other projects. Hopefully we only need to use this command for the Kiosk project.

kairstenfay:
Thanks for the explanation!

lib/seattleflu/db/cli/command/redcap_det.py
```diff
@@ -0,0 +1,180 @@
...
+    parameters = {
+        'content': 'record',
+        'format': 'json',
+        'type': 'flat',
+        'token': api_token,
+        'dateRangeBegin': start_date
+    }
+
+    records = get_redcap_data(api_url, parameters)
+    return records
```
kairstenfay: This can be simplified to `return get_redcap_data(...`

lib/seattleflu/db/cli/command/redcap_det.py
```diff
@@ -33,8 +33,8 @@ def redcap_det():
     required = True)
 @click.option("--token-name",
     metavar = "<token-name>",
-    help = "The name of the environment variable that holds the API token for the <project-id>",
-    required = True)
+    help = "The name of the environment variable that holds the API token",
+    default = "REDCAP_API_TOKEN")
```
kairstenfay: Oh, I realized: I think all projects share the same API Token. It's peculiarly only accessible through the Kiosks project, but it is the same one used for Shelters.

So we may not need this option at all.

joverlee521: Each project needs a different API Token, the API Token is actually how REDCap knows which project to access records from.

lib/seattleflu/db/cli/command/redcap_det.py
```diff
@@ -0,0 +1,180 @@
...
+    det_record = {
+       'redcap_url': redcap_url,
+       'project_id': project_id,
+       'record': record['record_id'],
+       'instrument': instrument,
+       instrument_complete: record[instrument_complete]
+    }
+
+    if 'redcap_event_name' in record:
+        det_record['redcap_event_name'] = record['redcap_event_name']
```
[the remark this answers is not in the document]
joverlee521: I haven't see these show up in the real DETs but that might be because we aren't receiving any from the Household studies. Not sure if they come in through those projects

joverlee521:
> Looks good! I added several commits onto the end of the branch to tweak minor things I noticed while reviewing. It seemed easier/clearer to do that than to write them up here. Take a look and see if you agree with the suggestions?

Looks good! Thanks for the tweaking 😄

lib/seattleflu/db/cli/command/redcap_det.py
kairstenfay: Oh... do you know how we go about requesting API tokens for the other projects?

tsibley: If you have API access, there will be "API" and "API Playground" links in the sidebar nav for the project. The first page will let you generate a token for the project.

lib/seattleflu/db/cli/command/redcap_det.py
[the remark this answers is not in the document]
tsibley: Looks like they do. I checked the kiosks project and the one-week followup is a repeating instrument. I pushed one more commit to add these fields.

PR #75, opened 2019-10-17
Fhir table

kairstenfay:
Create new table `recieving.fhir` and associated role to upload documents


kairstenfay:
[Receiving table in ID3C for REDCap FHIR documents](https://trello.com/c/dQHImcWF/247-receiving-table-in-id3c-for-redcap-fhir-documents)

tsibley:
I've pushed a commit up which accounts for a previous bad example on my part. Let me know if it looks good?

As a separate task, we'll also want an API endpoint for this table. I'll make a Trello card.

PR #76, opened 2019-10-18
Add sample origin and swab site

joverlee521:
Update `manifest parse` to include columns `sample_origin` and `swab_site`.

Manifests for 2019-2020 are all contained within one workbook and one sheet. To infer `site` from manifest data for reportable conditions, we need to include `sample_origin` and `swab_site`.

### DIFF
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -56,6 +56,18 @@ def manifest():
            "Must match exactly; shell-style glob patterns are supported.",
     required = True)
 
+@click.option("--sample-origin",
+    metavar = "<column>",
+    help = "Name of the single column containing sample origin. "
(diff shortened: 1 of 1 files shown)


joverlee521:
@tsibley had mentioned previously:
> I want to generalize how we capture those in manifest parse (instead of adding yet another command-line/config parameter).

I'm punting that yet again since it was easier to just add the two new fields for now.

joverlee521:
[Update reportable-conditions Slack logic to reflect changes in 2019-20 specimen manifests](https://trello.com/c/vJIrOpwl/241-update-reportable-conditions-slack-logic-to-reflect-changes-in-2019-20-specimen-manifests)

tsibley:
> I'm punting that yet again since it was easier to just add the two new fields for now.

Fair enough!

PR #77, opened 2019-10-18
Tests: Check that our SQL changes don't use unlocalized set search_path commands

tsibley:
While reviewing #75, I recalled wanting a test like this at the time we switched from `set` to `set local` (a985ed1). Now that we had an easier testing setup, I thought I'd be able to get this going in an hour, so I did.

### DIFF
--- a/tests/search-path.py
+++ b/tests/search-path.py
@@ -1,0 +1,35 @@
+import pytest
+import re
+from pathlib import Path
+from sqlparse import split
+
+topdir = Path(__file__).resolve().parent.parent
+
+sql_files = list(sorted(topdir.glob("schema/**/*.sql")))
+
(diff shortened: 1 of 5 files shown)


tests/search-path.py
```diff
@@ -0,0 +1,35 @@
...
+    r"""
+    # SET command not followed by LOCAL modifier
+    set \s+ (?!local\s+)
+
+    # …followed, potentially after other parameters first, by "search_path"
+    [^;]*? search_path
+
+    # …which is not followed by a comment telling us to ignore it
+    (?! [^\n]*? -- \s* tests/search-path: \s* ignore)
+    """,
```
kairstenfay: I love the multi-line, commented regex.

PR #73, opened 2019-10-15
Etl changes for redcap

kairstenfay:
See seattleflu/id3c-customizations/pull/10 for the etl routine accompaniment of this PR.

PR #78, opened 2019-10-18
REDCap ETL

tsibley:
Incorporates #73 + some additional robustness around standardizing race. I expect more work will be done here as we review/integrate our specific ETLs other than swab and send.

### DIFF
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -10,6 +10,7 @@ import logging
 import requests
 from datetime import datetime, timezone
 from enum import Enum
+from typing import Tuple
 from id3c.db.session import DatabaseSession
 from id3c.db.datatypes import Json
 from . import etl
@@ -31,14 +32,14 @@ def redcap_det():
     pass
 
 
-def get_redcap_record(record_id: str, token_name: str) -> dict:
+def get_redcap_record(record_id: str) -> dict:
     """
     Gets one REDCap record containing all instruments via web API based on the
     provided *record_id*.
-    Requires an environmental variable described by *token_name* for REDCap to
-    access the records.
     """
     LOG.debug(f"Getting REDCap record «{record_id}» for all instruments")
+
+    url, token = get_redcap_api_credentials()
 
     data = {
         'content': 'record',
@@ -46,7 +47,7 @@ def get_redcap_record(record_id: str, token_name: str) -> dict:
         'type': 'flat',
         'rawOrLabel': 'label',
         'exportCheckboxLabel': 'true',
-        'token': os.environ[token_name],
+        'token': token,
         'records': record_id,
     }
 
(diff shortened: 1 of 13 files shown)


joverlee521:
I'm guessing we will want to pull out all mapping functions into a shared module and do the same type of doctests. Although the project specific ones might live in id3c-customizations?

kairstenfay:
Neat, I never knew about `doctests` before.

tsibley:
> I'm guessing we will want to pull out all mapping functions into a shared module and do the same type of doctests. Although the project specific ones might live in id3c-customizations?

Ideally yes to both.

> Neat, I never knew about `doctests` before.

:sparkles: :grin:

tsibley:
Rebased this onto latest master to pick up some type annotation updates that make it possible to type-check our customizations code.

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -63,6 +62,39 @@ def get_redcap_record(project_id: str, record_id: str, token_name: str) -> dict:
...
+    >>> is_complete("test", {})
+    Traceback (most recent call last):
+        ...
+    KeyError: 'test_complete'
+    """
+    return record[f"{instrument}_complete"] in {
+        InstrumentStatus.Complete.name,
+        InstrumentStatus.Complete.value,
+        str(InstrumentStatus.Complete.value)
+    }
```
kairstenfay: @tsibley why did you put the complete statuses in a set instead of a list?

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -55,12 +56,34 @@ def get_redcap_record(record_id: str, token_name: str) -> dict:
...
 
+def get_redcap_api_credentials() -> Tuple[str, str]:
+    """
+    Returns a tuple of ``(url, token)`` for use with REDCap's web API.
+
+    Requires the environmental variables ``REDCAP_API_URL`` and
+    ``REDCAP_API_TOKEN``.
+    """
+    url = os.environ.get("REDCAP_API_URL")
+    token = os.environ.get("REDCAP_API_TOKEN")
```
kairstenfay: How does this function allow other ways than the environment to provide values when the environment variable appears to be hard-coded here as `REDCAP_API_TOKEN`? 

Each REDCap project requires a different API token, and I don't see why this revision is preferred.

kairstenfay: Would the `REDCAP_API_TOKEN` need to be declared before running every etl? 

e.g.
```
REDCAP_API_TOKEN=$REDCAP_SINGLE_SWAB_AND_SEND_API_TOKEN PGDATABASE=seattleflu pipenv run id3c etl redcap-det swab-n-send
```

lib/id3c/cli/command/etl/redcap_det.py
tsibley: It's a set. There are two ways to make a set in Python:

1. The `set()` constructor which takes an iterable.
2. Set literal syntax, which is `{ value, value, value, … }`. It is similar to dict literals (`{ key: value, key: value, … }`) but omits the `key:` part.

Set literals are used in various places around ID3C.

kairstenfay: I thought it was a set but then I doubted myself :) Thanks!

lib/id3c/cli/command/etl/redcap_det.py
tsibley: > How does this function allow other ways than the environment to provide values when the environment variable appears to be hard-coded…?

It does not allow other ways at the moment, but using a function as a credentials provider means it is easier to swap out later without changing calling code or repeating environment variable names in many places.

> Each REDCap project requires a different API token, and I don't see why this revision is preferred.
> Would the REDCAP_API_TOKEN need to be declared before running every etl?

Parameterizing the environment variable name on the REDCap project feels like an antipattern to me. A consistent environment variable for specifying the token makes it predictable and less to document and remember. Rather than setting all environment variables at once, I expect that each project's will be set in separate invocations (like your example) or with separate environment files or envdirs for each project. For example, the crontab we'll end up with might look like:

```
REDCAP_API_URL=https://redcap.iths.org

0 * * * * REDCAP_API_TOKEN=deadbeef pipenv run id3c etl redcap-det kiosk
5 * * * * REDCAP_API_TOKEN=abadcafe pipenv run id3c etl redcap-det swab-n-send
…
```

or

```
0 * * * * envdir /opt/secrets/redcap-kiosk/ pipenv run id3c etl redcap-det kiosk
5 * * * * envdir /opt/secrets/redcap-swab-n-send/ pipenv run id3c etl redcap-det swab-n-send
…
```

or

```
0 * * * * source /opt/secrets/redcap-kiosk.env && pipenv run id3c etl redcap-det kiosk
5 * * * * source /opt/secrets/redcap-swab-n-send.env && pipenv run id3c etl redcap-det swab-n-send
…
```

Environment variables are global variables (albeit special ones), and just like with global variables, it's good to minimize both the number of different ones and the number of places they're accessed in the code.

kairstenfay: Thanks for the examples! You have convinced me.

PR #82, opened 2019-10-29
Geocode, location lookup, and de-identify

joverlee521:
Adds `geocode`, `location lookup` and `de-identify` to ID3C. 
See commit messages for specifics of each command. 

A couple things I would like to discuss:
1. the names of the `geocode` subcommands (naming is hard 😭)
2. whether we should limit `location lookup` to return only one location
3.  where should PII be standardized? currently `de-identify` expects values to already be canonicalized.

### DIFF
--- a/lib/id3c/cli/command/geocode.py
+++ b/lib/id3c/cli/command/geocode.py
@@ -1,0 +1,445 @@
+"""
+Geocode addresses that are listed in tabular format, with one address per
+row.
+"""
+import click
+import logging
+import pickle
+import pandas as pd
+import sys
+import json
+import yaml
+from os import environ, chdir
+from os.path import dirname
+from textwrap import dedent
+from cachetools import TTLCache
+from typing import Optional, Tuple, Dict, Any
+from smartystreets_python_sdk import StaticCredentials, ClientBuilder
+from smartystreets_python_sdk.us_street import Lookup
+from id3c.cli import cli
+
+
+LOG = logging.getLogger(__name__)
+
+CACHE_TTL = 60 * 60 * 24 * 365  # 1 year
+
+@cli.group("geocode", help = __doc__)
+def geocode():
+    pass
+
+@geocode.command("using-options")
+@click.argument("filename",
+    metavar = "<filename.{csv,tsv,xlsx,xls}>",
+    required = True,
+    type = click.Path(exists=True))
+
+@click.option("--street-column",
+    metavar = "<column>",
+    help = "Column name for address street",
+    required = True)
+
+@click.option("--secondary-column",
+    metavar = "<column>",
+    help = "Column name for address apartment,suite, or office number",
+    required = False)
+
+@click.option("--city-column",
+    metavar = "<column>",
+    help = "Column name for address city",
+    required = True)
+
+@click.option("--state-column",
+    metavar = "<column>",
+    help = "Column name for address state",
+    required = True)
+
+@click.option("--zipcode-column",
+    metavar = "<column>",
+    help = "Column name for address zipcode",
+    required = True)
+
+@click.option("--cache-file",
+    metavar = "<cache.pickle>",
+    help = "Local cache file for storing address lookups. Must be specified to cache lookups",
+    required = False,
+    type = click.Path())
+
+def geocode_using_options(**kwargs):
+    """
+    Geocode addresses listed in <filename.{csv,tsv,xlsx,xls}>.
+
+    Options specify column names to extract addresse parts for geocoding.
+    Of these, --street, --city, --state, and --zipcode are required.
+
+    Requires two environment variables SMARTYSTREETS_AUTH_ID and
+    SMARTYSTREETS_AUTH_TOKEN.
+
+    Geocoded addresses are output to stdout as comma-separated values, with
+    extra values for addresses latitude, longitude, and canonicalized address.
+    If an address is invalid, then these extra columns are left blank.
+    """
+    geocoded_addresses = get_geocoded_addresses(**kwargs)
+    geocoded_addresses.to_csv(sys.stdout, index=False)
+
+
+@geocode.command("using-config")
+@click.argument("filename",
+    metavar = "<filename.{csv,tsv,xlsx,xls}>",
+    required = True,
+    type = click.Path(exists=True))
+
+@click.argument("config_file",
+    metavar = "<config.yaml>",
+    required = True,
+    type = click.File("r"))
+
+def geocode_using_config(filename, config_file):
+    """
+    Geocode addresses listed in <filename.{csv,tsv,xlsx,xls}>
+    with column names and cache file specified by a <config.yaml>.
+    If cache is not specified in <config.yaml>, then lookups will not be
+    cached locally and will not be re-used for future lookups.
+
+    <config.yaml> must be a file with one YAML document in it.  The document
+    corresponds closely to the command-line options taken by the
+    "using-options" command (a sibling to this command).
+    For example:
+
+    \b
+        ---
+        cache: cache.pickle
+        columns:
+          street: "Street"
+          secondary: "Street2"
+          city: "City"
+          state: "State"
+          zipcode: "ZipCode"
+
+    Relative paths in <config.yaml> are treated relative to the containing
+    directory of the configuration file itself.
+
+    Requires two environment variables: SMARTYSTREETS_AUTH_ID and
+    SMARTYSTREETS_AUTH_TOKEN.
+
+    Geocoded addresses are output to stdout as comma-separated values, with
+    extra values for addresses latitude, longitude, and canonicalized address.
+    If an address is invalid, then these extra columns are left blank.
+    """
+    configs = list(yaml.safe_load_all(config_file))
+
+    if config_file.name != "<stdin>":
+        config_dir = dirname(config_file.name)
+
+        # dirname is the empty string if we're in the same directory as the
+        # config file.
+        if config_dir:
+            chdir(config_dir)
+
+    for config in configs:
+        try:
+            kwargs = {
+                "filename":          filename,
+                "street_column":     config["columns"]["street"],
+                "city_column":       config["columns"]["city"],
+                "state_column":      config["columns"]["state"],
+                "zipcode_column":    config["columns"]["zipcode"],
+                "secondary_column":  config["columns"].get("secondary"),
+                "cache_file":        config.get("cache")
+            }
+        except KeyError as key:
+            LOG.error(f"Required key «{key}» missing from config {config}")
+            raise key from None
+
+        geocoded_addresses = get_geocoded_addresses(**kwargs)
+        geocoded_addresses.to_csv(sys.stdout, index=False)
+
+
+def get_geocoded_addresses(*,
+                           filename,
+                           street_column,
+                           city_column,
+                           state_column,
+                           zipcode_column,
+                           secondary_column = None,
+                           cache_file = None):
+    """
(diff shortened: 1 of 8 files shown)


joverlee521:
[Import geocoding tool into id3c](https://trello.com/c/agqRcoNb/243-import-geocoding-tool-into-id3c)

joverlee521:
[Generate household hashes as part of deidentification routines](https://trello.com/c/TXKCZVLP/134-generate-household-hashes-as-part-of-deidentification-routines)

lib/id3c/cli/command/geocode.py
```diff
@@ -0,0 +1,435 @@
...
+    trailing whitespaces.
+    """
+    address_columns = list(filter(None, address_column_map.values()))
+
+    if not address_columns:
+        raise NoAddressColumnsFoundError(address_column_map)
+
+    addresses_df = pd.DataFrame()
+
+    for dst, src in address_column_map.items():
```
kairstenfay: What do `dst` and `src` mean? I would prefer to lengthen the variable names here.

lib/id3c/cli/command/geocode.py
```diff
@@ -0,0 +1,435 @@
...
+
+    smartystreets_responses = pd.DataFrame(columns = ['lat','lng','canonicalized_address'])
+
+    for address in standardized_addresses:
+        response = get_response_from_cache_or_geocoding(address, cache)
+
+        smartystreets_responses = smartystreets_responses \
+            .append(response,ignore_index = True)
+
+    geocoded_addresses = address_file_df.join(smartystreets_responses)
```
kairstenfay: I wonder if there's a better way to geocode addresses than to convert from DataFrame -> dict, create an empty DataFrame, loop over the dict while appending to the new DataFrame, then joining to a DataFrame again. 

Perhaps we could standardize the address on the original DataFrame and use an `apply` function on it that creates several new columns from geocoding.

lib/id3c/cli/command/geocode.py
```diff
@@ -0,0 +1,435 @@
...
+        raise UnsupportedFileExtensionError(dedent(f"""
+            Unsupported file extension for file «{filename}».
+            Please choose from one of the following file extensions:
+                {SUPPORTED_EXTENSIONS}
+            """
+        ))
+
+    if filename.endswith(('.csv', '.tsv')):
+        separator = '\t' if filename.endswith('.tsv') else ','
+        df = pd.read_csv(filename, sep=separator, dtype = str)
```
kairstenfay: this is a nitpick but I prefer the style of using no spaces around named function arguments. Either is fine but there's a mix of each in this line of code.

lib/id3c/cli/command/geocode.py
```diff
@@ -0,0 +1,435 @@
...
+    else:
+        df = pd.read_excel(filename, dtype = str)
```
kairstenfay: ditto (whichever style is favored above should be applied here)

lib/id3c/cli/command/geocode.py
```diff
@@ -0,0 +1,435 @@
...
+def load_or_create_cache() -> TTLCache:
+    """
+    Tries to load a pickled cache from the filepath `cache.pickle`. If a cache
+    is not found at this location, creates a new one. Returns the cache.
+    """
+    try:
+        cache = pickle.load(open('cache.pickle', mode='rb'))
+    except FileNotFoundError:
+        LOG.warning("Couldn't find an existing cache file. Creating new cache.")
+        cache = TTLCache(maxsize=100000, ttl=CACHE_TTL)
```
kairstenfay: Should we increase the `maxsize`? AFAIK this number is completely arbitrary.

lib/id3c/cli/command/geocode.py
```diff
@@ -159,7 +162,7 @@ def get_geocoded_addresses(*,
     """
     LOG.debug(f"Reading addresses file {filename}")
 
-    address_file_df = load_file_as_dataframe(filename)
+    address_file_df = load_file_as_dataframe(filename, SUPPORTED_EXTENSIONS)
```
kairstenfay: I'm not sure I agree with adding a dynamic variable for supported extensions here.
I think we should avoid changing functions while moving them.
Also, the extensions argument gives the illusion of being able to handle multiple file extensions, but in reality, we're only handling `tsv`, `csv`, and `xls*`.

lib/id3c/cli/command/location.py
```diff
@@ -289,3 +292,161 @@ def parse_features(path):
...
+
+    for lat_lng in lat_lngs:
+        location = location_lookup(db, lat_lng, scale)
+        locations.append(location.identifier if location else None)
+
+    if drop_input_columns:
+        output_df = pd.DataFrame(locations, columns = [f'{scale}_identifier'])
+    else:
+        output_df = input_df
+        output_df[f'{scale}_identifier'] = locations
```
kairstenfay: I would prefer to create the `output_df` DataFrames the same way and then to conditionally drop the input columns later. 
Or, conversely, to create the DataFrame without the input columns and re-attach them only if they are requested.

lib/id3c/cli/command/location.py
```diff
@@ -289,3 +292,161 @@ def parse_features(path):
...
+
+    if not point:
+        LOG.error("Cannot find location without point created from lat/lng")
+        return None
+
+    location = db.fetch_row("""
+        select location_id as id, identifier
+          from warehouse.location
+         where scale = %s and
+               st_contains(polygon, st_setsrid(st_geomfromgeojson(%s), 4326))
```
kairstenfay: Could we please add a comment to this magic number?

lib/id3c/cli/command/location.py
```diff
@@ -335,7 +334,7 @@ def lookup(scale,
     location identifier as <scale>_identifier. Original input can be dropped
     with the --drop-input-columns flag.
     """
-    input_df = load_lookup_input(filename)
+    input_df = load_input_from_file_or_stdin(filename, SUPPORTED_EXTENSIONS)
```
kairstenfay: I have the same concerns here as the other load function: 
1. not to change functions while moving them, and 
2. that this falsely gives the illusion of being able to support more extensions than it actually does. (reality is we only support csv or tsv within the body of the function)

lib/id3c/cli/command/de-identify.py
```diff
@@ -0,0 +1,89 @@
...
+    joined_fields = fields_to_include.apply(lambda x: ' '.join(x), axis=1)
+
+    hash_secret = os.environ['PARTICIPANT_DEIDENTIFIER_SECRET']
+    hashes = joined_fields.map(lambda x: generate_hash(x, hash_secret) if x else None)
+
+    if drop_input_columns:
+        output_df = pd.DataFrame(hashes, columns = ['hash'])
+    else:
+        output_df = input_df
+        output_df['hash'] = hashes
```
kairstenfay: Similar comment as before: 
I think I would prefer to build the output dataframes the same way, dropping or appending the desired columns after construction.

tsibley: Agreed. The current implementation of `--drop-input-columns` means that all columns are dropped, not just those specified as arguments / input to the hash.

lib/id3c/cli/command/de-identify.py
```diff
@@ -0,0 +1,89 @@
...
+LOG = logging.getLogger(__name__)
+SUPPORTED_EXTENSIONS = ('.csv', '.tsv', '.xls', '.xlsx')
+
+@cli.command("de-identify")
+@click.argument("columns",
+    metavar = "<columns>",
+    nargs = -1)
+@click.option("--filename",
+    metavar = "<filename.{csv,tsv,xls,xlsx}>",
+    type = click.Path(exists=True))
```
tsibley: I think a more conventional/Unixy interface here would be `id3c de-identify column1 column2 path/to/file.csv` or `id3c de-identify column1 column2 -` for stdin. This can be supported with something like:

```python
@click.option("filename",
    metavar = "<filename.{csv,tsv,xls,xlsx}>",
    type = click.File("r"))
```

lib/id3c/cli/command/de-identify.py
```diff
@@ -0,0 +1,89 @@
...
+        LOG.error(f"{error}. Input columns: {list(input.columns)}")
+        raise error from None
+
+    return fields
+
+
+def generate_hash(identifier: str, secret: str) -> str:
+    """
+    Hash *secret* with *identifier* that is linked to identifiable records.
+    """
```
tsibley: For safety and defensiveness, let's add an `assert len(identifier)` (it's important to never hash just the secret) and `assert len(secret)` (it's important to always have a secret).

lib/id3c/cli/command/geocode.py
```diff
@@ -0,0 +1,435 @@
...
+    SmartyStreets API, stores them as a key-value pair in the given *cache*,
+    overwriting the value for the existing *standardized_address* key if it
+    already existed in the *cache*.
+    """
+    cache[json.dumps(standardized_address, sort_keys=True)] = response
+
+
+def save_cache(cache: TTLCache):
+    """ Given a *cache*, saves it to a hard-coded file `cache.pickle`. """
+    pickle.dump(cache, open('cache.pickle', mode='wb'))
```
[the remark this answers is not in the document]
joverlee521: I think I like option 3 the best. Yes, it's more to manage for the user, but I think it's better to be explicit about where the cache exists. Also seems to makes it easier to manage if they want to use the cache or not.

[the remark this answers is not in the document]
tsibley: Sounds good to me.

lib/id3c/cli/command/geocode.py
joverlee521: Do we want `maxsize=float("inf")` like shown here https://github.com/tkem/cachetools/issues/73 ?

tsibley: Ah, yes, that looks great.

lib/id3c/cli/command/location.py
```diff
@@ -289,3 +291,161 @@ def parse_features(path):
...
+    """
+    input_df = load_lookup_input(filename)
+    lat_lngs = extract_lat_lng_from_input(input_df, lat_column, lng_column)
+
+    db = DatabaseSession()
+    locations = []
+
+    for lat_lng in lat_lngs:
+        location = location_lookup(db, lat_lng, scale)
+        locations.append(location.identifier if location else None)
```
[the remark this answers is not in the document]
joverlee521: Not entirely sure how to change this to do a single query?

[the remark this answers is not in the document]
tsibley: There are a couple approaches, one of which is to construct a table of points being looked up using a [`with` CTE](https://www.postgresql.org/docs/10/queries-with.html) and then join that table to warehouse.location on `st_contains`. The points being looked up would be passed in as a nested array, I think.

That said, I think it's fine to put this aside for now and wait to see if it's an issue. I suspect it will make a large difference in runtime, but good to wait and see if that difference is actually material to us. There are also intermediate approaches like re-using prepared statements to avoid each query's overhead while still issuing lots of queries.

joverlee521:
@tsibley @kairstenfay I've made changes based on your feedback and rebased onto master. Ready for round two of reviews!

tsibley:
Since @joverlee521 is out this week, I'm going to pick up this PR and just make the small tweaks I'd normally make review notes. @kairstenfay If you have time/inclination, I'd be happy to have you take a look at my tweaks before merge.

tsibley:
I think this PR is now ready for merge.

PR #83, opened 2019-10-29
Update etl presence-absence

joverlee521:
Samplify will be sending documents with a top level "samples" key and
a "chip" key for each sample. 

~For each presence/absence result, the unique identifier will become "NWGC/`sampleId`/`chip`/`target_id`". `sampleId` and `chip` will both come from Samplify.
`target_id` for the target found or created within ID3C `warehouse.target`.~

`presence_absence.identifier` will become `NWGC/chip/sampleId/geneTarget` with all parts of the identifier coming from the Samplify document.

This change does bring up the question of how should we handle the old presence/absence results already within `warehouse.presence_absence`. Matthew did [offer](https://seattle-flu-study.slack.com/archives/CJU8RN2Q6/p1571943038028500?thread_ts=1571942534.026100&cid=CJU8RN2Q6) to send previous runs with chip information. In order for the results to be consistent and updated appropriately, I think it would be best to update the identifier for all presence/absence results to fit the new format.

### DIFF
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -118,6 +118,9 @@ def etl_presence_absence(*, action: str):
                     received_sample_id = str(received_sample["sampleId"])
                     chip = received_sample["chip"]
 
+                    # Guard against empty chip values
+                    assert chip, "Received bogus chip id"
+
                     for test_result in received_sample["targetResults"]:
                         test_result_target_id = test_result["geneTarget"]
                         LOG.debug(f"Processing target «{test_result_target_id}» for \
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -261,7 +261,7 @@ def update_sample(db: DatabaseSession,
     LOG.debug(f"Looking up sample «{identifier}»")
 
     sample = db.fetch_row("""
-        select sample_id as id, identifier
+        select sample_id as id, identifier, details
           from warehouse.sample
          where identifier = %s
            for update
@@ -272,6 +272,11 @@ def update_sample(db: DatabaseSession,
         raise SampleNotFoundError(identifier)
 
     LOG.info(f"Found sample {sample.id} «{sample.identifier}»")
+
+    if (sample.get("details") and
+        sample.details.get("nwgc_id") and
+        additional_details):
+        update_details_nwgc_id(sample, additional_details)
 
     if additional_details:
         LOG.info(f"Updating sample {sample.id} «{sample.identifier}» details")
@@ -286,6 +291,25 @@ def update_sample(db: DatabaseSession,
         assert sample.id, "Updating details affected no rows!"
 
     return sample
+
+
+def update_details_nwgc_id(sample: Any, additional_details: dict) -> None:
+    """
+    Given a *sample* fetched from `warehouse.sample`,
+    extend `sample.details.nwgc_id` to an array if needed.
+
+    Add provided "nwgc_id" within *additional_details* to the existing array
+    if it doesn't already exist
+    """
+    exisiting_nwgc_ids = sample.details["nwgc_id"]
+    new_nwgc_ids = additional_details["nwgc_id"]
+
+    # Extend details.nwgc_id to an array
+    if not isinstance(sample.details["nwgc_id"], list):
+        exisiting_nwgc_ids = [sample.details["nwgc_id"]]
+
+    # Concatenate exisiting and new nwgc_ids and deduplicate
+    additional_details["nwgc_id"] = list(set(exisiting_nwgc_ids + new_nwgc_ids))
 
 
 def sample_identifier(db: DatabaseSession, barcode: str) -> Optional[str]:
@@ -308,7 +332,7 @@ def sample_details(document: dict) -> dict:
     Capture details about the go/no-go sequencing call for this sample.
     """
     return {
-        "nwgc_id": document['sampleId'],
+        "nwgc_id": [document['sampleId']],
         "sequencing_call": {
             "comment": document['sampleComment'],
             "initial": document['initialProceedToSequencingCall'],
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -64,6 +64,7 @@ def etl_presence_absence(*, action: str):
     LOG.debug("Fetching unprocessed presence-absence tests")
 
     presence_absence = db.cursor("presence_absence")
+    presence_absence.itersize = 1
     presence_absence.execute("""
         select presence_absence_id as id, document
           from receiving.presence_absence
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -106,6 +106,10 @@ def etl_presence_absence(*, action: str):
                     received_sample_barcode = received_sample["investigatorId"]
                     LOG.info(f"Processing sample «{received_sample_barcode}»")
 
+                    if not received_sample.get("isCurrentExpressionResult"):
+                        LOG.warning(f"Skipping out-of-date results for sample «{received_sample_barcode}»")
+                        continue
+
                     received_sample_identifier = sample_identifier(db, received_sample_barcode)
 
                     if not received_sample_identifier:
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -301,15 +301,15 @@ def update_details_nwgc_id(sample: Any, additional_details: dict) -> None:
     Add provided "nwgc_id" within *additional_details* to the existing array
     if it doesn't already exist
     """
-    exisiting_nwgc_ids = sample.details["nwgc_id"]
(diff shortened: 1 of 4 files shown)


joverlee521:
[Update presence-absence etl to take new NWGC data format](https://trello.com/c/cYT03tdo/257-update-presence-absence-etl-to-take-new-nwgc-data-format)

joverlee521:
>How do you foresee us updating the existing presence-absence identifiers to use the new format? Our current upsert_presence_absence function looks for conflicts on identifier.

The only way I can think of is to delete all presence/absence records and ask Matthew to resend all previous runs in the new format, but this sounds way too extreme to me. 

We _could_ not change how the identifier works, but make a new "chip" key within `sample.details` and have the upsert look for conflicts on (sample_id, details.chip).

joverlee521:
> > We _could_ not change how the identifier works, but make a new "chip" key within `sample.details` and have the upsert look for conflicts on (sample_id, details.chip).
> 
> Interesting. What does a "chip" mean IRL?

We would have to confirm with Matthew, but I'm assuming "chip" is the actual chip that the presence/absence test was run on. Therefore if we get a new result for the same sample on the same chip then it's just an updated result, but a new result for the same sample on a different chip means they ran the sample through a different presence/absence test.

joverlee521:
> Thanks. How did we disambiguate multiple runs vs. updated results previously?

I don't think we had a way to differentiate, I think that's the reason for the new format.

joverlee521:
@kairstenfay and I talked in person and concluded that it would make sense to change `presence_absence.identifier` to `NWGC/chip/sampleId/geneTarget` so that all of the parts of the identifier are from external sources.

tsibley:
Hmm, there are several options here. It's not entirely clear to me which is best, but trying to synthesize what's in #data-transfer-nwgc [since this message on 1 Oct](https://seattle-flu-study.slack.com/archives/CJU8RN2Q6/p1569948099006900) and keep it simple, how about:

- [ ] Switch `presence_absence.identifier` to `NWGC/{sampleId}/{geneTarget}` because we can no longer trust that the `id` of each target result won't drift. According to Matthew, there is a one to many correspondence between our samples and their internal sample ids in the case of replicate runs, so we'll still collect multiple presence/absence results for a single (sample, target) pair. We'll be able to pick one in queries based on timestamps, our internal serial pk, or some other criteria that makes sense. Redos will, if I understand correctly, use the same (sample, target) pair, updating the original result.

- [ ] Handle existing records using the `NWGC_{targetId}` format by doing an `update warehouse.presence_absence set identifier = {new} where identifier = {old}` before our `upsert` in the ETL. This will require bumping the ETL revision.

- [ ] While we're at it, turn `sample.details->'nwgc_id'` into an array ([Trello card](https://trello.com/c/TP1Rxdiy/228-extend-sampledetailsnwgcid-to-an-array))

Thoughts? Have I understood everything correctly?

joverlee521:
Added `update_presence_absence_identifier` to the ETL. When I ran this locally, there was a UniqueViolation, so I think we already have undetected duplicates within `warehouse.presence_absence`

tsibley:
Yeah. A lot.

```sql
select
    "sampleId" as nwgc_sample,
    "geneTarget" as target,
    jsonb_agg(distinct to_jsonb(array[presence_absence_id, result.id])) as dups

from receiving.presence_absence,

    jsonb_to_recordset(coalesce(document->'store'->'items', document->'Update')::jsonb)
        as sample("sampleId" int, "investigatorId" text, "targetResults" jsonb),

    jsonb_to_recordset("targetResults")
        as result("id" int, "geneTarget" text)

group by 1, 2 having count(distinct result.id) > 1;
```

returns ~25k rows.

tsibley:
Those 25k results from the query above represent ~53k rows in `warehouse.presence_absence`.

tsibley:
Thinking about this some more, it maybe isn't surprising. We've been using individual target result ids, so we've been creating a new `warehouse.presence_absence` row for every redo. But [my proposal above](https://github.com/seattleflu/id3c/pull/83#issuecomment-550027285) changes that to having redos update the result _in place_ in `warehouse.presence_absence`.

Not sure best way forward. We'll have to think about this some more.

tsibley:
Latest thoughts from @joverlee521 and I are: https://seattle-flu-study.slack.com/archives/CL1Q38ATG/p1573754334073200

joverlee521:
Waiting for Matthew to send all previous runs with `chip` to testing environment. Then we'll see if there are any orphan presence/absence records with the old identifier format.

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -80,16 +80,26 @@ def etl_presence_absence(*, action: str):
...
+                    #   -Jover, 14 Nov 2019
+                    if (group.document.get("store") is not None or
+                        group.document.get("Update") is not None):
+
+                        LOG.info("Skipping presence_absence record that is in old format")
+                        mark_processed(db, group.id)
+                        continue
+
+                    else:
+                        raise error from None
```
tsibley: My first thought was that it'd be slightly more robust/future-proof to skip each _sample_ `if not received_sample.get("chip")` instead of skipping each _document_ based on the container/collection format (which assumes the container format is fundamentally linked to the inclusion of `chip`).

Upon further consideration though, I think I prefer the way you have it :smile:, because it will fail loudly if we get a sample that doesn't include a `chip`.

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -106,6 +116,7 @@ def etl_presence_absence(*, action: str):
                         additional_details = sample_details(received_sample))
 
                     received_sample_id = str(received_sample["sampleId"])
+                    chip = received_sample["chip"]
```
tsibley: For extra robustness, how about an `assert chip` so that it must have a truthy value?

lib/id3c/cli/command/etl/presence_absence.py
joverlee521: I hadn't thought that far ahead, but glad my way works 😆

joverlee521:
[Extend sample.details.nwgc_id to an array](https://trello.com/c/TP1Rxdiy/228-extend-sampledetailsnwgcid-to-an-array)

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -288,6 +291,27 @@ def update_sample(db: DatabaseSession,
...
+    new_nwgc_ids = additional_details["nwgc_id"]
+
+    # Extend details.nwgc_id to an array
+    if isinstance(sample.details["nwgc_id"], int):
+        exisiting_nwgc_ids = [sample.details["nwgc_id"]]
+
+    # Concatenate exisiting and new nwgc_ids and deduplicate
+    additional_details["nwgc_id"] = list(set(exisiting_nwgc_ids + new_nwgc_ids))
+
+    return additional_details
```
tsibley: This function modifies `additional_details` in place, so returning it is unnecessary. I think it's a good habit to omit return values from functions which use in-place modification because it makes it more clear how values flow in the code.

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -288,6 +291,27 @@ def update_sample(db: DatabaseSession,
...
+
+    Add provided "nwgc_id" within *additional_details* to the existing array
+    if it doesn't already exist
+    """
+    exisiting_nwgc_ids = sample.details["nwgc_id"]
+    new_nwgc_ids = additional_details["nwgc_id"]
+
+    # Extend details.nwgc_id to an array
+    if isinstance(sample.details["nwgc_id"], int):
+        exisiting_nwgc_ids = [sample.details["nwgc_id"]]
```
tsibley: I think better here to encapsulate in a list _any_ value that isn't already a list, not just ints specifically, e.g. making the condition `if not isinstance(…, list):`.

For example, if one of the NGWC ids is a string, then a TypeError will be thrown because of the `+` operator below. I realize that's not the case right now, but it could easily happen in the future and good to be defensive.

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -273,6 +273,9 @@ def update_sample(db: DatabaseSession,
 
     LOG.info(f"Found sample {sample.id} «{sample.identifier}»")
 
+    if sample.details.get("nwgc_id") and additional_details:
+        additional_details = update_details_nwgc_id(sample, additional_details)
+
```
tsibley: Might be good to be defensive against the case that `sample.details` is null/None.

joverlee521: Not sure what you mean here. If `sample.details` is null/None, then it gets set to the value of `additional_details` which already contains `nwgc_id` as a list. I can add in a comment to make this more apparent.

tsibley: Ah, sorry, I should have used more words. I mean that if `sample.details` is `None` then `sample.details.get(…)` will fail.

tsibley:
Thanks! I'm testing this now against the docs pushed to our testing environment by Matthew yesterday.

tsibley:
Pushed several commits here which work on the data Matthew pushed to testing and would appreciate review.

joverlee521:
> Pushed several commits here which work on the data Matthew pushed to testing and would appreciate review.

Looks good to me. Are we ready to merge?

PR #85, opened 2019-11-07
Load extra API routes from extensions

joverlee521:
Loads all extra API endpoints from extensions through entry point "id3c.api.routes"

This must be at the end of route declarations to allow customization of existing routes and to avoid dependency errors.

Required for PR https://github.com/seattleflu/id3c-customizations/pull/16

### DIFF
--- a/lib/id3c/api/routes.py
+++ b/lib/id3c/api/routes.py
@@ -3,6 +3,7 @@ API route definitions.
 """
 import json
 import logging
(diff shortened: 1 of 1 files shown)


joverlee521:
[Connect return of results to ID3C](https://trello.com/c/m9uMWQYM/220-connect-return-of-results-to-id3c)

PR #89, opened 2019-11-23
etl presence-absence: Preserve existing details when updating an existing row

tsibley:
Our custom reporting process stashes its log entries in presence/absence
details, but the log was being blown away when updates came in.  This
caused mass re-reporting today as we processed a large repush of
existing data.  More thorough testing (mea culpa) all the way through
the reporting process would have caught this.

This coalesce + append is the general pattern we should use so that
multiple things can cooperatively use the details column without knowing
about each other.

---

I haven't tested this yet, but we should!


tsibley:
Now tested locally and confirmed that it does indeed preserve existing reporting logs.

tsibley:
Going to merge and deploy this so we don't re-notify accidentally anymore, but I'm happy to make review-driven changes after the fact.

PR #86, opened 2019-11-12
Etl fhir

kairstenfay:
Ready for review! I have a couple of lingering TODOs related to using test data plus some leftover conceptual questions. 99% of the work here is done!

PR #90, opened 2019-11-27
FHIR ETL

kairstenfay:
GitHub decided to close the old PR for this, https://github.com/seattleflu/id3c/pull/86.

I'll still reference it for bug-fixing and other changes, but moving forward, we should use this PR.

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -126,6 +126,8 @@ def process_bundle_entries(db: DatabaseSession, bundle: Bundle):
     Loads Encounter, DiagnosticReport, and other dependent FHIR DomainResources
     from a given *Bundle* into the database warehouse as appropriate for each
     Resource type.
+    Ensures that all Encounter resources are processed before DiagnosticReport
+    resources to avoid potential :class:`SampleNotFoundError`s.
     """
     for entry in bundle.entry:
         resource, resource_type = resource_and_resource_type(entry)
@@ -141,7 +143,10 @@ def process_bundle_entries(db: DatabaseSession, bundle: Bundle):
             process_encounter_samples(db, resource, encounter.id, related_resources)
             process_locations(db, encounter.id, resource)
 
-        elif resource_type == 'DiagnosticReport':
+    for entry in bundle.entry:
+        resource, resource_type = resource_and_resource_type(entry)
+
+        if resource_type == 'DiagnosticReport':
             for reference in resource.specimen:
                 if not matching_system(reference.identifier, INTERNAL_SYSTEM):
                     continue
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -1,0 +1,582 @@
+"""
+Process FHIR documents into the relational warehouse
+"""
+import re
+import json
+import click
+import logging
+from collections import defaultdict
+from datetime import datetime, timezone
+from typing import Any, List, Dict, Optional, Tuple
+from fhir.resources.bundle import Bundle, BundleEntry
+from fhir.resources.codeableconcept import CodeableConcept
+from fhir.resources.domainresource import DomainResource
+from fhir.resources.encounter import Encounter
+from fhir.resources.identifier import Identifier
+from fhir.resources.location import Location
+from fhir.resources.observation import Observation
+from fhir.resources.patient import Patient
+from fhir.resources.questionnaireresponse import QuestionnaireResponse
+from fhir.resources.specimen import Specimen
+from id3c.cli.command import with_database_session
+from id3c.db.session import DatabaseSession
+from id3c.db.datatypes import Json
+from . import (
+    etl,
+
+    find_or_create_site,
+    find_location,
+    upsert_individual,
+    upsert_encounter,
+    upsert_encounter_location,
+    upsert_location,
+    upsert_sample,
+)
+
+
+LOG = logging.getLogger(__name__)
+
+
+# The revision number and etl name are stored in the processing_log of each
+# FHIR record when the FHIR document is successfully processed by this ETL
+# routine. The routine finds new-to-it records to process by looking for FHIR
+# documents lacking this etl revision number and etl name in their log.
+# If a change to the ETL routine necessitates re-processing all FHIR documents,
+# this revision number should be incremented.
+# The etl name has been added to allow multiple etls to process the same
+# receiving table
+REVISION = 1
+ETL_NAME = 'fhir'
+INTERNAL_SYSTEM = 'https://seattleflu.org'
+LOCATION_RELATION_SYSTEM = 'http://terminology.hl7.org/CodeSystem/v3-RoleCode'
+TARGET_SYSTEM = 'http://snomed.info/sct'
+
+@etl.command("fhir", help = __doc__)
+
+@with_database_session
+def etl_fhir(*, db: DatabaseSession):
+    LOG.debug(f"Starting the FHIR ETL routine, revision {REVISION}")
+
+    # Fetch and iterate over FHIR documents that aren't processed
+    #
+    # Use a server-side cursor by providing a name.  This ensures we limit how
+    # much data we fetch at once, to limit local process size.
+    #
+    # Rows we fetch are locked for update so that two instances of this
+    # command don't try to process the same FHIR documents.
+    LOG.debug("Fetching unprocessed FHIR documents")
+
+    fhir_documents = db.cursor("fhir")
+    fhir_documents.execute("""
+        select fhir_id as id, document
+          from receiving.fhir
+         where not processing_log @> %s
+         order by id
+           for update
+        """, (Json([{ "etl": ETL_NAME, "revision": REVISION }]),))
+
+    for record in fhir_documents:
+        with db.savepoint(f"FHIR document {record.id}"):
+            LOG.info(f"Processing FHIR document {record.id}")
+
+            assert_bundle_collection(record.document)
+            bundle      = Bundle(record.document)
+            resources   = extract_resources(bundle)
+
+            assert_required_resource_types_present(resources)
+
+            # Loop over every Resource the Bundle entry, processing what is
+            # needed along the way.
+            for entry in bundle.entry:
+                resource, resource_type = resource_and_resource_type(entry)
+
+                if resource_type == 'Encounter':
+                    LOG.debug(f"Processing Encounter Resource «{entry.fullUrl}».")
+
+                    related_resources = extract_related_resources(bundle, entry)
+
+                    encounter = process_encounter(db, resource, related_resources)
+                    if not encounter:
+                        LOG.warning("Skipping encounter with insufficient information.")
+                        mark_skipped(db, record.id)
+                        continue
+
+                    process_encounter_samples(db, resource, encounter.id, related_resources)
+                    process_locations(db, encounter.id, resource)
+
+
+
+def assert_bundle_collection(document: Dict[str, Any]):
+    """
+    Raises an :class:`AssertionError` if the given *document* is not a Bundle
+    resource of type 'collection'.
+    """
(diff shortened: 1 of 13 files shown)


kairstenfay:
[Core ID3C ETL routine to import FHIR bundles](https://trello.com/c/MGHNJu4V/246-core-id3c-etl-routine-to-import-fhir-bundles)

kairstenfay:
[Test RedCap, FHIR etls for both kiosks and swab-n-send (production)](https://trello.com/c/uw1VbbXm/271-test-redcap-fhir-etls-for-both-kiosks-and-swab-n-send-production)

kairstenfay:
This is now ready for review :tada:

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,680 @@
...
+            except AssertionError:
+                LOG.info("FHIR document doesn't meet minimum requirements for processing.")
+                mark_skipped(db, record.id)
+                continue
+
+            # Loop over every Resource the Bundle entry, processing what is
+            # needed along the way.
+            try:
+                process_bundle_entries(db, bundle)
+
```
joverlee521: Need to add `mark_processed` after processing bundle

lib/id3c/cli/command/geocode.py
```diff
@@ -320,6 +326,7 @@ def geocode_address(address: dict) -> dict:
     if not CLIENT:
         CLIENT = smartystreets_client_builder().build_us_street_api_client()
 
+    print('setup client')
```
joverlee521: This gets printed to the output file when using `--log-ouput`

lib/id3c/cli/command/geocode.py
```diff
@@ -362,6 +369,7 @@ def us_street_lookup(address: dict) -> Lookup:
     SmartyStreets geocoding API is not present in the given *address* data.
     """
     lookup = Lookup()
+    print(address)
```
joverlee521: Also printed to output file. Also, we probably just don't want to print addresses to the logs?

lib/id3c/cli/command/geocode.py
kairstenfay: Oops! Thanks for catching that

lib/id3c/cli/command/geocode.py
kairstenfay: Right, those were debugging statements for me that slipped in.

lib/id3c/cli/command/etl/fhir.py
kairstenfay: Thanks.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,683 @@
...
+    for specimen in specimens:
+        barcode = identifier(specimen, f"{INTERNAL_SYSTEM}/sample")
+
+        if not barcode:
+            raise Exception("No barcode detectable. Either the barcode identification system is "
+                            f"not «{INTERNAL_SYSTEM}/sample», or the barcode value is empty, which "
+                            "violates the FHIR docs.")
+
+        upsert_sample(db,
+            collection_identifier   = barcode,
```
joverlee521: `upsert_sample` requires the full identifier, not just the barcode.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,683 @@
...
+    unique_codes = list(set(codes))
+    if len(unique_codes) > 1:
+        raise Exception(f"Expected only one Location code. Instead received {unique_codes}.")
+
+    return unique_codes[0]
+
+
+def process_sample(db: DatabaseSession, barcode: str) -> Any:
+    """ Given a *barcode*, returns its matching sample from ID3C. """
+    sample = find_sample(db, barcode)
```
joverlee521: `find_sample` requires the full identifier not just the barcode

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,683 @@
...
+        # updates.
+        target = find_or_create_target(db,
+            identifier  = target_identifier,
+            control     = False)
+
+        upsert_presence_absence(db,
+            identifier = f'{barcode}/{target_identifier}/{observation.device.identifier.value}',
+            sample_id = sample_id,
+            target_id = target.id,
+            present = observation.valueBoolean,
```
joverlee521: Just flagging that we need to be able to ingest both `valueBoolean` and `valueCodeableConcept` per [slack convo](https://seattle-flu-study.slack.com/archives/CPE5K1848/p1576523853002400)

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -16,12 +16,16 @@
...
 from id3c.db.session import DatabaseSession
 from id3c.db.datatypes import Json
+from id3c.cli.command.geocode import pickled_cache
 from . import etl
 
 
 LOG = logging.getLogger(__name__)
 
 
+CACHE_FILE = 'cache.pickle'
```
tsibley: I don't think we should hardcode a cache name like this, particularly with a name that doesn't give any hint as to what uses it or what it contains. The `id3c geocode` command, for instance, explicitly parameterizes the cache file as an option.

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -89,49 +93,50 @@ def decorated(*args, db: DatabaseSession, log_output: bool, **kwargs):
...
+                                if not is_complete(instrument, redcap_record)
+                        }
+
+                        if incomplete_instruments:
+                            LOG.debug(f"The following required instruments «{incomplete_instruments}» are not yet marked complete. " + \
+                                      f"Skipping REDCap DET {det.id}")
+                            mark_skipped(db, det.id, etl_id)
+                            continue
+
+                        bundle = routine(db = db, cache = cache, det = det, redcap_record = redcap_record)
```
tsibley: Similarly with the generic-ness of the `cache` parameter name here. I don't know if it's used for caching geocoding results, or REDCap records, or what. I don't think I'm convinced that the cache should be managed by this decorator instead of each routine which needs a cache.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,712 @@
...
+        resources[contained.resource_type].append(contained)
+
+    return resources
+
+
+def assert_required_resource_types_present(resources: Dict[str, List[DomainResource]]):
+    """
+    Raises an :class:`AssertionError` if the given *resources* do not meet the
+    following requirements:
+    * At least one Specimen Resource exists in the given *resources*
```
joverlee521: This doesn't work with the Bundle examples from Ellume. Do we want to loosen our expectations for bundles or ask Ellume to include a Specimen Resource?

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,712 @@
...
+    * There is at least one Patient or DiagnosticReport Resource
+    * The number of Observation Resources equals or exceeds the total number
+      of Encounter or Patient Resources.
+    """
+    def total_resources(resources: Dict[str, List[DomainResource]], resource_type: str) -> int:
+        return len(resources[resource_type]) if resources.get(resource_type) else 0
```
tsibley: By using the optional second argument to `dict.get`, there is a nicer pattern here:

```python
return len(resources.get(resource_type, []))
```

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,712 @@
...
+
+    assert resources.get('Specimen') and len(resources['Specimen']) >= 1, \
```
tsibley: Similarly here by using `.get(…, [])`

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -89,49 +97,56 @@ def decorated(*args, db: DatabaseSession, log_output: bool, **kwargs):
...
+                            mark_skipped(db, det.id, etl_id)
+                            continue
 
-                    if log_output:
-                        print(json.dumps(bundle, indent=2))
+                        if log_output:
+                            if indent:
+                                print(json.dumps(bundle, indent=indent))
+                            else:
+                                print(json.dumps(bundle))
```
tsibley: I think the default log output should be ndjson (so `indent=0`) and any reformatting should be done externally to ID3C (so dropping the `--indent` option). Programs such as `jq` make quick work of JSON manipulation from basic pretty-printing to filtering, searching, and modification.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,712 @@
...
+    for result in report.result:
+        observation = result.resolved(Observation)
+
+        target_identifier = matching_system_code(observation.code, TARGET_SYSTEM)
+        # Most of the time we expect to see existing targets so a
+        # select-first approach makes the most sense to avoid useless
+        # updates.
+        target = find_or_create_target(db,
+            identifier  = target_identifier,
+            control     = False)
```
joverlee521: This currently uses the SNOMED code as the `target.identifier`. Is this the behavior that we want, or do we want to map the SNOMED code to human readable identifiers?

tsibley: I think best to use the SNOMED code as the `target.identifier`, but importantly, it needs to be a fully-qualified SNOMED URL, not just the numeric `code` value (which is meaningless without qualifying the system it's from). https://www.hl7.org/fhir/snomedct.html#4.3.1.0.6 describes how to form those URLs, which is basically:

```python
system_concept_url = {
    "http://snomed.info/sct": "http://snomed.info/id/",
}
return urljoin(system_concept_url[concept.system], concept.code)
```

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -0,0 +1,740 @@
...
+                mark_skipped(db, record.id)
+                continue
+
+            # Loop over every Resource the Bundle entry, processing what is
+            # needed along the way.
+            try:
+                process_bundle_entries(db, bundle)
+
+            except AssertionError:
+                LOG.warning("Insufficient Encounter information.")
```
tsibley: Catching `AssertionError` here will catch more exceptions than the one you're anticipating from `process_encounter_bundle_entry()` (for example, there's an assert in `upsert_encounter`). I think a custom exception class is warranted here to more precisely target the catch. Ditto on the similar catch above, which is admittedly more tightly scoped. We want other assertions to still cause the whole process to bomb out and fail so we can investigate and intervene.

lib/id3c/cli/command/etl/fhir.py
tsibley: I would be nice to support processing of resources related to Specimens that only refer to them by identifier without including a stub Specimen resource.

tsibley: This seems like a blocker for deploy tomorrow.

lib/id3c/cli/command/etl/redcap_det.py
tsibley: This does not need to block deploy, but should be addressed later.

lib/id3c/cli/command/etl/redcap_det.py
tsibley: This does not need to block deploy, but should be addressed later.

lib/id3c/cli/command/etl/fhir.py
tsibley: This seems like a blocker for deploy tomorrow.

lib/id3c/cli/command/etl/fhir.py
tsibley: I think this is a blocker as it can cause silent-ish failure.

tsibley:
I commented on some previous notes above about which I think are "blockers" for tomorrow. I will try to address these tomorrow, unless someone beats me to it tonight. (Which I doubt since it's late already!)

PR #91, opened 2019-11-28
location: Require EPSG:4326 (WGS84) not EPSG:4269 (NAD83)

tsibley:
While I got the right CRS in the schema, the location command was
enforcing the wrong one.  I clearly mixed up the numbers.

To correct existing data, a manual update can be run which does:

    st_transform(st_setsrid(polygon, 4269), 4326)

for the point, polygon, and simplified_polygon columns.

For future data which isn't already in 4326 (like official Census tract
geometries), we'll need to add support for re-projection during location
import.  Alternatively, we can re-project before import as a
pre-processing step, but I think it'd be nicer to add built-in support.

### DIFF
--- a/lib/id3c/cli/command/location.py
+++ b/lib/id3c/cli/command/location.py
@@ -279,7 +279,7 @@ def parse_features(path):
     * tabular (CSV or TSV, delimiter auto-detected, no geometry construction)
 
     Raises a :class:`Exception` if the parsed coordinate reference system (CRS)
-    is defined and is not EPSG:4269.
+    is defined and is not EPSG:4326.
 
     Returns a list of dictionaries conforming to the GeoJSON ``Feature`` spec.
     """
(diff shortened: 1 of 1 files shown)


tsibley:
This is the SQL script I will use to fix the existing data:

```sql
begin;

update warehouse.location set
    polygon            = st_transform(st_setsrid(polygon, 4269), 4326),
    simplified_polygon = st_transform(st_setsrid(simplified_polygon, 4269), 4326),
    point              = st_transform(st_setsrid(point, 4269), 4326)
;
```

tsibley:
I pushed another commit which adds support for on-the-fly re-projection during location import. This will make things easier to import in the (near!) future.

I also tested locally that the SQL `update` statement above and re-importing all the Census tracts again with re-projection are equivalent fixes for our current data.

tsibley:
Merged, and I ran the update statement above on production and testing.

PR #94, opened 2019-12-27
Update census tract lookups in modeling views

kairstenfay:
Now that we're using FHIR document format, we no longer can rely on the
Audere-style of storing location information. Instead, rely on the
`warehouse.encounter_location` table to find residence and workplace
census tracts.


kairstenfay:
I'm attempting to work on other fixes to the shipping views, but having trouble processing the jsonb documents into regular arrays or text. 

I've tried:

- Symptoms
  To existing views (e.g. incidence_model_observation), I added
  ```
      lateral (
          select jsonb_agg(symptoms2->'id')
          from warehouse.encounter, 
          jsonb_array_elements(details->'Condition') as symptoms2
          group by encounter_id 
        ) as symptoms2,
   ```
   The problem is that the resulting column from the lateral join is of type `record` and looks like this:
   `("[""headaches"", ""chillsOrShivering"", ""soreThroat"", ""runnyOrStuffyNose""]")`. 

- Any QuestionnaireResponse result
  I started with a completely new query to learn how to parse a QuestionnaireResponse.
  ```
  select 
    encounter_id, 
    questionnaireResponse,
    quest,
    to_json(quest) as quest2,
    pg_typeof(quest)
    
    from warehouse.encounter, 
    
    lateral (
      select questionnaireResponse, to_json(jsonb_array_elements(questionnaireResponse->'item')->'linkId')
      from warehouse.encounter, 
      jsonb_array_elements(details->'QuestionnaireResponse') as questionnaireResponse
      -- where quest->'item'->'linkId' = 'race' -- '{"answer": ["temp"]}'::jsonb
    ) as quest 
    
    order by encounter_id desc 
    limit 20
    ;
   ```
  I am unable to filter QuestionnaireResponse items based on the question ID. e.g., adding (`where to_json(quest)::text = 'race'` returns no results).

kairstenfay:
[Update modeling views to account for new encounter details (from FHIR ETL)](https://trello.com/c/9ZuXsq9D/285-update-modeling-views-to-account-for-new-encounter-details-from-fhir-etl)

kairstenfay:
In a [recent conversation with Trevor](https://seattle-flu-study.slack.com/archives/GJXFQ209K/p1578004312001400), he agrees with me that merging census tract is much more important (for current conditions) than race, ethnicity, vaccine, and symptoms. I marked this PR as ready for review because the census tract work is already complete.

tsibley:
This extraction of census tract looks ok to me, thanks for doing it!

> The problem is that the resulting column from the lateral join is of type `record` and looks like this:
`("[""headaches"", ""chillsOrShivering"", ""soreThroat"", ""runnyOrStuffyNose""]")`. I cannot figure out how to convert it to a regular array.

I'm guessing that you're probably selecting just `symptoms2` which is both the name of the `json_agg` column in the subquery and the table (or technically "relation") name for the subquery results. So `symptom2` is the table name and gets you a record containing one column (your JSON array). You can do `symptom2.symptom2` to get the column, but it'll still be a JSON array of JSON values instead of a Pg array of text values. There a multiple ways to do this, but try something like this:

```sql
select array_agg(id)
from warehouse.encounter,
     jsonb_to_recordset(details->'Condition') as condition(id text)
group by encounter_id;
```

> I am unable to filter QuestionnaireResponse items based on the question ID from within the lateral query. As a result, I also don't know how to return only the question answer from the lateral query instead of the linkId (question name) and answer.

In the example query you were working on, there's repeated use of `jsonb_array_elements()`, which is a red flag to me. That function is a set-returning function and should be used only in a `from` statement or as a simple column in the column list of a `select`. In general when processing JSON in SQL, the best approach is to describe how to map the JSON structure into records using tools like `jsonb_to_recordset` or `jsonb_to_record`. It will then be easier to write the rest of the SQL query. Try something like this as a new starting point:

```sql
select
    encounter_id,
    item."linkId",
    item.answer

from warehouse.encounter,
     jsonb_to_recordset(details->'QuestionnaireResponse') as response(item jsonb),
     jsonb_to_recordset(response.item) as item("linkId" text, answer jsonb)

where item."linkId" in ('vaccine', 'race')

order by encounter_id desc, item."linkId"
limit 20;
```

PR #95, opened 2020-01-10
REDCap and FHIR ETL roles

tsibley:
Necessary to setup automation here.

### DIFF
--- a/dev/add-role
+++ b/dev/add-role
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -euo pipefail
+cd "$(dirname "$0")/.."
+
+main() {
+    role="${1:?The name of the new role is required as the first argument.}"
+    comment="${2:?A comment describing the new role is required as the second argument.}"
+
+    role_name_pattern='^[a-z][a-z-]+[a-z]$'
+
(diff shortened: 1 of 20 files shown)


kairstenfay:
The dev script for creating new roles is awesome.

tsibley:
> We have verified grants for roles in the past, do we still want that?

A good question. I've been generally unhappy with how we verify grants, as it seems fragile and repetitive. I think using pgTAP's [privilege test functions](https://pgtap.org/documentation.html#privilegedaccess) would be an improvement, but I haven't enabled that extension and Just Done It yet. Maybe I should now!?

tsibley:
Following up on my pgTAP thoughts after poking at it/thinking a little more, it may not be the best fit for verification scripts [given the discussion here](https://github.com/sqitchers/sqitch/issues/78) about exit status/exception raising and the larger purpose envisioned of verification scripts. Not sure I agree with the distinction between verification and testing, but note that theory (David Wheeler) is the lead developer of Sqitch, so there's some weight there. I also don't think I'd want to install pgTAP into production (or least into the `public` schema) since it comes with so many functions (ripe for collisions and greater surface area for privilege escalation), so for now I think I want to punt on this.

PR #96, opened 2020-01-14
Remove SFS specific views

joverlee521:
Removes the following SFS specific views from the core shipping views:
- incidence_model_observation_v1
- incidence_model_observation_v2
- incidence_model_observation_v3
- observation_with_presence_absence_result_v1
- observation_with_presence_absence_result_v2

This will need to be deployed with the sqitch `--log-only` option since there are other views within ID3C customizations that depend on these views and we don't actually want to drop these views.


joverlee521:
[Update modeling views to account for new encounter details (from FHIR ETL)](https://trello.com/c/9ZuXsq9D/285-update-modeling-views-to-account-for-new-encounter-details-from-fhir-etl)

PR #99, opened 2020-01-17
FHIR etl updates

joverlee521:
Bumps FHIR etl to REVISION 2 to include updates in response to warning logs in recent etl runs.

* `strip()` barcode values
* skip location references within Encounter resource if the location is a site because it will get processed separately than the other locations. This will reduce the number of "No reference set, cannot resolve" warnings. 
* If a location that is _not_ a site and has no reference, then we also include our own warning. Not entirely sure how to make this warning more meaningful/actionable.

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -52,7 +52,7 @@ LOG = logging.getLogger(__name__)
 # this revision number should be incremented.
 # The etl name has been added to allow multiple etls to process the same
 # receiving table
-REVISION = 1
+REVISION = 2
 ETL_NAME = 'fhir'
 INTERNAL_SYSTEM = 'https://seattleflu.org'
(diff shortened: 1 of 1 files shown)


joverlee521:
[Resolve warnings during REDCap DET and FHIR ETLs on real data](https://trello.com/c/sBpnqH8H/313-resolve-warnings-during-redcap-det-and-fhir-etls-on-real-data)

joverlee521:
> So maybe something like the `receiving.fhir.fhir_id` and the `location_reference.location` structure (as JSON) at the end of the log message?

Hmm looks like `receiving.fhir.fhir_id` would have to be passed down several levels for this log message...

tsibley:
> Hmm looks like `receiving.fhir.fhir_id` would have to be passed down several levels for this log message...

Bah. More and more I'm wishing we had structured logging for preserving this sort of contextual information. But until things calm down enough to do that, just the `location_reference.location` should suffice.

PR #100, opened 2020-01-17
Add function `date_or_null`

joverlee521:
This function casts a string to a date or returns null if the string is in an invalid datetime format.

~This function does not protect against dates that are out of range! Do we want it to?~
This function also returns null if a date is out of range i.e. `2020/01/50`

### DIFF
--- a/schema/verify/functions/date_or_null.sql
+++ b/schema/verify/functions/date_or_null.sql
@@ -1,0 +1,40 @@
+-- Verify seattleflu/schema:functions/date_or_null on pg
+
+begin;
+
+create temporary table tests (
+    input text,
+    expected date
+);
+
+copy tests from stdin with (delimiter ';');
+2020-01-17;2020-01-17
+2020/01/17;2020-01-17
+2020.01.17;2020-01-17
+2020-01-17 00:00:00;2020-01-17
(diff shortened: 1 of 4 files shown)


joverlee521:
[valid date function](https://trello.com/c/HmlXxzLY/319-valid-date-function)

tsibley:
> This function does not protect against dates that are out of range! Do we want it to?

Nice catch! I think we do probably want it to protect against all bad input, so returning null given something like `2020-01-50`.

schema/verify/functions/date_or_null.sql
```diff
@@ -0,0 +1,47 @@
...
+
+        begin
+            perform date_or_null('2020-01-50');
+            raise exception 'date_or_null() did not raise an error on out of range date';
+        exception
+            when datetime_field_overflow then
+                null; -- expected!
+        end;
+    end
+$$;
```
tsibley: Great to see these tests! I don't know about you, but I find it really satisfying to write tests of simple functions like this. :nerd_face:

joverlee521: Yes. _Very_ satisfying 🤓

PR #102, opened 2020-01-23
geocode: Lower log level of invalid address

kairstenfay:
Lower the log level of the invalid address warning when a lookup returns
no response from SmartyStreets. As a result, we will receive fewer,
cluttering warnings in our #id3c-alerts channel.

### DIFF
--- a/lib/id3c/cli/command/geocode.py
+++ b/lib/id3c/cli/command/geocode.py
@@ -335,7 +335,7 @@ def geocode_address(address: dict) -> dict:
     result = lookup.result
 
     if not result:
-        LOG.warning(f"Invalid address: no response from SmartyStreets.")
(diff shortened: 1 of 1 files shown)


kairstenfay:
> +1 for quieting some things down.
> 
> Relatedly, this has me thinking about one additional thing I'd like to start doing in the future: automatically keeping a tally of the kinds of log messages each ETL job produces and warn/alert if some of those tallies exceed "reasonable" or "acceptable" thresholds. This sort of thing is made possible by structured logging, and I've found it really helpful in past ETL frameworks I've written. As a concrete example, we could alert on, say, ETL jobs where "greater than X% of records had an invalid address" and help catch ETL bugs.

I have never thought of that kind of structured logging before. I think that would work very well in ID3C.

PR #103, opened 2020-01-23
location: add `hierarchy-by-feature` option

joverlee521:
This makes hierarchy dynamic per feature without modifying the GeoJSON.

`--hierarchy-by-feature` option takes a CSV file that must have the column "feature_identifier". All other column headers are assumed to be hierarchy keys.

This option overrides `--hierarchy-from` and can override `--hierarchy` if the they have the same keys.

My questions are:
1. should we have a hard coded list of supported hierarchy keys instead of just ingesting all columns within the CSV file?
1. how should we deal with duplicates of a feature_identifier? (it currently just takes the first record)

### DIFF
--- a/lib/id3c/cli/command/location.py
+++ b/lib/id3c/cli/command/location.py
@@ -82,6 +82,12 @@ def location():
     metavar = "<hierarchy>",
     help = "Comma-separated string of key=>value pairs describing where these locations fit within a hierarchy; if this is provided, it overrides --hierarchy-from")
 
+@click.option("--hierarchy-by-feature",
+    metavar = "<hierarchy.csv>",
+    type = click.File("r"),
+    help = "CSV file of hierachy to include for features that must include `feature_identifier` column. All other column headers are assumed to be hierarchy keys; " +
+           "if provided, it overrides --hierarchy-from and can override --hierarchy if same key provided")
+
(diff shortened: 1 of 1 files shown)


joverlee521:
[Switch to exporting neighborhood and PUMA instead of census tracts](https://trello.com/c/F4RMJCzJ/262-switch-to-exporting-neighborhood-and-puma-instead-of-census-tracts)

kairstenfay:
Hi @joverlee521 could you please provide an example of a geojson CSV that would be used in this import process?

joverlee521:
> Hi @joverlee521 could you please provide an example of a CSV that would be used in this import process?

The CSV would be generated if you run `snakemake hierarchy` in [id3c-customization's](https://github.com/seattleflu/id3c-customizations/pull/370) location directory

tsibley:
> 1. should we have a hard coded list of supported hierarchy keys instead of just ingesting all columns within the CSV file?

I intended the location hierarchy to be freeform for whatever local convention is desired, so I don't think ID3C should impose restrictions on it.

And I think reasonable for the interface to automatically use all columns in the CSV instead of requiring another argument specifying which columns to use.

> 2. how should we deal with duplicates of a feature_identifier? (it currently just takes the first record)

I think probably not allow it, e.g. throw an assertion error.

PR #104, opened 2020-01-24
Logging config

tsibley:
A slew of changes to logging to make it more configurable, plus a fix for [the urllib3 problem](https://trello.com/c/mB4AuRjY/330-fix-logging-config-so-that-geocoded-addresses-arent-debug-logged-by-urllib3).


tsibley:
> I'd love to go over this in person with you at some point.

Happy to do so!

PR #106, opened 2020-01-30
logging: Add a filter to suppress log records by their attributes 

tsibley:
This will let us precisely target certain messages to suppress in
production, removing noise from our ETL processes based on
study-specific values.

---

Used by seattleflu/backoffice#11.


tsibley:
Yes, would be happy to chat about logging. :-)

PR #108, opened 2020-02-05
manifest: Genericize extraction of non-essential columns

tsibley:
Avoids needing to add a new, study-specific option every time we need a
new column.  This was my intent from the start, but so far it's been
slightly more pragmatic to just add new options.

Requires config file changes in our specimen-manifests repo, which I've
prepared on a corresponding branch (seattleflu/specimen-manifests#10).

---

This is part of my prep for pushing up new retrospective samples to the UW retro REDCap project.

### DIFF
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -25,7 +25,7 @@ from deepdiff import DeepHash
 from hashlib import sha1
 from os import chdir
 from os.path import dirname
-from typing import Iterable, List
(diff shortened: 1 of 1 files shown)


tsibley:
> Quick question: what do `dst` and `src` mean here? My guess is `src` = "source".

`dst` is destination, `src` is source.

These are pretty ingrained abbreviations for me that are somewhat common (at least in C), but it's good to remember that these abbreviations aren't necessarily widely known!

PR #110, opened 2020-02-06
geocode: extract address from text

joverlee521:
If no address is found from first attempt at geocoding, try to extract address from text using SmartyStreet US Extract API.

Realized we need this feature for the clinical data pull REDCap project, since addresses are in a single text field.

### DIFF
--- a/lib/id3c/cli/command/geocode.py
+++ b/lib/id3c/cli/command/geocode.py
@@ -31,6 +31,7 @@ from cachetools import TTLCache
 from typing import Optional, Tuple, Dict, Any
 from smartystreets_python_sdk import StaticCredentials, ClientBuilder
 from smartystreets_python_sdk.us_street import Lookup
+from smartystreets_python_sdk.us_extract import Lookup as ExtractLookup
 from id3c.cli import cli
 from id3c.cli.command import (
     load_file_as_dataframe
@@ -44,6 +45,7 @@ CACHE_SIZE = float("inf")       # Unlimited
 
 # Shared SmartyStreets client, initialized when first needed.
 STREET_CLIENT = None
+EXTRACT_CLIENT = None
 
 
 @cli.group("geocode", help = __doc__)
@@ -200,10 +202,10 @@ def get_geocoded_addresses(*,
     with pickled_cache(cache_file) as cache:
         address_column_map = {
             'street': street_column,
+            'secondary': secondary_column,
             'city': city_column,
             'state': state_column,
-            'zipcode': zipcode_column,
-            'secondary': secondary_column
+            'zipcode': zipcode_column
         }
 
         addresses_df = load_file_as_dataframe(filename)
@@ -335,6 +337,10 @@ def geocode_address(address: dict) -> dict:
     result = lookup.result
 
     if not result:
+        LOG.info("Previous lookup failed. Looking up address as free text")
+        result = extract_address(address)
+
+    if not result:
         LOG.info(f"Invalid address: no response from SmartyStreets.")
 
     return parse_first_smartystreets_result(result)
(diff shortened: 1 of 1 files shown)


lib/id3c/cli/command/geocode.py
```diff
@@ -304,7 +305,7 @@ def get_response_from_cache_or_geocoding(address: dict,
         response = cache[key]
         LOG.debug('Response found in cache.')
     except KeyError:
-        response = cache[key] = geocode_address(address)
+        response = cache[key] = geocode_address(address) or extract_address(address)
```
tsibley: Extract address also does geocoding, so it's weird to me that it's the fallback to something called `geocode_address`. I think design-wise it'd be nicer to incorporate this fallback into `geocode_address` itself or something?

lib/id3c/cli/command/geocode.py
```diff
@@ -411,6 +412,34 @@ def parse_first_smartystreets_result(result: list) -> dict:
...
+
+    result = client.send(lookup)
+    addresses = result.addresses
+
+    for add in addresses:
+        if len(add.candidates) > 0:
+            result = add.candidates
+            return parse_first_smartystreets_result(result)
+
+    return {}
```
tsibley: The other geocoding function returns `None` when it doesn't succeed instead of an empty dict. Would be nice to be consistent.

lib/id3c/cli/command/geocode.py
```diff
@@ -411,6 +412,34 @@ def parse_first_smartystreets_result(result: list) -> dict:
...
+    Given an *address*, converts it to text and returns a result from
+    the SmartyStreets US Extract API containing information about an address
+    connected to the text.
+
+    Note that this API is not consistent with the US Street API, and the lookup
+    and responses must be handled differently.
+    """
+    LOG.debug("Previous lookup failed. Looking up address as free text.")
+
+    client = smartystreets_client_builder().build_us_extract_api_client()
```
tsibley: For large batches of looksup, it'll be quick if we use a singleton copy of the client instead of building it for each request. This is what `geocode_address()` does, for example. The global `CLIENT` in that function will need to be renamed to be more specific.

lib/id3c/cli/command/geocode.py
```diff
@@ -411,6 +412,34 @@ def parse_first_smartystreets_result(result: list) -> dict:
...
+def extract_address(address: dict) -> dict:
+    """
+    Given an *address*, converts it to text and returns a result from
+    the SmartyStreets US Extract API containing information about an address
+    connected to the text.
+
+    Note that this API is not consistent with the US Street API, and the lookup
+    and responses must be handled differently.
+    """
+    LOG.debug("Previous lookup failed. Looking up address as free text.")
```
tsibley: It's cleaner to write functions that don't care about where they're called from or why, so logging like this is generally better suited at the callsite instead of inside the called function.

lib/id3c/cli/command/geocode.py
```diff
@@ -411,6 +412,34 @@ def parse_first_smartystreets_result(result: list) -> dict:
...
+
+    client = smartystreets_client_builder().build_us_extract_api_client()
+    address_text = ', '.join([str(val) for val in list(address.values()) if val])
```
tsibley: This implicitly relies on the insertion order of keys in the address dictionary. That's ok, I think, since the CPython implementation did this since 3.6 (which we're using) and the Python language spec guarantees it since 3.7.

However, the insertion order comes from `address_column_map`, which puts `secondary` as the last key. We probably don't use this key, but if we did, I believe for address extraction to work it would have to come after `street`:

https://github.com/seattleflu/id3c/blob/36de719553321eb9be89a4745d3ed31eab7e9525/lib/id3c/cli/command/geocode.py#L202-L208

lib/id3c/cli/command/geocode.py
joverlee521: Yup! Done in rework

lib/id3c/cli/command/geocode.py
joverlee521: Ah, yeah. I made sure they were in the right order within redcap-det etls but forgot to check within the geocode command itself 🤦‍♀

PR #111, opened 2020-02-07
manifest parse: Support reading from remote manifest files

tsibley:
This is particularly handy with sensitive manifest data kept on S3 and avoids boilerplate file download/delete juggling around the parse step.

A useful future improvement might be to wrap up fsspec into something like id3c.cli.types.File, which would provide a click.File-like argument "type", complete with proper closing behaviour.  It seemed premature to do so now, but may become more timely if we start adding URL-support to other commands.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -11,7 +11,6 @@ sqlparse = "*"
 
 [packages]
 id3c = {editable = true,path = "."}
-fhir-resources = "*"
 
 [requires]
 python_version = "3.6"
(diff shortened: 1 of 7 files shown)


tsibley:
Worth noting that I [filed an issue](https://github.com/dask/s3fs/issues/282) with s3fs about its rogue logging messages, which will appear in id3c's output if given an S3 URL.

tsibley:
Added one more commit to bring remote URL support to the `manifest diff` and `manifest upload` commands, which I realized I'd want for the UW retros too: https://github.com/seattleflu/id3c/pull/111/commits/f009ad9506cfef52ba85f9c22a8668532dc40e92

PR #112, opened 2020-02-07
etl redcap-det: add --include-incomplete flag

joverlee521:
This allows the etl to pull redcap-records even if the instrument in the DET is marked incomplete.

This is necessary for pulling REDCap records from the Clinical Data Pulls project since the instruments will not be marked "complete" by ITHS.


joverlee521:
> Looks good. My only wonder is if this would be more suitable as an internal parameter to `@command_for_project` instead of a CLI option. It seems like a project would always want one behaviour or the other, but not necessarily need the ability to switch between the two? Thoughts?

Hmmm that's true, I can't think of why a project would want to switch. I'll change it to an internal parameter.

PR #118, opened 2020-02-14
etl manifest: prevent ingestion of samples without collection barcode

joverlee521:
The only samples that are expected not to have collection barcodes are UW retrospectives. 
This prevents the ingestion of samples without collection barcodes if they are not a UW retrospective sample.

### DIFF
--- a/lib/id3c/cli/command/etl/manifest.py
+++ b/lib/id3c/cli/command/etl/manifest.py
@@ -15,6 +15,7 @@ or updated records.
 """
 import click
 import logging
+import re
 from datetime import datetime, timezone
 from typing import Any, Tuple, Optional
 from id3c.cli.command import with_database_session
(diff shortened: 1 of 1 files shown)


lib/id3c/cli/command/etl/manifest.py
```diff
@@ -111,6 +112,14 @@ def etl_manifest(*, db: DatabaseSession):
             collection_barcode = manifest_record.document.pop("collection", None)
             collection_identifier = find_identifier(db, collection_barcode) if collection_barcode else None
 
+            sample_origin = manifest_record.document["sample_origin"]
+            # Only UW retrospectives do not have collection barcodes, prevent
+            # ingestion of other samples that do not have collection barcode yet
+            if not collection_barcode and not re.match("(uwmc|nwh|hmc)_retro", sample_origin):
+                LOG.info(f"Skipping sample that is not a UW retrospective sample that does not have a collection barcode.")
+                mark_skipped(db, manifest_record.id)
+                continue
```
tsibley: My main concern here is that it's a new addition of very SFS-specific logic to core ID3C, and that's something which we've been trying to make decrease with time.

I wonder about doing this filtering in the `specimen-manifests` repo instead, before `manifest upload`? That feels like the right place and NDJSON streams make such filtering ~easy (which is why we're using NDJSON). Granted, that wouldn't address the existing records in `receiving.manifest`, but presumably those won't be an issue anymore since the correct/full `warehouse.sample` records exist now.

joverlee521: Ah, that is true! Can definitely move the filtering up

joverlee521:
Seems too SFS specific, moving filtering to `specimen-manifests` repo instead

PR #119, opened 2020-02-25
Add presence absence device

joverlee521:
Add `device` to `presence_absence.details` to help differentiate presence/absence results. 

In `etl presence-absence` the device is hardcoded to `OpenArray` since that it only ingests NWGC results.

`etl fhir` uses the Observation Device Identifier value. 

Also created a new view `shipping.presence_absence_results_v2` to include a device column.

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -51,7 +51,7 @@ LOG = logging.getLogger(__name__)
 # this revision number should be incremented.
 # The etl name has been added to allow multiple etls to process the same
 # receiving table
-REVISION = 2
+REVISION = 3
 ETL_NAME = 'fhir'
 INTERNAL_SYSTEM = 'https://seattleflu.org'
 LOCATION_RELATION_SYSTEM = 'http://terminology.hl7.org/CodeSystem/v3-RoleCode'
@@ -822,12 +822,14 @@ def process_presence_absence_tests(db: DatabaseSession, report: DiagnosticReport
         if result_value is None:
             continue
 
+        details = { "device": observation.device.identifier.value }
+
         upsert_presence_absence(db,
             identifier = f'{barcode}/{snomed_code}/{observation.device.identifier.value}',
             sample_id = sample_id,
             target_id = target.id,
             present = result_value,
-            details = {})
+            details = details)
 
 
 def mark_skipped(db, fhir_id: int) -> None:
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -41,7 +41,7 @@ LOG = logging.getLogger(__name__)
 # presence-absence tests lacking this revision number in their log.  If a
 # change to the ETL routine necessitates re-processing all presence-absence tests,
 # this revision number should be incremented.
-REVISION = 5
+REVISION = 6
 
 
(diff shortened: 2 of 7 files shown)


schema/deploy/shipping/views.sql
```diff
@@ -38,4 +38,30 @@ grant select
...
+    'View of warehoused presence-absence results for modeling and viz teams';
+
+revoke all
+    on shipping.presence_absence_result_v2
+  from "incidence-modeler";
+
+grant select
+    on shipping.presence_absence_result_v2
+    to "incidence-modeler";
+
```
kairstenfay: If the only change we're making to this view is adding a new column, I don't think it's necessary to bump the version number.

joverlee521: The only reason I bumped it was because `sqitch revert` won't work, but if we're fine with that then I can change back to v1

kairstenfay: Ah do you mean `sqitch revert` won't work because we added a new column? I recall running into a similar problem. I _think_ `sqitch revert` should work if you drop the view then re-create it. I'd prefer to keep our versioning system of only bumping the revision number if we're making breaking changes to the view.

joverlee521: I can't drop the view because there are other views dependent on it

kairstenfay:
Tangentially -- now that we're thinking again about which parts of core `id3c` are still SFS-specific: do we want to port over the original `shipping.presence_absence_result_v1` view to `id3c-customizations`?

joverlee521:
> Tangentially -- now that we're thinking again about which parts of core `id3c` are still SFS-specific: do we want to port over the original `shipping.presence_absence_result_v1` view to `id3c-customizations`?

Good question. I think the original `shipping.presence_absence_result_v1` seems pretty generic and not very SFS specific. Although adding `device` might be more SFS specific? (I feel like I might be splitting hairs here...)

kairstenfay:
> > Tangentially -- now that we're thinking again about which parts of core `id3c` are still SFS-specific: do we want to port over the original `shipping.presence_absence_result_v1` view to `id3c-customizations`?
> 
> Good question. I think the original `shipping.presence_absence_result_v1` seems pretty generic and not very SFS specific. Although adding `device` might be more SFS specific? (I feel like I might be splitting hairs here...)

Ah sorry I deleted this comment. I wrote it before I realized that this PR is in the id3c repo, not id3c-customizations. I think this view makes sense to leave in `id3c`.

joverlee521:
After speaking with @tsibley in person, I've changed the view to include `details` instead of just `details ->> device` to keep the view more generic.

joverlee521:
Just wanted to flag another thing I've discovered: `shipping.presence_absence_result_v1` uses `sample.identifier` so there is a delay in returning kiosk Cepheid results while waiting for the lab to assign a sample barcode. 

I talked briefly with Misja about this and she says from the kiosk to the lab there is about a 4 day delay and that it shouldn't be that much of an issue.

schema/deploy/shipping/views.sql
kairstenfay: I'm going to add the COVID-19 results on top of v2, so I'm fine keeping it as is.

kairstenfay:
This looks good @joverlee521. @tsibley said it's OK to merge if we both think it looks good. I plan on building on top of this for COVID-19 stuff soon.

joverlee521:
Merged and deployed

PR #120, opened 2020-02-26
etl fhir: handle literal references to Specimen

joverlee521:
Our internally created Diagnostic Reports have a literal reference to
a Specimen within the same bundle rather than a logical reference.
This updates the `process_diagnostic_report_bundle_entry` to handle both

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -180,7 +180,13 @@ def process_diagnostic_report_bundle_entry(db: DatabaseSession, bundle: Bundle, 
     LOG.debug(f"Processing DiagnosticReport Resource «{entry.fullUrl}».")
 
     for reference in resource.specimen:
-        if not matching_system(reference.identifier, INTERNAL_SYSTEM):
+        if not reference.identifier:
+            specimen = reference.resolved(Specimen)
+            reference.identifier = specimen.identifier[0]
+            if not matching_system(reference.identifier, f"{INTERNAL_SYSTEM}/sample"):
+                continue
+
+        elif not matching_system(reference.identifier, INTERNAL_SYSTEM):
             continue
 
         barcode = reference.identifier.value.strip()


joverlee521:
This worked locally, so merging to stop the alerts

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -180,7 +180,13 @@ def process_diagnostic_report_bundle_entry(db: DatabaseSession, bundle: Bundle,
     LOG.debug(f"Processing DiagnosticReport Resource «{entry.fullUrl}».")
 
     for reference in resource.specimen:
-        if not matching_system(reference.identifier, INTERNAL_SYSTEM):
+        if not reference.identifier:
+            specimen = reference.resolved(Specimen)
+            reference.identifier = specimen.identifier[0]
```
tsibley: It'd be nice to find all the `specimen.identifier` items which match our system, instead of only the first. This would correctly handle Specimens with multiple identifiers where ours isn't first.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -180,7 +180,13 @@ def process_diagnostic_report_bundle_entry(db: DatabaseSession, bundle: Bundle,
...
+            if not matching_system(reference.identifier, f"{INTERNAL_SYSTEM}/sample"):
+                continue
+
+        elif not matching_system(reference.identifier, INTERNAL_SYSTEM):
```
kairstenfay: Why are we checking for two different internal systems endpoints here?

joverlee521: Our internal bundles from redcap-det ETLs give use `https://seattleflu.org/sample` as the system for specimens while the Ellume team's bundles just use `https://seattleflu.org`

This was just a quick fix, but we should probably reconcile the two in the future.

lib/id3c/cli/command/etl/fhir.py
joverlee521: Yes definitely!

PR #121, opened 2020-02-26
Add default sheet_name to load_file_as_dataframe

kairstenfay:
Without a default `sheet_name`, our `read_excel` function defaults to
`None`. Unfortuantely, this causes `pd.read_excel` to return a
`Dict[str, pd.DataFrame]` instead of a DataFrame as we would expect.

We're not specifying a sheet name as an argument in the
`load_file_as_dataframe` function, so hard-code the first sheet (0)
as the `sheet_name` argument.

### DIFF
--- a/lib/id3c/cli/io/pandas.py
+++ b/lib/id3c/cli/io/pandas.py
@@ -25,6 +25,8 @@ def load_file_as_dataframe(filename: str) -> pd.DataFrame:
     Given a *filename*, loads its data as a pandas DataFrame.
     Supported extensions are csv, tsv, xls, and xlsx.
 
+    If given an Excel workbook, defaults to loading only the first sheet.
+
     Raises a :class: `UnsupportedFileExtensionError` if the given *filename*
     ends with an unsupported extension.
     """
@@ -61,7 +63,7 @@ def load_input_from_file_or_stdin(filename: click.File) -> pd.DataFrame:
     return input_df
 
 
-def read_excel(io, sheet_name = None, na_filter: bool = True):
+def read_excel(io, sheet_name = 0, na_filter: bool = True):
     """
     Reads an Excel file while ensuring all values are treated as strings.
 


kairstenfay:
This bug first became apparent to me via https://github.com/seattleflu/id3c-customizations/pull/53

kairstenfay:
[Read xls or csv in `clinical parse-sch`](https://trello.com/c/3uZSm38H/344-read-xls-or-csv-in-clinical-parse-sch)

kairstenfay:
> Would it make sense to set the default value to `0` for our `read_excel` function to mimic pandas and prevent the same problem when we use the `read_excel` function in the future?

Perhaps it would. I'm open to either solution.

kairstenfay:
I'm running into another bug with this function: empty strings for columns are preserved rather than cast to `<NA>`.

tsibley:
> Would it make sense to set the default value to `0` for our `read_excel` function to mimic pandas and prevent the same problem when we use the `read_excel` function in the future?

Yeah, I think this is a better fix. I intended `id3c.cli.io.pandas.read_excel` to be a ~drop-in replacement for `pandas.read_excel`.

> I'm running into another bug with this function: empty strings for columns are preserved rather than cast to `<NA>`.

Hmm. I can see reasons to both a) leave that behaviour and expect callers to `.replace({"": pd.NA})` if they want or b) do the `.replace({"": pd.NA})` automatically and expect callers to replace it back to empty strings if that's what they want. I think choosing (a) vs. (b) depends on what our code typically wants?

tsibley:
If we do want to automatically `.replace({"": pd.NA})`, there's probably less casting weirdness if we don't actually call `.replace` but instead update our converter function from `str` to something like `lambda x: str(x) or None`.

kairstenfay:
> I think choosing (a) vs. (b) depends on what our code typically wants?

I think our code typically wants empty strings to be `<NA>`.

kairstenfay:
I ended up not changing the current dtype casting done in the `read_excel` function, so empty strings remain as empty strings in dataframes returned there. In https://github.com/seattleflu/id3c-customizations/pull/53, I'm using `.replace({"": None})` so that when I use `pd.to_json()`, I don't get empty dicts for every `pd.NA`. 

I suppose here I could have converted empty strings to `pd.NA` then have replaced `pd.NA` with none in https://github.com/seattleflu/id3c-customizations/pull/53. 

Do you have a preference between the two proposed solutions? @tsibley

PR #126, opened 2020-03-04
etl presence-absence: Support ingestion of non-OpenArray results

tsibley:
These results come in a very similar JSON format, but lack a "chip" field which only applies to OpenArray.  In order to avoid confusing logic, the following decision table is used:

```
has results? | chip key exists? | chip is empty? | action  | comment                                                   |
-------------|------------------|----------------|---------|-----------------------------------------------------------|
yes          | yes              | no             | ingest  | normal p/a ingest                                         |
yes          | yes              | yes            | error   | Samplify bug                                              |
yes          | no               | -              | ingest  | non-OpenArray results, currently only in id3c-restricted  |
no           | -                | -              | skip    | skeletal OpenArray results without any actual data        |
```

Note that the table only considers the interactions of chip and results, not other fields like barcode.

---

- [x] Making this draft PR for visibility, but this still needs testing before we merge and deploy. Due to being pulled away to other tasks, I have not yet had time to test this locally with both the problematic data pushed to id3c-production yesterday + the non-OpenArray data in id3c-restricted.

- [ ] Device is still set to OpenArray, even though some of these are now non-OpenArray results. Should set this to null, I guess?

Also note that even with merging, this is still anticipatory, and we can't load the restricted data yet. Working on clarifications there.

### DIFF
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -102,31 +102,42 @@ def etl_presence_absence(*, db: DatabaseSession):
                     LOG.info(f"Skipping sample «{received_sample['sampleId']}» without SFS barcode")
                     continue
 
+                # Don't go any further if there are no results to import.
+                test_results = received_sample["targetResults"]
+
+                if not test_results:
+                    LOG.warning(f"Skipping sample «{received_sample_barcode}» without any results")
+                    continue
+
                 received_sample_id = str(received_sample["sampleId"])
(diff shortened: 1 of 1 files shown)


tsibley:
Tested and works locally.

tsibley:
Just repushed with a change to set device to null when not OpenArray.

joverlee521:
> Device is still set to OpenArray, even though some of these are now non-OpenArray results. Should set this to null, I guess?

Null or another device name? I guess the null will on apply to HCoV-19 results?

PR #129, opened 2020-03-09
etl redcap-det: allow raw coded values for multiple choice

joverlee521:
The default value for `raw_coded_values` is False, so we don't need to update the existing custom redcap-det ETLs. 

This will allow us to pull the coded values for the SCAN redcap project.


joverlee521:
[Update id3c cli redcap to ingest answer ids or answer labels](https://trello.com/c/yRYJD6zZ/377-update-id3c-cli-redcap-to-ingest-answer-ids-or-answer-labels)

PR #130, opened 2020-03-11
etl `upsert_location`: include new location's own hierarchy

joverlee521:
Always include a new location's own `scale => identifier` in
its hierarchy.

On update, new hierarchy and exisiting hierarchy are concatenated, with
new hierarchy taking precedence if keys overlap.

### DIFF
--- a/lib/id3c/cli/command/etl/__init__.py
+++ b/lib/id3c/cli/command/etl/__init__.py
@@ -278,18 +278,28 @@ def upsert_location(db: DatabaseSession,
     """
     Upserts a location by its *scale* and *identifier*.
 
-    If *hierarchy* is None and the location already exists, any existing
-    hierarchy is preserved.
+    If *hierarchy* is None, it will be set to the location's
+    `scale => identifier`. Otherwise, the location's `scale => identifier`
+    will be appended to the *hierarchy*.
+
+    On update, new hierarchy and existing hierarchy are concatenated, with
+    new hierarchy taking precedence if there is overlap of keys.
     """
     LOG.debug(f"Upserting location {(scale, identifier)}")
+
+    # Always includes the new location's own scale => identifier in hierarchy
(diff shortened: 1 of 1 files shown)


joverlee521:
[Include address in hierarchy for newly created addresses](https://trello.com/c/I7MA0Gik/388-include-address-in-hierarchy-for-newly-created-addresses)

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -278,18 +278,28 @@ def upsert_location(db: DatabaseSession,
...
+        hierarchy = hierarchy + "," + location_hierarchy
+
     location = db.fetch_row("""
         insert into warehouse.location (scale, identifier, hierarchy)
         values (%s, %s, %s)
 
         on conflict (scale, identifier) do update
-            set hierarchy = coalesce(excluded.hierarchy, location.hierarchy)
-                    || hstore(lower(location.scale), lower(location.identifier))
+            set hierarchy = location.hierarchy || excluded.hierarchy
```
tsibley: If the current `location.hierarchy` is null, then `location.hierarchy || excluded.hierarchy` will evaluate to null even if `excluded.hierarchy` is defined. I think want something like this instead:

```
coalesce(location.hierarchy, '') || excluded.hierarchy
```

my testing:

```
tom@[local]/seattleflu=# select null::hstore || 'baz=>bat';
 ?column? 
──────────
 
(1 row)

tom@[local]/seattleflu=# select coalesce(null::hstore, '') || 'baz=>bat';
   ?column?   
──────────────
 "baz"=>"bat"
(1 row)

tom@[local]/seattleflu=# select coalesce('foo=>bar'::hstore, '') || 'baz=>bat';
          ?column?          
────────────────────────────
 "baz"=>"bat", "foo"=>"bar"
(1 row)
```

joverlee521: Good catch! I will update this

PR #131, opened 2020-03-13
Add the concept of a "primary" encounter location

tsibley:
Creates a framework for prioritizing encounter-locations by relation and
uses that in a new view that subsets warehouse.encounter_location to
just the "primary" rows.  A default set of relations and priorities is
added, but these can be tailored for each database since they are just
rows in a controlled vocabulary table.

This provides support to queries which want to only follow a single link
across the one-to-many relationship without hardcoding a specific
relation (e.g. to choose the "best" location available when the data is
sparse).

### DIFF
--- a/schema/deploy/warehouse/encounter-location-relation.sql
+++ b/schema/deploy/warehouse/encounter-location-relation.sql
@@ -1,0 +1,20 @@
+-- Deploy seattleflu/schema:warehouse/encounter-location-relation to pg
+-- requires: warehouse/schema
+
+begin;
+
+create table warehouse.encounter_location_relation (
+    relation citext primary key,
+    priority integer unique
+);
+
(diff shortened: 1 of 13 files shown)


kairstenfay:
> Just a thought, do we want to use or include FHIR Location type codes in `warehouse.encounter_location_relation`?

Interesting. @joverlee521  do you mean in a column in addition to our own labels (`residence`, `lodging`, etc.)

tsibley:
Yeah, I think it'd be smart to include mappings to FHIR in the database and then those within the FHIR ETL instead of [our hardcoded map](https://github.com/seattleflu/id3c/blob/da8fb7a189eb7c8806a5af5d5e72b017b9ec5e8a/lib/id3c/cli/command/etl/fhir.py#L384-L390). But I don't want to tackle that now.

PR #135, opened 2020-03-21
etl presence-absence: Handle new "sampleState" field

tsibley:
This field is being used for the results of CLIA-certified COVID-19
targets.  Other research-use-only targets continue to use
"targetResults".

Indeterminate/inconclusive results will be ingested with the "present"
column set to null, under the rationale that the presence known to be
uncertain/unknown.  These kind of results are typically because some
compound targets or target replicates are positive but some are
negative.  While the schema allows for "present is null", this is the
first time we're ingesting such data.

Several workflow states which aren't actual results are skipped during
ingestion.

Only the values "Positive", "Negative", and "Review" have been observed
so far, but the other possible values were communicated to us by NWGC.
It remains to be seen if those are accurate yet.

### DIFF
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -142,6 +142,13 @@ def etl_presence_absence(*, db: DatabaseSession):
                     LOG.debug(f"Processing target «{test_result_target_id}» for \
                     sample «{received_sample_barcode}»")
 
+                    # Skip this result if it's actually a non-result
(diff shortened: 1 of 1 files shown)


tsibley:
It's handy when you need to ad-hoc extend three-value logic (true/false/null) to four values, and I think fine for limited use like that. The a bit more proper way, especially if we were using this outside a small chunk of code, would be to make an [enum](https://docs.python.org/3/library/enum.html) instead and return enum members.

PR #136, opened 2020-03-24
etl fhir: update `assert_required_resource_types_present`

joverlee521:
Now allows Bundles without Specimen Resources. If a Patient Resource
is present, make sure that there is at least one Encounter Resource.

Only check the number of Observation Resources if a Specimen Resource
is present because we are using these Observation Resources to link
Patient, Encounter, and Specimen.


joverlee521:
[Ingest data from SCAN screening questionnaire before smaple received by lab (e.g. early REDCap metadata)](https://trello.com/c/VbQXhRCV/430-ingest-data-from-scan-screening-questionnaire-before-smaple-received-by-lab-eg-early-redcap-metadata)

PR #139, opened 2020-03-31
etl fhir: Ingest Specimen.note

kairstenfay:
In the SCAN REDCap-DET ETL, we recently started adding a specimen note
if a sample is unable to be tested ("never-tested").

Ingest this attribute, placing it in sample.details under the key,
"note".

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -51,7 +51,7 @@ LOG = logging.getLogger(__name__)
 # this revision number should be incremented.
 # The etl name has been added to allow multiple etls to process the same
 # receiving table
-REVISION = 3
+REVISION = 4
 ETL_NAME = 'fhir'
 INTERNAL_SYSTEM = 'https://seattleflu.org'
 LOCATION_RELATION_SYSTEM = 'http://terminology.hl7.org/CodeSystem/v3-RoleCode'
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Update FHIR etl to ingest `Specimen.note` into `sample.details`](https://trello.com/c/ak15T4rJ/475-update-fhir-etl-to-ingest-specimennote-into-sampledetails)

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -553,13 +553,19 @@ def related_specimens(encounter: Encounter, resources: Dict[str, List[DomainReso
         else:
             assert False, "logic bug"
 
+        sample_details = {}
+        if specimen.note:
+            sample_details['note'] = specimen.note[0].text
+
+        additional_details = { **specimen.type.as_json(), **sample_details}
```
tsibley: This would leave us with both the `note` and `notes` key; the latter comes from the manifest.

I'd like to avoid this if we can…

kairstenfay: I forgot about the `notes` key. 

Should it be named `status_code`?

joverlee521: Stick with the `able_to_test` key?

tsibley: Hmm. I'd also like to avoid encoding more study-specific handling of the data into this FHIR ingest, which is supposed to be trending towards generic.

I do think the current handling of note in this PR is good and fine. It's only the interaction with our manifest ETL that's poor. Perhaps we should ingest manifest notes as a different key altogether and do a bulk-update, for example?

kairstenfay: I added a [Trello card](https://trello.com/c/seCBMUxV/485-store-specimen-manifest-details-under-manifest-key-in-sampledetails) to move the specimen manifest details under a new key in `sample.details`. 

OK to keep the `Specimen.note` annotation under `sample.details->>'note'` (as I've done here)?

tsibley: Yes!

PR #140, opened 2020-03-31
New samplify fields

joverlee521:
Ingest new fields `assayName` and `assayDate`. Also added a bonus `extractionDate` field since I've seen that get mentioned on Slack as well.

### DIFF
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -41,7 +41,7 @@ LOG = logging.getLogger(__name__)
 # presence-absence tests lacking this revision number in their log.  If a
 # change to the ETL routine necessitates re-processing all presence-absence tests,
 # this revision number should be incremented.
-REVISION = 6
+REVISION = 7
 
 
 @etl.command("presence-absence", help = __doc__)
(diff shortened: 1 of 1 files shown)


joverlee521:
[Ingest assay date and assay name from presence/absence results](https://trello.com/c/tOld8PXj/462-ingest-assay-date-and-assay-name-from-presence-absence-results)

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -41,7 +41,7 @@
 # presence-absence tests lacking this revision number in their log.  If a
 # change to the ETL routine necessitates re-processing all presence-absence tests,
 # this revision number should be incremented.
-REVISION = 6
+REVISION = 7
```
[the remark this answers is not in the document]
joverlee521: We already have results in `receiving.presence_absence` with these fields, so bump is necessary to re-process those...

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -285,13 +290,27 @@ def sample_details(document: dict) -> dict:
...
+    if assay_name:
+        assert assay_name in {'OpenArray', 'TaqmanQPCR'}, f"Found unknown assay name «{assay_name}»"
+        device = assay_name
+    elif chip:
+        device = "OpenArray"
+
     return {
-        "device": "OpenArray" if chip else None,
+        "device": device,
+        "assay_date": assay_date,
```
kairstenfay, line 316: Do we want `null` assay_date for p/a records where that isn't available?

joverlee521: That's what we've done with `device` so I followed the same pattern. 
My thought is this way we know we've tried to ingest `assay_date` but the value is `null` because it was not provided to us.

PR #141, opened 2020-04-03
etl presence-absence: Ingest `assayType` for p/a results

joverlee521:
Checks provided `assayType` is either "Clia" or "Research".

If no `assayType` is provided, then count the number of `wellResults`
to infer `assay_type`. 4 wells indicates "Clia" and everything else is
"Research". (added this assuming NWGC will not be resending everything)


joverlee521:
[Ingest p/a assayType field into pa.details](https://trello.com/c/J5V8Xlbh/500-ingest-p-a-assaytype-field-into-padetails)

PR #145, opened 2020-04-16
labelmaker: update SamplesLayout to print duplicates

joverlee521:
Sample barcodes are only needed in duplicate since the lab is no longer
doing the C aliquot.

This updates the labels to print duplicate barcodes, with 3 pairs
per row and the last column blank.


joverlee521:
[200 sheets of sample barcodes in duplicate](https://trello.com/c/YqP70qGI/527-200-sheets-of-sample-barcodes-in-duplicate)

PR #146, opened 2020-04-21
Sample collection date

joverlee521:
Add `collection_date` to `warehouse.sample` and ingest this field from the manifest ETL.

I've added a `validate_date` function to clean up the manually entered collection date column from the aliquoting manifest. 
1. Do we want to use this to check that the date is not too old or a future date as well? (We currently have a sample from 1911-04-04 😆 )
2. Do we want `sample.details.date` to hold the raw value from the aliquoting sheet or should we update that to be the cleaned up date value as well?

### DIFF
--- a/lib/id3c/cli/command/etl/manifest.py
+++ b/lib/id3c/cli/command/etl/manifest.py
@@ -128,6 +128,9 @@ def etl_manifest(*, db: DatabaseSession):
                 or collection_identifier.set_name in expected_identifier_sets["collections"], \
                     f"Collection identifier found in set «{collection_identifier.set_name}», not {expected_identifier_sets['collections']}" # type: ignore
 
+            # Sample collection date
+            collection_date = manifest_record.document.get("date")
+
             # Upsert sample cooperatively with enrollments ETL routine
             #
             # The details document was intentionally modified by two pop()s
@@ -137,6 +140,7 @@ def etl_manifest(*, db: DatabaseSession):
             sample, status = upsert_sample(db,
                 identifier            = sample_identifier.uuid,
                 collection_identifier = collection_identifier.uuid if collection_identifier else None,
+                collection_date       = collection_date,
                 additional_details    = manifest_record.document)
 
             mark_loaded(db, manifest_record.id,
@@ -149,19 +153,21 @@ def etl_manifest(*, db: DatabaseSession):
 def upsert_sample(db: DatabaseSession,
                   identifier: str,
                   collection_identifier: Optional[str],
+                  collection_date: Optional[str],
                   additional_details: dict) -> Tuple[Any, str]:
     """
     Upsert sample by its *identifier* and/or *collection_identifier*.
 
-    An existing sample has its *identifier* and *collection_identifier* updated,
-    and the provided *additional_details* are merged (at the top-level only)
-    into the existing sample details, if any.
+    An existing sample has its *identifier*, *collection_identifier*,
+    *collection_date* updated, and the provided *additional_details* are
+    merged (at the top-level only) into the existing sample details, if any.
 
     Raises an exception if there is more than one matching sample.
     """
     data = {
         "identifier": identifier,
         "collection_identifier": collection_identifier,
+        "collection_date": collection_date,
         "additional_details": Json(additional_details),
     }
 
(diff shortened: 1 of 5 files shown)


joverlee521:
[Add collection_date column to warehouse.sample](https://trello.com/c/VzEDLk0i/540-add-collectiondate-column-to-warehousesample)

kairstenfay:
It looks like the `sqitch.plan` for id3c was missing a tag after the latest addition (before your work). Could you add a tag there so that this change can be properly reverted?

joverlee521:
> I think we want to bump the revision number for the manifest ETL.

I didn't bump the revision number because we were going to back-fill previous records with result of `date_or_null` on `details.date`. 

When I did test locally by bumping the revision number, I ran into a lot of breaking errors (wrong collection barcodes that were skipped manually at one point or the two samples found error for bad records in `receiving.manifest`).

kairstenfay:
> > I think we want to bump the revision number for the manifest ETL.
> 
> I didn't bump the revision number because we were going to back-fill previous records with result of `date_or_null` on `details.date`.
> 
> When I did test locally by bumping the revision number, I ran into a lot of breaking errors (wrong collection barcodes that were skipped manually at one point or the two samples found error for bad records in `receiving.manifest`).

Ah yes the manual skips... How did you get it to run locally without bumping the revision number, though?

kairstenfay:
> > I think we want to bump the revision number for the manifest ETL.
> 
> I didn't bump the revision number because we were going to back-fill previous records with result of `date_or_null` on `details.date`.
> 
> When I did test locally by bumping the revision number, I ran into a lot of breaking errors (wrong collection barcodes that were skipped manually at one point or the two samples found error for bad records in `receiving.manifest`).

Ah yes the manually skipped barcodes. 
What's the plan for back-filling previous records?

joverlee521:
> Ah yes the manually skipped barcodes.
> What's the plan for back-filling previous records?

Running the following should do the trick:
```
update warehouse.sample set collection_date = date_or_null(details ->> 'date') 
```

kairstenfay:
> > Ah yes the manually skipped barcodes.
> > What's the plan for back-filling previous records?
> 
> Running the following should do the trick:
> 
> ```
> update warehouse.sample set collection_date = date_or_null(details ->> 'date') 
> ```

Cool, thanks for humoring me. 

I still think we should add the missing sqitch tag before your most recent work. Otherwise looks good to merge.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -126,6 +126,13 @@ def etl_manifest(*, db: DatabaseSession):
                 or collection_identifier.set_name in expected_identifier_sets["collections"], \
                     f"Collection identifier found in set «{collection_identifier.set_name}», not {expected_identifier_sets['collections']}" # type: ignore
 
+            # Validate sample collection date if available
+            collection_date = manifest_record.document.get("date")
+            if collection_date:
+                # These are the current date formats found in `receiving.manifest`
+                expected_date_formats = ("%m/%d/%Y", "%m/%d/%y", "%Y-%m-%d %H:%M:%S")
+                collection_date = validate_date(collection_date, expected_date_formats)
```
tsibley, line 134: I'm not sure it's sustainable to try to validate all datetime formats we might see. We'll have to update the code whenever a new format is used. Hardcoding formats also doesn't align with the broader vision of ID3C as not just for SFS. I'd be tempted to rely on Pg's extensive native date parsing instead. For example, the statements in `upsert_sample()` can use `date_or_null(%(collection_date)s)` when setting `collection_date`.

One downside would be losing the warnings about bogus dates, but I think that might actually be an upside in terms of ETL job noise. I don't think we want to be in the business of policing all those dates as they roll in. 🌊 We would still be able to do bulk QC by writing a query to find samples without a `collection_date` and compare it to the original manifest record's (unparsed) `date` (either from `sample.details` or `receiving.manifest`).

joverlee521: I am 100% onboard with less ETL job noise. 😄 

I'll update `upsert_sample` to use our custom `date_or_null` function to set the collection date value.

kairstenfay:
I realize now I didn't answer your questions from your initial post.

> * Do we want to use this to check that the date is not too old or a future date as well? (We currently have a sample from 1911-04-04 laughing )

No, I think we shouldn't assume anything about dates to make ID3C as portable as possible.

> * Do we want `sample.details.date` to hold the raw value from the aliquoting sheet or should we update that to be the cleaned up date value as well?

Yes I think it should continue to hold the raw date. It could be useful to compare with if a `collected` date is null (e.g. because it's invalid) and we could easily point to the bad `sample.details.date`.

PR #148, opened 2020-04-30
REDCap fields and version

tsibley:
See commits. I added these when I was prototyping a REDCap project "switchboard" which constructs direct links to REDCap records. These changes aren't directly necessary by anything yet, but seemed generally useful to add to this growing REDCap library (which maybe should be spun out into a separate package sooner than later?).

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -28,6 +28,7 @@ class Project:
     _details: dict
     _instruments: List[str] = None
     _fields: List[dict] = None
+    _redcap_version: str = None
 
     def __init__(self, api_url: str, api_token: str, project_id: int) -> None:
         self.api_url = api_url
@@ -89,6 +90,17 @@ class Project:
         the first field in a project.
         """
         return self.fields[0]["field_name"]
+
+
+    @property
+    def redcap_version(self) -> str:
+        """
+        Version string of the REDCap instance.
+        """
+        if not self._redcap_version:
+            self._redcap_version = self._fetch("version", format = "text")
+
+        return self._redcap_version
 
 
     def record(self, record_id: str, *, raw: bool = False) -> List['Record']:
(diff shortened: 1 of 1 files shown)


lib/id3c/cli/redcap.py
```diff
@@ -173,20 +185,20 @@ def _fetch(self, content: str, parameters: Dict[str, str] = {}) -> Any:
 
         headers = {
             'Content-type': 'application/x-www-form-urlencoded',
-            'Accept': 'application/json'
+            'Accept': 'application/json, */*;q=0.1'
         }
```
kairstenfay, line 189: What does `*/*;q=0.1` do here?

tsibley: https://developer.mozilla.org/en-US/docs/Glossary/Quality_values

tsibley: Worth noting that this Accept header is really more descriptive than functional, as I don't think REDCap actually cares about it (unlike other APIs and servers which do). I am realizing that I should probably change it to:

     Accept: application/json, text/*

which more accurately reflects the way the response is handled. (Or just remove it entirely.)

tsibley: Repushed adjusting the Accept header:

```
tom@whunk id3c (redcap-fields-and-version<>) $ git diff-rebase u
1:  39a12c1 = 1:  39a12c1 redcap: Support limiting the record fields fetched
2:  529a09a ! 2:  aa402b7 redcap: Support fetching the REDCap version
    @@ lib/id3c/cli/redcap.py: class Project:
              headers = {
                  'Content-type': 'application/x-www-form-urlencoded',
     -            'Accept': 'application/json'
    -+            'Accept': 'application/json, */*;q=0.1'
    ++            'Accept': 'application/json' if format == "json" else 'text/*'
              }
      
              data = {

```

I think this is clearer and more accurate, so thanks for making me think about it again. :-)

PR #149, opened 2020-04-30
Scan follow up fhir

joverlee521:
Some updates to the FHIR ETL to allow ingestion of the SCAN follow-up encounter:
* Allow FHIR bundle to have multiple Encounter resources even if it only has one Specimen resource
* Ingest `Encounter.reasonCode` into `encounter.details.reason` to allow for differentiation of follow-up encounters from primary encounters.

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -10,6 +10,7 @@ from datetime import datetime, timezone
 from typing import Any, List, Dict, Optional, Tuple
 from fhir.resources.bundle import Bundle, BundleEntry
 from fhir.resources.codeableconcept import CodeableConcept
+from fhir.resources.coding import Coding
 from fhir.resources.diagnosticreport import DiagnosticReport
 from fhir.resources.domainresource import DomainResource
 from fhir.resources.encounter import Encounter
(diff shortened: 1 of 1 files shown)


joverlee521:
[Ingest follow up survey for SCAN REDCap](https://trello.com/c/KqwnAbeQ/530-ingest-follow-up-survey-for-scan-redcap)

joverlee521:
> I also think the latest commit pushed, [f182d0f](https://github.com/seattleflu/id3c/commit/f182d0fbe4c58b6747c231ca224870933f2ab22b), should not be something that happens during ETL. I think it belongs instead in a query. A couple of reasons for this:
> 
> 1. The internal ids of records should only ever be referred to by real FKs, which can be properly constrained and verified. The external/logical/"business" identifiers of records are more suitable to being referenced by non-FKs.
> 2. The internal id associated with an identifier can change, whereas the identifier itself should not.
> 
> These are the reasons why FHIR itself has two concepts of ids: internal resource `id`s and external/logical/business `identifier`s. ID3C uses the same concepts in a similar way.

Yeah, that makes sense. I did it that way to make it easy to change to a FK in the future if we wanted to formalize the data structure and add the FK to the encounters table.

joverlee521:
> I started a discussion about processing of the reason codes [in Slack](https://seattle-flu-study.slack.com/archives/GJXFQ209K/p1588973950014100) because it impacts two PRs (this one and [seattleflu/id3c-customizations#102](https://github.com/seattleflu/id3c-customizations/pull/102)), and I wanted to make sure we figure it out before anything gets merged.

I'll update to just ingest the JSON object as is without parsing out the SNOMED code.

tsibley:
> Yeah, that makes sense. I did it that way to make it easy to change to a FK in the future if we wanted to formalize the data structure and add the FK to the encounters table.

Nod. We'll be able to do the join and update at that time if we do.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -463,6 +476,22 @@ def process_patient_language(patient: Patient) -> Optional[str]:
...
+    if not encounter.reasonCode:
+        return None
+
+    reason_codes: List[Coding] = []
+
+    for codeable_concept in encounter.reasonCode:
+        codes = codeable_concept.coding
+        reason_codes.extend(codes)
+
+    return [r.as_json() for r in reason_codes]
```
tsibley, line 492: FWIW, a more Pythonic approach here (which I do think is more clear) is a list comprehension:

```python
    return [
        coding.as_json()
        for concept in encounter.reasonCode
        for coding in concept.coding ]
```

PR #150, opened 2020-05-08
optimize REDCap DET ETL

joverlee521:
1. Aggregate DETs by record. Skip old DETs and only process the latest
complete DET for each record.

2. Make a batch request for records instead of requesting each record
individually from the REDCap API.

Note that this ETL still has the caveat of not being able to handle
records with repeating instruments or longitudinal events.

### DIFF
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -126,7 +126,7 @@ def command_for_project(name: str,
                         # the current instrument is complete
                         if not include_incomplete and not is_complete(instrument, det.document):
                             LOG.debug(f"Skipping incomplete or unverified REDCap DET {det.id}")
-                            mark_skipped(db, det.id, etl_id)
+                            mark_skipped(db, det.id, etl_id, "incomplete/unverified")
                             continue
 
                         # Check if this is record has an older DET
@@ -135,7 +135,7 @@ def command_for_project(name: str,
                         # semantics of our receiving tables
                         elif first_complete_dets.get(record_id):
                             LOG.debug(f"Skipping latest REDCap DET {det.id}")
-                            mark_skipped(db, det.id, etl_id)
+                            mark_skipped(db, det.id, etl_id, "repeat record id")
 
                         else:
                             first_complete_dets[record_id] = det
@@ -185,13 +185,13 @@ def command_for_project(name: str,
                         if incomplete_instruments:
                             LOG.debug(f"The following required instruments «{incomplete_instruments}» are not yet marked complete. " + \
                                       f"Skipping REDCap DET {det.id}")
-                            mark_skipped(db, det.id, etl_id)
+                            mark_skipped(db, det.id, etl_id, "required instruments incomplete")
                             continue
 
                         bundle = routine(db = db, cache = cache, det = det, redcap_record = redcap_record)
 
                         if not bundle:
-                            mark_skipped(db, det.id, etl_id)
+                            mark_skipped(db, det.id, etl_id, "insufficient data in record")
                             continue
 
                         if log_output:
@@ -257,7 +257,7 @@ def skip_missing_records(db: DatabaseSession, dets: Iterable[Any], etl_id: dict)
     for det in dets:
         with db.savepoint(f"redcap_det {det.id}"):
             LOG.debug(f"REDCap record is missing or invalid.  Skipping REDCap DET {det.id}")
-            mark_skipped(db, det.id, etl_id)
+            mark_skipped(db, det.id, etl_id, "invalid record")
 
 
 def mark_loaded(db: DatabaseSession, det_id: int, etl_id: dict, bundle_uuid: str) -> None:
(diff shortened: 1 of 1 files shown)


joverlee521:
[REDCap optimization](https://trello.com/c/1QCxnAUP/558-redcap-optimization)

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -122,12 +122,37 @@ def decorated(*args, db: DatabaseSession, log_output: bool, **kwargs):
...
 
-                        redcap_record = get_redcap_record_from_det(det.document, raw_coded_values)
+                        # Check if this is record has an older DET
+                        # Skip older DET in favor of the latest DET
+                        elif latest_complete_dets.get(record_id):
+                            old_det = latest_complete_dets[record_id]
+                            LOG.debug(f"Skipping older REDCap DET {old_det.id}")
+                            mark_skipped(db, old_det.id, etl_id)
+
+                        latest_complete_dets.update({record_id: det})
```
kairstenfay, line 132: Since we're working on a PR related to optimization, I wanted to point out that since we're only updating one key-value, python's "regular" (don't know a better word here) way of updating a single value in a dict is faster than using `.update()`.

e.g.
```
import timeit

timeit.timeit("foo.update({'a': 4})", setup="foo = {'a': 3}")
```
Runs in ~0.18s, whereas
```
timeit.timeit("foo['a'] = 4", setup="foo = {'a': 3}")
```
runs in ~0.06s (about 1/3 that of `.update()`).

joverlee521: You are absolutely right. I was probably overthinking things when I wrote this. 😄

kairstenfay:
Running this locally now and WOW it's so much faster! :heart_eyes:

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -219,6 +227,23 @@ def get_redcap_api_token(api_url: str) -> str:
     return token
 
 
+def get_redcap_records(project: Project, record_ids: List[str], raw: bool) -> Optional[List[Record]]:
```
kairstenfay, line 230: I think, if given a an empty list of `record_ids`, we would want this function to return without fetching from REDCap. The current behavior is that I get a log output of 
```
[2020-05-12 13:52:00] Fetching REDCap project 20759
[2020-05-12 13:52:01] Fetching 0 REDCap records from project 20759
```
followed by a long pause, then a bunch of log statements about duplicate REDCap record IDs. These aren't real duplicate REDCap record IDs but the result of the `KeyError` exception where we're processing records returned from REDCap but don't have those records tracked locally in `latest_complete_dets`.

kairstenfay: This makes me wonder if `get_redcap_records` is doing what we want; if we're passing it an empty list of `record_ids`, why are so many `records` being returned? (e.g. 16619 for `scan-en`)

joverlee521: OH. If you don't provide `record_id` to the REDCap API then it returns all records by default.

joverlee521: Updated in latest commit

joverlee521:
> I think it'd be easier to reason about how documents/records are processed if we still preserve the first-in-first-out semantics of our receiving tables. This would mean processing a) the _first_ "complete" DET rather than last one and b) records in the order the DETs appear, rather than by record id (which is what REDCap returns).

> When running with `--commit`, an error during the `for redcap_record in redcap_records` loop may leave a bunch of DETs marked skipped (from the earlier pass through DETs) even if the ETL never made it to processing their associated record. I'm not sure if this might have subtle unintended effects, but it feels a little weird to me. I think of the ETL processes as something that march linearly through the receiving tables (or a subset of each table) and only ratchet forward (until a version bump).

Both of these problems stem from the process not marching linearly through the receiving table. I'm thinking to create another dictionary that holds the `redcap_det_id` as the keys and the process for that det record as the value:
```python
all_dets = {
      1: {"status": "load", "record_id": "1"},
      2: {"status": "skip", "skip_reason": "repeat redcap record"},
      3: {"status": "skip", "skip_reason": "incomplete or unverified det"},
      4: {"status" : "load", "record_id": "2"}
}
```
This dict is filled in during the first loop of the det records. A batch REDCap request is made. Then finally we process everything by going through the sorted `all_dets` dict.

tsibley:
> Both of these problems stem from the process not marching linearly through the receiving table. I'm thinking to create another dictionary that holds the `redcap_det_id` as the keys and the process for that det record as the value:
> …
> This dict is filled in during the first loop of the det records. A batch REDCap request is made. Then finally we process everything by going through the sorted `all_dets` dict.

Nod. An approach like this makes sense to me. A couple observations:

* `all_dets` can probably just be a list instead of a dict, which avoids the need to sort later.

* The initial loop over `receiving.redcap_det` may be large if we've bumped the ETL revision. For this reason, good to keep the cached data structure `all_dets` minimal.

* The fetched batch of REDCap records will need to be transformed from a list of dicts to an indexed/keyed data structure to support lookups by record id, e.g. a dict of dicts keyed by record id. This transformation will have to account for duplicate record ids, but the [same approach as currently](https://github.com/seattleflu/id3c/blob/08460c71aeef637274fc4711a6c51a987eaa45cb/lib/id3c/cli/command/etl/redcap_det.py#L191-L193) is still fine.

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -257,17 +257,17 @@ def skip_missing_records(db: DatabaseSession, dets: Iterable[Any], etl_id: dict)
...
 def mark_loaded(db: DatabaseSession, det_id: int, etl_id: dict, bundle_uuid: str) -> None:
     LOG.debug(f"Marking REDCap DET record {det_id} as loaded")
     mark_processed(db, det_id, {**etl_id, "status": "loaded", "fhir_bundle_id": bundle_uuid})
 
 
-def mark_skipped(db: DatabaseSession, det_id: int, etl_id: dict) -> None:
+def mark_skipped(db: DatabaseSession, det_id: int, etl_id: dict, reason: str) -> None:
     LOG.debug(f"Marking REDCap DET record {det_id} as skipped")
-    mark_processed(db, det_id, {**etl_id, "status": "skipped"})
+    mark_processed(db, det_id, {**etl_id, "status": "skipped", "skip_reason": reason})
```
kairstenfay, line 284: Adding skip reason here will be very helpful

PR #151, opened 2020-05-18
Materialized view refresher

joverlee521:
Add role and command to refresh materialized views.

Moved from id3c-customizations with @tsibley's suggestion:
>Hmm, and actually, what about building this into a core ID3C CLI command? It's something I expect will be not uncommon. (I'm actually really surprised it took us this long to need materialized views! I'd thought we'd hit a need for them much sooner.) I guess this would mean moving the materialized-view-refresher role to core, though, which is maybe more shuffling than we want to bother with right now…

### DIFF
--- a/lib/id3c/cli/command/refresh_materialized_view.py
+++ b/lib/id3c/cli/command/refresh_materialized_view.py
@@ -28,7 +28,7 @@ def refresh_materialized_view(schema_name, view_name, db: DatabaseSession):
 
     LOG.info(f"Refreshing materialized view «{schema_name}.{view_name}»")
 
-    db.cursor("refresh materialized view").execute("""
+    db.cursor().execute("""
         select refresh_materialized_view(%s, %s)
     """, (schema_name, view_name ))
 
--- a/lib/id3c/cli/command/refresh_materialized_view.py
+++ b/lib/id3c/cli/command/refresh_materialized_view.py
@@ -1,0 +1,36 @@
+"""
+Refresh materialized view in ID3C.
+"""
+import click
+import logging
+from id3c.cli import cli
+from id3c.cli.command import with_database_session
+from id3c.db.session import DatabaseSession
+
+
+LOG = logging.getLogger(__name__)
+
+
+@cli.command("refresh-materialized-view")
+@with_database_session
+
+@click.argument("schema-name",
+    metavar = "<schema>",
+    nargs = 1)
+@click.argument("view-name",
+    metavar = "<view>",
+    nargs = 1)
+
+def refresh_materialized_view(schema_name, view_name, db: DatabaseSession):
+    """
+    Refresh materialized view <schema>.<view> in ID3C.
+    """
+
+    LOG.info(f"Refreshing materialized view «{schema_name}.{view_name}»")
+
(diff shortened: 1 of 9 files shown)


lib/id3c/cli/command/refresh_materialized_view.py
```diff
@@ -0,0 +1,36 @@
...
+def refresh_materialized_view(schema_name, view_name, db: DatabaseSession):
+    """
+    Refresh materialized view <schema>.<view> in ID3C.
+    """
+
+    LOG.info(f"Refreshing materialized view «{schema_name}.{view_name}»")
+
+    db.cursor("refresh materialized view").execute("""
+        select refresh_materialized_view(%s, %s)
+    """, (schema_name, view_name ))
```
tsibley, line 33: By explicitly naming the cursor, this creates a server-side cursor. This ends up suppressing any errors. For example, `id3c refresh-materialized-view doesnotexist notevenalittle` reports success. I'm not entirely sure why, but I assume this is because Pg defers raising the exception, or maybe executing the query at all, until the cursor is first read. Since we never read from the cursor, nothing happens. Removing the argument to `db.cursor()` means we use a client-side cursor, which works as expected.

schema/deploy/roles/materialized-view-refresher/grants.sql
```diff
@@ -0,0 +1,43 @@
...
+revoke all on schema receiving, warehouse, shipping from "materialized-view-refresher";
+revoke all on all tables in schema receiving, warehouse, shipping from "materialized-view-refresher";
+
+grant connect on database :"DBNAME" to "materialized-view-refresher";
+
+-- Regular users cannot refresh materialized views because only the view
+-- owner has permission to refresh them.
+-- Create a function that runs in the security context of its owner so that
+-- other users can refresh the view.
+create or replace function public.refresh_materialized_view(view_schema text, view_name text)
```
tsibley, line 16: I was a bit surprised to see the function definition in this grants file. I think I know why you did so; not sure I have a better solution. It might be a little less weird to move the function create and execute grants to _functions/refresh-materialized-view_?

joverlee521: I don't really have a good reason for this. I was following the the grants set up of the [idle-session-disconnector](https://github.com/seattleflu/id3c-customizations/blob/master/schema/deploy/roles/idle-session-disconnector/grants.sql).

lib/id3c/cli/command/refresh_materialized_view.py
joverlee521: Thank you for catching this!

schema/deploy/roles/materialized-view-refresher/grants.sql
tsibley: Ah, ok. I think I probably missed that the function was inside the grants change in that PR. :-)

One reason for co-locating the grants to the function with the function definition itself is that reworking either doesn't create awkward inter-change dependencies in sqitch. I think there's room for improvement in some of the sqitch patterns we adopted early on (generally at my suggestion or example, so mea culpa), but I haven't had time to write up how I think we could do things differently.

That's all to say, this is fine for now. :-)

PR #153, opened 2020-06-03
etl fhir: ingest `effectiveDateTime` from Diagnostic Reports

joverlee521:
Ingests into `warehouse.presence_absence.details.effective_datetime`

This change was made due to a request from the self-test team to report the time at which the participants performed the self-test for Ellume.


joverlee521:
[Request from Rachel(self-test): Ingest and provide self-test timestamp for Ellume.](https://trello.com/c/IVGkeeQ9/591-request-from-rachelself-test-ingest-and-provide-self-test-timestamp-for-ellume)

joverlee521:
If we don't think this field is important enough to ingest, we can also write queries against the receiving table to pull out this field in Metabase.

PR #156, opened 2020-07-11
Redcap det event

kairstenfay:
Add an `--event` option to the REDCap DET generate command.
Add an `events` attribute to the REDCap `Project` class.

Do some minor reworking/refactoring of the generate command.

### DIFF
--- a/lib/id3c/cli/command/redcap_det.py
+++ b/lib/id3c/cli/command/redcap_det.py
@@ -39,7 +39,7 @@ def redcap_det():
            "Used as a sanity check that the correct API token is provided.",
     required = True)
 
-@click.option("--token-name",
+@click.option("--token",
     metavar = "<token-name>",
     help = "The name of the environment variable that holds the API token",
     default = "REDCAP_API_TOKEN")
@@ -55,7 +55,7 @@ def redcap_det():
            "Format must be YYYY-MM-DD HH:MM:SS (e.g. '2019-01-01 00:00:00')")
 
 @click.option("--instrument", "instruments",
-    metavar = "<name>",
+    metavar = "<instrument-name>",
     help = "Limit generated DET notifications to the named instrument; may be used multiple times",
     multiple = True)
 
@@ -64,7 +64,7 @@ def redcap_det():
     is_flag = True,
     flag_value = True)
 
-def generate(record_ids: List[str], project_id: int, token_name: str, since_date: str, until_date: str, instruments: List[str], include_incomplete: bool):
+def generate(record_ids: List[str], project_id: int, token: str, since_date: str, until_date: str, instruments: List[str], include_incomplete: bool):
     """
     Generate DET notifications for REDCap records.
 
@@ -74,7 +74,7 @@ def generate(record_ids: List[str], project_id: int, token_name: str, since_date
     of specific record ids with date filters, so this command does not either.
 
     Requires environmental variables REDCAP_API_URL and REDCAP_API_TOKEN (or
-    whatever you passed to --token-name).
+    whatever you passed to --token).
 
     DET notifications are output for all completed instruments for each record
     by default.  Pass --include-incomplete to output DET notifications for
@@ -84,7 +84,7 @@ def generate(record_ids: List[str], project_id: int, token_name: str, since_date
     All DET notifications are output to stdout as newline-delimited JSON
     records.  You will likely want to redirect stdout to a file.
     """
-    api_token = os.environ[token_name]
+    api_token = os.environ[token]
     api_url = os.environ['REDCAP_API_URL']
 
     project = Project(api_url, api_token, project_id)
--- a/lib/id3c/cli/command/redcap_det.py
+++ b/lib/id3c/cli/command/redcap_det.py
@@ -114,6 +114,10 @@ def generate(record_ids: List[str], project_id: int, token: str, since_date: str
 
     if events:
         LOG.debug(f"Producing DET notifications for the following events: {events}")
+        unknown_events = set(events) - set(project.events)
+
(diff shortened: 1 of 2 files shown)


kairstenfay:
[Set up DET generation for group enrollments](https://trello.com/c/m4HM18kU/660-set-up-det-generation-for-group-enrollments)

lib/id3c/cli/redcap.py
```diff
@@ -69,6 +70,18 @@ def instruments(self) -> List[str]:
...
 
 
+    @property
+    def events(self) -> List[str]:
+        """
+        Names of all events in this REDCap project.
+        """
+        if not self._events:
+            nameof = itemgetter("unique_event_name")
+            self._events = list(map(nameof, self._fetch("event")))
```
joverlee521, line 80: Not all REDCap projects have events, so I'm not sure if we need some sort of special handling here.

joverlee521: Using the REDCap API Playground for the Swab & Send project, it looks like this API request returns a default value `event_1_arm_1` for `unique_event_name`. So maybe it's not a problem?

tsibley: @joverlee521 Your first intuition is correct. Many (maybe most?) of our REDCap projects are "longitudinal" projects, which means they define events. "Classic" projects, however, do not use events and the `content=event` API call will return an error:

    HTTP/1.1 400 Bad Request
    {"error":"You cannot export events for classic projects"}

This setting is determined by the project's "Use longitudinal data collection with defined events?" checkbox on the project setup page. There's more documentation about how classic vs. longitudinal projects and how they interact with repeatable instruments/events on [this doc page](https://redcap.iths.org/redcap_v9.9.0/ProjectSetup/repeating_instruments_events_info.php).

lib/id3c/cli/command/redcap_det.py
```diff
@@ -111,23 +118,41 @@ def generate(record_ids: List[str], project_id: int, token_name: str, since_date
...
 
+def assert_known_attribute_value(project: Project, attribute: str, values: List[str]):
+    """
+    Throws an :class:`AssertionError` if any of the given *values* are not
+    contained in the *attribute* of the given *project*.
+    """
+    unknown_values = set(values) - set(getattr(project, attribute))
+
+    assert not unknown_values, \
+        f"The following {attribute} aren't in the REDCap project: {unknown_values}"
```
tsibley, line 153: I ever-so-slightly prefer the previous message text:

> The following --instrument names …

over 

> The following instruments…

because I think the former very clearly ties the problematic values back to the passed in option. With the latter, I wouldn't necessarily know right away it was an error with how I ran the command and not some issue with REDCap or a bug in the code.

lib/id3c/cli/command/redcap_det.py
```diff
@@ -111,23 +118,41 @@ def generate(record_ids: List[str], project_id: int, token_name: str, since_date
...
 
-    unknown_instruments = set(instruments) - set(project.instruments)
-
-    assert not unknown_instruments, \
-        f"The following --instrument names aren't in the REDCap project: {unknown_instruments}"
 
     for record in records:
+        if record['redcap_event_name'] not in events:
+            continue
+
```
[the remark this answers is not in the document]
kairstenfay: Ah, thank you! I should have checked the REDCap API docs for that but instead I tried to speedily get this code out the door before my SISMID courses. I'll push a fix.

lib/id3c/cli/redcap.py
kairstenfay: Thanks. It's not clear to me whether a classic REDCap project would silently ignore an `events` parameter in the data payload or throw an HTTP error. I'm also not sure how to test this, as seemingly every REDCap project of ours is longitudinal, not classic.

tsibley: The project we use to coordinate clinical data pulls from UW is "classic": https://redcap.iths.org/redcap_v9.9.0/index.php?pid=19915

tsibley: (which is the project I used to see that making an API request with `content=event` fails)

lib/id3c/cli/command/redcap_det.py
kairstenfay: That makes sense. I added a new argument for the option name which makes the function a bit harder to use, but it doesn't make any assumptions about how to pluralize an option name.

lib/id3c/cli/redcap.py
kairstenfay: Thanks!

lib/id3c/cli/redcap.py
```diff
@@ -69,6 +71,24 @@ def instruments(self) -> List[str]:
...
+        """
+        if self._events is None:
+            nameof = itemgetter("unique_event_name")
+
+            try:
+                self._events = list(map(nameof, self._fetch("event")))
+            except HTTPError:
+                LOG.debug("HTTP error encountered when fetching events. "
+                "Is this a classic project?")
+                self._events = []
```
tsibley, line 87: This will mask many kinds of real network errors unrelated to the availability of events in the project. Could we check the `is_longitudinal` field returned by the project info endpoint instead? It should already be in `self._details["is_longitudinal"]` and have a `"0"` or `"1"` value.

kairstenfay: Thanks, I wasn't a fan of this general HTTP Error exception, either. I thought I looked through all the fields for one matching "longitudinal", but I guess I didn't. I'll push a fix.

kairstenfay: Updated in the "Account for classic projects" commit.

PR #159, opened 2020-07-23
redcap is_complete: return None if {instrument}_complete does not exist

joverlee521:
Extend encapsulation by supporting three-valued logic. Makes it easier
for us to differentiate between an incomplete instrument and the
instrument not existing in data.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -247,6 +247,9 @@ def is_complete(instrument: str, data: dict) -> bool:
 
     The *data* may be a DET notification or a record.
 
+    Will return `None` if the *instrument*_complete key does not exist in
+    given *data*.
+
     >>> is_complete("test", {"test_complete": "Complete"})
     True
     >>> is_complete("test", {"test_complete": 2})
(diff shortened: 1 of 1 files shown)


joverlee521:
[Update `is_complete()` to return `None` if instrument does not exist](https://trello.com/c/WCsPqnXD/693-update-iscomplete-to-return-none-if-instrument-does-not-exist)

lib/id3c/cli/redcap.py
```diff
@@ -255,10 +258,15 @@ def is_complete(instrument: str, data: dict) -> bool:
     True
     >>> is_complete("test", {"test_complete": "Incomplete"})
     False
-    >>> is_complete("test", {})
-    False
+    >>> is_complete("test", {}) is None
+    True
```
jccraft-uw, line 262: The is_complete("test", {}) case will return None or True? Is the "True" dangling here?

joverlee521: `is_complete("test", {})` should return `None`. This is checking `is_complete("test", {}) is None`, which should return `True`. 

There might be a better way of writing this particular test?

jccraft-uw: That makes sense!

PR #160, opened 2020-07-29
Ingest incident report manifest

jccraft-uw:
In this PR, the manifest and etl/manifest modules are updated to accommodate other types of manifests that may have a collection barcode but not a sample barcode and may not have a collected date. A google module is also added to io, and in this module is utility code for downloading a document from Google Drive.

Restrictions are loosened so that a sample barcode is not required. Checks are in place, though, to ensure that a sample barcode or a collection barcode (or both) are provided. We need at least one of the two barcodes. 

For manifest YAML configuration (and the options for the parse method) three settings are now special:
1. sample (This was already there. The change is that is no longer required). In the receiving.manifest table, data mapped to 'sample' end up the in the JSON with the 'sample' key.
2. collection: In the receiving.manifest table, data mapped to 'collection' end up the in the JSON with the 'collection' key.
3. collected_date: In the receiving.manifest table, data mapped to 'collected_date' end up the in the JSON with the 'collected_date' key.
A check ensures that extra_columns does not contain a column with the same name as one of those three columns.

Existing manifest YAMLs have to be udpated to accommodate these changes. For those see
https://github.com/seattleflu/specimen-manifests/pull/15/

In the SQL in the ETL code I build the update statement dynamically based on which keys existing in the record being processed.

Testing:
On my local machine I truncated my receiving.manifest and warehouse.sample tables. I ingested aliquot records from S3 and incident report records from Google Sheets. I confirmed that the records looked right in receiving.manifest and then ran the ETL to populate warehouse.sample from them. As this ran I looked for failed assertions. When it was done I looked at which records got created, updated, and skipped and made sure those made sense. Finally, I cleared out the processing_log column on the receive.manifest table and ran the ETL again to exercise the update path. I explicitly tested the update case for one record from the aliquot file and one record from the incident report file to make sure that the correct columns got updated.

### DIFF
--- a/lib/id3c/cli/command/etl/manifest.py
+++ b/lib/id3c/cli/command/etl/manifest.py
@@ -3,7 +3,7 @@ Process sample manifest records into the relational warehouse.
 
 Manifest records are JSON documents in the ``receiving.manifest`` table which
 have a sample barcode, an optional collection barcode, and other metadata.
-These records are parsed out of Excel workbooks into JSON documents using
+These records are parsed out of spreadsheets into JSON documents using
 ``id3c manifest parse`` and received via ``id3c manifest upload``.  This ETL
 command is the final step in creating/updating the samples in our data
 warehouse.
@@ -16,6 +16,7 @@ or updated records.
 import click
 import logging
 from datetime import datetime, timezone
+from psycopg2 import sql
 from typing import Any, Tuple, Optional
 from id3c.cli.command import with_database_session
 from id3c.db import find_identifier
@@ -98,55 +99,75 @@ def etl_manifest(*, db: DatabaseSession):
         with db.savepoint(f"manifest record {manifest_record.id}"):
             LOG.info(f"Processing record {manifest_record.id}")
 
-            # Convert sample barcode to full identifier, ensuring it's
-            # known and from the correct identifier set.
-            sample_barcode = manifest_record.document.pop("sample")
-            sample_identifier = find_identifier(db, sample_barcode)
+            # When updating an existing row, update the identifiers
+            # only if the record has both the 'sample' and
+            # 'collection' keys.
+            should_update_identifiers = "sample" in manifest_record.document \
+                and "collection" in manifest_record.document
 
-            if not sample_identifier:
+             # Sample collection date
+             # Don't pop this entry off the document. For backwards
+             # compatibility reasons, keep it in the document so that 'date'
+             # also gets written to the 'details' column in warehouse.sample.
+            collected_date = manifest_record.document.get("date", None)
+
+            # Attempt to find barcodes and their related identifiers
+            sample_barcode = manifest_record.document.pop("sample", None)
+            sample_identifier = find_identifier(db, sample_barcode) if sample_barcode else None
+            collection_barcode = manifest_record.document.pop("collection", None)
+            collection_identifier = find_identifier(db, collection_barcode) if collection_barcode else None
+
+            # Skip a record if it has no sample_identifier and no collection_identifier
+            if not sample_identifier and not collection_identifier:
+                LOG.warning(f"Skipping record «{manifest_record.id}» because it has neither a known sample "
+                + "identifier nor a collection identifier")
+                mark_skipped(db, manifest_record.id)
+                continue
+
+            # Skip a record if it has a sample barcode but the barcode doesn't match an identifier
+            if sample_barcode and not sample_identifier:
                 LOG.warning(f"Skipping sample with unknown sample barcode «{sample_barcode}»")
                 mark_skipped(db, manifest_record.id)
                 continue
 
-            if (manifest_record.document.get("sample_type") and
-                manifest_record.document["sample_type"] == "rdt"):
-                assert sample_identifier.set_name in expected_identifier_sets["rdt"], \
-                    (f"Sample identifier found in set «{sample_identifier.set_name}», " +
-                    f"not {expected_identifier_sets['rdt']}")
-            else:
-                assert sample_identifier.set_name in expected_identifier_sets["samples"], \
-                    (f"Sample identifier found in set «{sample_identifier.set_name}», " +
-                    f"not {expected_identifier_sets['samples']}")
-
-            # Optionally, convert the collection barcode to full
-            # identifier, ensuring it's known and from the correct
-            # identifier set.
-            collection_barcode = manifest_record.document.pop("collection", None)
-            collection_identifier = find_identifier(db, collection_barcode) if collection_barcode else None
-
+            # Skip a record if it has a collection barcode but the barcode doesn't match an identifier
             if collection_barcode and not collection_identifier:
                 LOG.warning(f"Skipping sample with unknown collection barcode «{collection_barcode}»")
                 mark_skipped(db, manifest_record.id)
                 continue
 
-            assert not collection_identifier \
-                or collection_identifier.set_name in expected_identifier_sets["collections"], \
-                    f"Collection identifier found in set «{collection_identifier.set_name}», not {expected_identifier_sets['collections']}" # type: ignore
+             # Skip a record if the collection identifier is from an unexpected set
+            if collection_identifier and collection_identifier.set_name not in expected_identifier_sets["collections"]:
+                LOG.warning(f"Skipping sample because collection identifier found in set «{collection_identifier.set_name}», not \
+                    {expected_identifier_sets['collections']}")
+                mark_skipped(db, manifest_record.id)
+                continue
 
-            # Sample collection date
-            collection_date = manifest_record.document.get("date")
+            # Validate the sample identifer and assert if a record fails
+            if sample_identifier:
+                if (manifest_record.document.get("sample_type") and
+                    manifest_record.document["sample_type"] == "rdt"):
+                    assert sample_identifier.set_name in expected_identifier_sets["rdt"], \
+                        (f"Sample identifier found in set «{sample_identifier.set_name}»," +
+                        f"not {expected_identifier_sets['rdt']}")
+                else:
+                    assert sample_identifier.set_name in expected_identifier_sets["samples"], \
+                        (f"Sample identifier found in set «{sample_identifier.set_name}», " +
+                        f"not {expected_identifier_sets['samples']}")
+
 
             # Upsert sample cooperatively with enrollments ETL routine
             #
             # The details document was intentionally modified by two pop()s
-            # earlier to remove barcodes that were looked up.  The
-            # rationale is that we want just one clear place in the
+            # earlier to remove barcodes that were looked up.
+            # The rationale is that we want just one clear place in the
             # warehouse for each piece of information.
             sample, status = upsert_sample(db,
-                identifier            = sample_identifier.uuid,
-                collection_identifier = collection_identifier.uuid if collection_identifier else None,
-                collection_date       = collection_date,
-                additional_details    = manifest_record.document)
+                update_identifiers          = should_update_identifiers,
+                identifier                  = sample_identifier.uuid if sample_identifier else None,
+                collection_identifier       = collection_identifier.uuid if collection_identifier else None,
+                collection_date             =  collected_date,
+                additional_details          = manifest_record.document)
 
             mark_loaded(db, manifest_record.id,
                 status = status,
@@ -156,6 +177,7 @@ def etl_manifest(*, db: DatabaseSession):
 
 
 def upsert_sample(db: DatabaseSession,
+                  update_identifiers: bool,
                   identifier: str,
                   collection_identifier: Optional[str],
                   collection_date: Optional[str],
@@ -207,18 +229,25 @@ def upsert_sample(db: DatabaseSession,
         sample = samples[0]
 
         LOG.info(f"Updating existing sample {sample.id}")
-        sample = db.fetch_row("""
+
+        # Update identifier and collection_identifier if update_identifiers is True
+        identifiers_update_composable = sql.SQL("""
+             identifier = %(identifier)s,
+                collection_identifier = %(collection_identifier)s, """)  \
+                    if update_identifiers else sql.SQL("")
+
+        sample = db.fetch_row(sql.SQL("""
             update warehouse.sample
-               set identifier = %(identifier)s,
-                   collection_identifier = %(collection_identifier)s,
-                   collected = coalesce(date_or_null(%(collection_date)s), collected),
-                   details = coalesce(details, '{}') || %(additional_details)s
+                set {}
+                    collected = coalesce(date_or_null(%(collection_date)s), collected),
+                    details = coalesce(details, {}) || %(additional_details)s
 
              where sample_id = %(sample_id)s
 
             returning sample_id as id, identifier, collection_identifier, encounter_id
-            """,
-            { **data, "sample_id": sample.id })
+            """).format(identifiers_update_composable,
+                sql.Literal(Json({}))),
+                { **data, "sample_id": sample.id })
 
         assert sample.id, "Update affected no rows!"
 
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -25,7 +25,7 @@ from deepdiff import DeepHash
 from hashlib import sha1
 from os import chdir
 from os.path import dirname
-from typing import Iterable, List, Tuple, Union
+from typing import Iterable, List, Optional, Tuple, Union
 from id3c.cli import cli
 from id3c.cli.io import LocalOrRemoteFile, urlopen
 from id3c.cli.io.pandas import read_excel
@@ -73,6 +73,13 @@ def manifest():
            "Option value is parsed as a YAML fragment, so additional options supported by the sibling command \"parse-with-config\" may be inlined for testing, but you're likely better off using a config file at that point.",
     multiple = True)
 
+@click.option("--row-filter",
+    metavar = "<query>",
+    help = "The pandas query to filter rows (using the python engine) in the manifest.  "
+           "Column names refer to columns in the manifest itself.  "
+           "Example: `corrective action`.notnull() and `corrective action`.str.lower().str.startswith(\"discard\") ",
+    required = False)
+
 def parse(**kwargs):
     """
     Parse a single manifest workbook sheet.
@@ -84,6 +91,11 @@ def parse(**kwargs):
     The required --sample-column option specifies the name of the column
     containing the sample barcode.  Other columns may be extracted into the
     manifest records as desired using the --extra-column option.
+
+    The row-filter entry specifies a pandas query to filter
+    (using the python engine) rows in the manifest. Column names refer to columns
+    in the manifest itself.
+    Example: `corrective action`.notnull() and `corrective action`.str.lower().str.startswith("discard")
 
     Manifest records are output to stdout as newline-delimited JSON records.
     You will likely want to redirect stdout to a file.
@@ -147,6 +159,14 @@ def parse_using_config(config_file):
           test_results: "Test ResulTS"
         ...
 
+    The sample_column entry specifies the name of the column
+    containing the sample barcode.
+
+    The row_filter entry specifies a pandas query to filter
+    (using the python engine) rows in the manifest. Column names refer to columns
+    in the manifest itself.
+    Example: `corrective action`.notnull() and `corrective action`.str.lower().str.startswith("discard")
+
     The key: value pairs in "extra_columns" name destination record fields (as
     the key) and source columns (as the value).  For most source columns, a
     simple string name (or shell-glob pattern) is enough.  Other behaviour is
@@ -194,7 +214,8 @@ def parse_using_config(config_file):
                 "sheet": config["sheet"],
                 "sample_column": config["sample_column"],
                 "extra_columns": list(config.get("extra_columns", {}).items()),
-                "sample_type": config.get("sample_type")
+                "sample_type": config.get("sample_type"),
+                "row_filter" : config.get("row_filter")
             }
         except KeyError as key:
             LOG.error(f"Required key «{key}» missing from config {config}")
@@ -208,7 +229,8 @@ def _parse(*,
            sheet,
            sample_column,
            extra_columns: List[Tuple[str, Union[str, dict]]] = [],
-           sample_type = None):
+           sample_type = None,
+           row_filter: Optional[str] = None):
     """
     Internal function powering :func:`parse` and :func:`parse_using_config`.
     """
@@ -236,6 +258,11 @@ def _parse(*,
                 .str.strip()
                 .replace({pandas.NA: ""})
                 .replace({"": None, "na": None})))
+
+    # If a filter query was provided filter the manifest rows
+    # using the python engine.
+    if row_filter:
+        manifest = manifest.query(row_filter, engine="python")
 
     # Construct parsed manifest by copying columns from source to destination.
     # This approach is used to allow the same source column to end up as
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -28,6 +28,7 @@ from os.path import dirname
 from typing import Iterable, List, Optional, Tuple, Union
 from id3c.cli import cli
 from id3c.cli.io import LocalOrRemoteFile, urlopen
+from id3c.cli.io.google import export_file_from_google_drive, extract_document_id_from_google_url, GoogleDriveExportFormat
 from id3c.cli.io.pandas import read_excel
 from id3c.db.session import DatabaseSession
 from id3c.db.datatypes import as_json, Json
@@ -45,7 +46,7 @@ def manifest():
 
 
 @manifest.command("parse")
-@click.argument("workbook", metavar = "<manifest.xlsx>")
+@click.argument("workbook", metavar = "<filepath>")
 
 @click.option("--sheet",
     metavar = "<name>",
@@ -84,8 +85,8 @@ def parse(**kwargs):
     """
     Parse a single manifest workbook sheet.
 
-    <manifest.xlsx> must be a path or URL to an Excel workbook with at least
-    one sheet in it, identified by name using the required option --sheet.
+    <filepath> must be a path or URL to an Excel workbook or Google Sheets
+    spreadsheet with at least one sheet in it, identified by name using the required option --sheet.
     Supported URL schemes include http[s]:// and s3://, as well as others.
 
     The required --sample-column option specifies the name of the column
@@ -234,10 +235,20 @@ def _parse(*,
     """
     Internal function powering :func:`parse` and :func:`parse_using_config`.
     """
-    LOG.debug(f"Reading Excel workbook «{workbook}»")
+    # Determine if the workbook URL is for a Google Document and if so
+    # retrieve the Google Sheets file as an Excel spreadsheet. Otherwise,
+    # retrieve it using urlopen.
+    google_docs_document_id = extract_document_id_from_google_url(workbook)
 
-    with urlopen(workbook, "rb") as file:
-        workbook_bytes = file.read()
+    if google_docs_document_id:
+        LOG.debug(f"Reading Google Sheets document «{workbook}»")
+        with export_file_from_google_drive(google_docs_document_id, GoogleDriveExportFormat.EXCEL) as file:
+            workbook_bytes = file.read()
+
+    else:
+        LOG.debug(f"Reading Excel workbook «{workbook}»")
+        with urlopen(workbook, "rb") as file:
+            workbook_bytes = file.read()
 
     LOG.debug(f"Parsing sheet «{sheet}» in workbook «{workbook}»")
 
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -390,7 +390,7 @@ def select_columns(table: pandas.DataFrame, name: str) -> pandas.DataFrame:
     *table*.
     """
     pattern = re.compile(fnmatch.translate(name), re.IGNORECASE)
-    matches = list(filter(pattern.match, table.columns))
+    matches = list(filter(pattern.match, table.columns.astype(str)))
 
     assert matches, f"No column name matching «{name}» found; column names are: {list(table.columns)}"
     return table[matches]
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -39,6 +39,8 @@ LOG = logging.getLogger(__name__)
 
 PROVENANCE_KEY = "_provenance"
 
+RESERVED_COLUMNS = {"sample", "collection", "date"}
+
 
 @cli.group("manifest", help = __doc__)
 def manifest():
@@ -57,7 +59,18 @@ def manifest():
     metavar = "<column>",
     help = "Name of the single column containing sample barcodes.  "
            "Must match exactly; shell-style glob patterns are supported.",
-    required = True)
+    required = False)
+
+@click.option("--collection-column",
+    metavar = "<column>",
+    help = "Name of the single column containing collection barcodes.  "
+           "Must match exactly; shell-style glob patterns are supported.",
+    required = False)
+
+@click.option("--date-column",
+    metavar = "<column>",
+    help = "Name of the single column containing the sample collected date.",
+    required = False)
 
 @click.option("--sample-type",
     metavar = "<type>",
@@ -89,9 +102,15 @@ def parse(**kwargs):
     spreadsheet with at least one sheet in it, identified by name using the required option --sheet.
     Supported URL schemes include http[s]:// and s3://, as well as others.
 
-    The required --sample-column option specifies the name of the column
-    containing the sample barcode.  Other columns may be extracted into the
-    manifest records as desired using the --extra-column option.
+    The --sample-column option specifies the name of the column
+    containing the sample barcode. The --collection-column option specifies
+    the name of the column containing the collection barcode. You must supply one
+    or both of those options.
+
+    The --date-column specifies the name of the column containing the sample collected date.
+
+    Other columns may be extracted into the manifest records as desired using the
+    --extra-column option.
 
     The row-filter entry specifies a pandas query to filter
     (using the python engine) rows in the manifest. Column names refer to columns
@@ -105,7 +124,8 @@ def parse(**kwargs):
         (dst, yaml.safe_load(src))
             for dst, src
             in [arg.split(":", 1) for arg in kwargs["extra_columns"]]
-    ]
+
+       ]
 
     manifest = _parse(**kwargs)
     dump_ndjson(manifest)
@@ -130,6 +150,7 @@ def parse_using_config(config_file):
         workbook: OneDrive/SFS Prospective Samples 2018-2019.xlsx
         sheet: HMC
         sample_column: "Barcode ID*"
+        date_column: "Coll_date"
         extra_columns:
           collection:
             name: "Collection ID*"
@@ -161,7 +182,11 @@ def parse_using_config(config_file):
         ...
 
     The sample_column entry specifies the name of the column
-    containing the sample barcode.
+    containing the sample barcode. The collection_column entry specifies
+    the name of the column containing the collection barcode. You must supply one
+    or both of those entries.
+
+    The date_column specifies the name of the column containing the sample collected date.
 
     The row_filter entry specifies a pandas query to filter
     (using the python engine) rows in the manifest. Column names refer to columns
@@ -213,7 +238,9 @@ def parse_using_config(config_file):
             kwargs = {
                 "workbook": config["workbook"],
                 "sheet": config["sheet"],
-                "sample_column": config["sample_column"],
+                "sample_column": config.get("sample_column"),
+                "collection_column": config.get("collection_column"),
+                "date_column": config.get("date_column"),
                 "extra_columns": list(config.get("extra_columns", {}).items()),
                 "sample_type": config.get("sample_type"),
                 "row_filter" : config.get("row_filter")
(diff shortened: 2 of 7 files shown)


lib/id3c/cli/command/etl/manifest.py
```diff
@@ -55,6 +56,7 @@ def etl_manifest(*, db: DatabaseSession):
     expected_identifier_sets = {
         "samples": {"samples"},
         "collections": {
+            "collections-clia-compliance",
```
jccraft-uw, line 59: We have one record from the incident report manifest that has a barcode from the collections-clia-compliance identifier set. This record was recorded on 5/21/2020. 
It is OK to add collections-clia-compliance to the expected set, or was there some sort of barcode mixup with this one?

joverlee521: This should not include `collections-clia-compliance`, it must be a barcode mixup in the incident report.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -206,9 +227,9 @@ def upsert_sample(db: DatabaseSession,
         LOG.info(f"Updating existing sample {sample.id}")
         sample = db.fetch_row("""
             update warehouse.sample
-               set identifier = %(identifier)s,
-                   collection_identifier = %(collection_identifier)s,
-                   collected = date_or_null(%(collection_date)s),
+               set identifier = coalesce(%(identifier)s, identifier),
+                   collection_identifier = coalesce(%(collection_identifier)s, collection_identifier),
+                   collected = coalesce(date_or_null(%(collection_date)s), collected),
```
jccraft-uw, line 232: See if you agree with the 3 coalesces I added here. The idea is not to wipe out a value with an update that doesn't have the field populated.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: Thanks! I removed it.

lib/id3c/cli/command/etl/manifest.py
joverlee521: @tsibley made a similar change in his PR #161 😄

lib/id3c/cli/command/etl/manifest.py
kairstenfay: It looks like this didn't get removed as expected.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -95,29 +97,40 @@ def etl_manifest(*, db: DatabaseSession):
...
-            # known and from the correct identifier set.
-            sample_barcode = manifest_record.document.pop("sample")
-            sample_identifier = find_identifier(db, sample_barcode)
-
-            if not sample_identifier:
-                LOG.warning(f"Skipping sample with unknown sample barcode «{sample_barcode}»")
+            manifest_type = None
+            try:
+                manifest_type = receiving_manifest.ManifestType(manifest_record.document['manifest_type'])
+            except:
```
kairstenfay, line 102: I would caution against using a general `except` here and narrow it to the Exception you'd expect to be thrown.

lib/id3c/cli/io/adapters.py
```diff
@@ -0,0 +1,50 @@
...
+    sheet_values = sheet_contents_dict.get('values', [])
+
+    # Convert the dictionary returned to json in order to create a
+    # digest representing the data as it was returned by Google.
+    contents_json = json.dumps(sheet_contents_dict)
+    digest = sha1(contents_json.encode('utf-16')).hexdigest()
+
+    # Use values from the first row as column names.
+    # Sometimes users populate columns that have no header/title.
+    # Add unique dummy column names for those.
```
kairstenfay, line 42: This isn't really a code suggestion as much as a documentation suggestion -- 
Recently the Blab dev team has been working toward being more inclusive in the language we choose to write in our programs. @tsibley pointed me to [Google's inclusive documentation writing guide](https://developers.google.com/style/inclusive-documentation) which I've found super informative. Would you consider replacing the word "dummy" here with something else? Thanks!

lib/id3c/cli/io/adapters.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+from googleapiclient import discovery
+from google.oauth2 import service_account
+from typing import Tuple
+from os import environ
+from hashlib import sha1
+import pandas as pd
+import json
+
+GOOGLE_SERVICE_ACCOUNT_CREDS_ENV_VAR = "GOOGLE_SERVICE_ACCOUNT_CREDENTIALS_FILE_PATH"
```
kairstenfay, line 12: Storing this string as a new variable feels a bit redundant .. perhaps we could just read 'GOOGLE_SERVICE_ACCOUNT_CREDENTIALS_FILE_PATH' from the environment?

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: Absolutely. I'm catching specific exceptions now. Thanks for the call out.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: These changes are no longer in the new implementation, so I'm resolving this comment thread.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: The implementation now is quite different from before, so I'm resolving this comment thread. I've caught specific exception types in the new code, but please comment if you see a spot that isn't what you would expect.

jccraft-uw:
> I don't believe id3c should be reading from Google Sheets, but rather that should be handled in the specimen-manifests repo. See [my comment](https://github.com/seattleflu/specimen-manifests/pull/15#issuecomment-670220535) in [seattleflu/specimen-manifests/pull/15](https://github.com/seattleflu/specimen-manifests/pull/15) for more context.

Thanks for raising the question. This led us to discuss this project as a group and walk through where the pieces of code live. I'm going to resolve this comment thread.

lib/id3c/cli/io/adapters.py
jccraft-uw: Thanks a lot for pointing me to this doc, Kairsten. The new implementation doesn't include this code, so I'm resolving this thread.

lib/id3c/cli/io/adapters.py
jccraft-uw: This was a super valid point. Now that Thomas showed me how the Google API can work by setting one env var we don't need these all over the place, so that is gone from the code.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -204,17 +224,24 @@ def upsert_sample(db: DatabaseSession,
...
+        # Only update the identifier or collection_identifier columns if the source data provides a sample barcode
+        if record_has_sample:
+            update_statement += ", identifier = %(identifier)s "
+            if record_has_collection:
+                update_statement += ", collection_identifier = %(collection_identifier)s "
+
+        if record_has_collected_date:
+            update_statement += ", collected = date_or_null(%(collected_date)s) "
+
+        update_statement += "where sample_id = %(sample_id)s returning sample_id as id, identifier, collection_identifier, encounter_id "
```
joverlee521, line 242: I know this is written like this to avoid repeating code, but this seems more convoluted and confusing than just writing out separate SQL queries

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -95,55 +95,72 @@ def etl_manifest(*, db: DatabaseSession):
...
+            record_has_sample = "sample" in manifest_record.document
+            record_has_collection = "collection" in manifest_record.document
+            record_has_collected_date = "collected_date" in manifest_record.document
 
-            if not sample_identifier:
-                LOG.warning(f"Skipping sample with unknown sample barcode «{sample_barcode}»")
-                mark_skipped(db, manifest_record.id)
-                continue
+             # Sample collection date
+            collected_date = manifest_record.document.pop("collected_date", None)
```
joverlee521, line 104: As discussed on Slack, I think we should make this backwards compatible so that we can re-process the existing records in `receiving.manifest` if needed.

lib/id3c/cli/command/manifest.py
```diff
@@ -84,8 +85,8 @@ def parse(**kwargs):
     """
     Parse a single manifest workbook sheet.
 
-    <manifest.xlsx> must be a path or URL to an Excel workbook with at least
-    one sheet in it, identified by name using the required option --sheet.
+    <filepath> must be a path or URL to an Excel workbook or Google Sheeets
```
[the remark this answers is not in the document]
jccraft-uw: Fixed!

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    mime_type = GOOGLE_FORMAT_MIME_MAP[destination_format]
+
+    drive_service = discovery.build('drive', 'v2')
+    request = drive_service.files().export_media(fileId=document_id, mimeType=mime_type)
+    buffer = BytesIO()
+    downloader = MediaIoBaseDownload(buffer, request)
+    done = False
+    while done is False:
+        done = downloader.next_chunk()
```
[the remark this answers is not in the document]
jccraft-uw: That would have been a nasty bug to track down. Large documents might not have been fully downloaded. Thanks for seeing that. Fixed.

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Move the buffer to the beginning so the consumer can call read()
+    buffer.seek(0)
+    return buffer
+
+def extract_google_document_id_from_url(url: str) -> str:
```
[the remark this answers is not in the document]
jccraft-uw: Fixed.

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    Determines whether the input url is to a Google Documents resource and if so
+    extracts the document ID.
+    """
+    document_id = None
+
+    google_sheets_pattern = re.compile(".*//docs.google.com/spreadsheets/d/(.*)/?", flags=re.IGNORECASE)
```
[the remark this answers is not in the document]
jccraft-uw: I updated the function name slightly to extract_google_document_id_from_docs_url, and I made the matching work for Google documents under docs.google.com. I tested for Sheets, Docs, and Slides.
I also used that regex that you pointed out.

lib/id3c/cli/command/manifest.py
```diff
@@ -222,19 +259,33 @@ def parse_using_config(config_file):
...
            extra_columns: List[Tuple[str, Union[str, dict]]] = [],
            sample_type = None,
            row_filter = None):
     """
     Internal function powering :func:`parse` and :func:`parse_using_config`.
     """
+
+    # Determine if the workbook URL is for a Google Document and if so
+    # retrieve it as a Google Sheets spreadsheet. Otherwise, retrieve it
+    # using urlopen.
```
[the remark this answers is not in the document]
jccraft-uw: Good point. I moved this to the previous commit.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -153,9 +170,12 @@ def etl_manifest(*, db: DatabaseSession):
 
 
 def upsert_sample(db: DatabaseSession,
+                  record_has_sample: bool,
+                  record_has_collection: bool,
+                  record_has_collected_date: bool,
```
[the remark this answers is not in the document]
jccraft-uw: There is an important distinction I'm making here. I want to make sure this is getting across.

These parameters are passing in the values pulled out of the record. They may be real values or None.
identifier: Optional[str],
collection_identifier: Optional[str],
collected_date: Optional[str],
additional_details: dict) -> Tuple[Any, str]:

These 3 new parameters are saying whether the key existed in the record at all.
record_has_sample: bool,
record_has_collection: bool,
record_has_collected_date: bool,

From looking only at collection_identifier, for example, you can't tell whether the manifest settings tried to extract collection_identifier from the record and that record happened not to have a value for it or if the manifest wasn't configured to try to pull out collection_identifier. 

The SQL logic is driven by whether the keys existed in the record, not by whether the columns had a non-None value for the keys.

Does that explain why the new parameters are needed?
If there is a  tri-state data type that would combine these into one sec of parameters let me know. I could imagine the three states as Not Set (the manifest didn't even try), None, and a real value.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: I don't think it's especially convoluted to create SQL command text dynamically. If I were to hardcode the SQL for the various scenarios I'd be repeating a good chunk of code 5 times.
sample + no collection + no collected date
sample + collection + no collected date
sample + collection + collected date
no sample + no collected date
no sample + collected date

lib/id3c/cli/command/etl/manifest.py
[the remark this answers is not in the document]
joverlee521: I think the SQL logic _should_ be driven by a non-None value. I don't think we want to update any existing value to `None` especially once we also ingest collected date from REDCap.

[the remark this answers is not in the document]
jccraft-uw: I was on the same train of thought with you with the previous implementation when I had the coalesce's in there.

If I made the update dependent on whether the record had a non-None value for the field this would be a change from current behavior. If an aliquot record came in with collection ABC and sample 123 and then later they updated the sheet removing the sample ID, after this change we would keep the collection mapped to the sample.  

Is that what we want?

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: Thanks for the good discussion. I've changed the setting to "date_column" and the JSON key to "date" to be compatible with how things are done today.

jccraft-uw:
[Ingest the lab's incident report spreadsheet to get data for bad samples](https://trello.com/c/McH4xpP9/695-ingest-the-labs-incident-report-spreadsheet-to-get-data-for-bad-samples)

jccraft-uw:
[Ingest the lab's incident report spreadsheet to get data for bad samples](https://trello.com/c/McH4xpP9/695-ingest-the-labs-incident-report-spreadsheet-to-get-data-for-bad-samples)

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,75 @@
...
+    downloader = MediaIoBaseDownload(buffer, request)
+    done = False
+    while done is False:
+        status, done = downloader.next_chunk()
+
+    # Move the buffer to the beginning so the consumer can call read()
+    buffer.seek(0)
+    return buffer
+
+def extract_google_document_id_from_docs_url(url: str) -> Optional[str]:
```
kairstenfay, line 61: While precise, this function's name seems a tad verbose. Notably, it says both "document" and "docs" which is a bit redundant. May I suggest one of the following?
* extract_id_from_google_docs_url
* extract_google_id_from_docs_url
* extract_google_document_id_from_url
* google_document_id / google_docs_id

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,75 @@
...
+def export_file_from_google_drive(document_id: str, destination_format: GoogleDriveExportFormat) -> BytesIO:
+    """
+    From Google Drive export the specified document in the specified MIME format.
+    Requires that the GOOGLE_APPLICATION_CREDENTIALS environment variable be set
+    pointing to the service account credentials file and that the service account have
+    read access to the document being exported.
+
+    Not every type of document can be exported as every GoogleDriveExportFormat type.
+
+    If you export a Google Sheets spreadsheet as a CSV, the first tab will be exported.
```
kairstenfay, line 45: What does "the first tab will be exported" mean here?

lib/id3c/cli/command/etl/manifest.py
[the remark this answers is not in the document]
kairstenfay: I agree with @joverlee521 re: creating the boolean flags within `upsert_sample` rather than passing them in as additional parameters. In my opinion, it's better practice to include fewer parameters and equip functions with the knowledge to parse a document we pass it. We try to employ this in other areas of id3c/id3c-customizations as well.

lib/id3c/cli/command/etl/manifest.py
kairstenfay: Instead of doing string concatenation, could we do a string substitution, coupling @joverlee521 and @jccraft-uw 's ideas?

e.g.
```
record_has_sample = ", identifier = %(identifier)s " if "sample" in manifest_record.document else ""
...
query_string = f"""
   update warehouse.sample set details = coalesce(details, '{}') || %(additional_details)s 
   {record_has_sample}
   ...
"""
```
and so on for the other boolean flags

lib/id3c/cli/command/manifest.py
```diff
@@ -483,3 +542,12 @@ def load_ndjson(file: Iterable[str]) -> Iterable:
     Load newline-delimited JSON records from *file*.
     """
     yield from (json.loads(line) for line in file)
+
+
+def find_violating_column_names(extra_columns: List[Tuple[str, Union[str, dict]]]):
+    """
+    Find any reserved column names that also appeaer in extra_columns.
+    """
+    reserved_columns_list = []
```
kairstenfay, line 551: the `reserved_columns_list` variable is immediately overwritten in the following line, making this line unnecessary.

lib/id3c/cli/command/manifest.py
```diff
@@ -483,3 +542,12 @@ def load_ndjson(file: Iterable[str]) -> Iterable:
...
+    reserved_columns_list = [x for x in extra_columns if str(x[0]).lower() in RESERVED_COLUMNS]
```
kairstenfay, line 552: you can just return this list comprehension instead of storing it in a new variable

lib/id3c/cli/command/manifest.py
```diff
@@ -295,9 +348,15 @@ def _parse(*,
...
-    # stripping.
-    parsed_manifest = parsed_manifest.dropna(subset = ["sample"])
+    # Drop rows that have no data for the sample_column and/or the collection_column, depending
+    # on which columns are configured. If both sample_column and collection_column are configured,
+    # drop rows if both columns don't have data.
+    if sample_column and collection_column:
+        parsed_manifest = parsed_manifest.dropna(subset = ["sample", "collection"], how='all')
+    elif sample_column:
+        parsed_manifest = parsed_manifest.dropna(subset = ["sample"])
+    else:
```
kairstenfay, line 358: having neither `sample_column` nor `collection_column` would fall into this edge case. I think it's better to make this `elif collection_column:` here to guard against future changes we may make.

lib/id3c/cli/command/manifest.py
```diff
@@ -483,3 +542,12 @@ def load_ndjson(file: Iterable[str]) -> Iterable:
     Load newline-delimited JSON records from *file*.
     """
     yield from (json.loads(line) for line in file)
+
+
+def find_violating_column_names(extra_columns: List[Tuple[str, Union[str, dict]]]):
```
kairstenfay, line 547: the term "violating" here is a bit vague to me. Could we rename it to something like "protected" or "reserved" to match the global variable? maybe something like "ensure no reserved column names", etc.

kairstenfay:
@jccraft-uw I just pushed a commit that adds a couple of doctests to the google io functionality.

lib/id3c/cli/command/manifest.py
jccraft-uw: Removed!

lib/id3c/cli/command/manifest.py
jccraft-uw: Changed.

lib/id3c/cli/command/manifest.py
jccraft-uw: Good point. I've made that change.

lib/id3c/cli/command/manifest.py
jccraft-uw: I went with find_reserved_columns_used_in_extra_column_list

lib/id3c/cli/command/etl/manifest.py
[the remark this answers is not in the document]
jccraft-uw: I've gone with coalescing the identifier and collection_identifier columns. Thomas recently added in the coalesce in the update for collected. This obviates those flag parameters and the SQL command string building. FYI @joverlee521

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: I've gone with coalescing the identifier and collection_identifier columns. Thomas recently added in the coalesce in the update for collected. This obviates those flag parameters and the SQL command string building.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -206,8 +215,13 @@ def upsert_sample(db: DatabaseSession,
         LOG.info(f"Updating existing sample {sample.id}")
         sample = db.fetch_row("""
             update warehouse.sample
+<<<<<<< HEAD
                set identifier = %(identifier)s,
                    collection_identifier = %(collection_identifier)s,
+=======
+               set identifier = coalesce(%(identifier)s, identifier),
+                   collection_identifier = coalesce(%(collection_identifier)s, collection_identifier),
+>>>>>>> 8033726... Make manifest parsing support multiple source types
```
kairstenfay, line 224: Looks like a git merge conflict slipped into this commit

lib/id3c/cli/command/manifest.py
kairstenfay: Thanks! Do you think we should throw an error in an `else:` statement that follows those conditionals? (I'm not sure)

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: GRRR. Fixed. I saw it. But thanks.

lib/id3c/cli/command/manifest.py
```diff
@@ -208,7 +229,8 @@ def _parse(*,
            sheet,
            sample_column,
            extra_columns: List[Tuple[str, Union[str, dict]]] = [],
-           sample_type = None):
+           sample_type = None,
+           row_filter = None):
```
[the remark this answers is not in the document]
jccraft-uw: Added that.

lib/id3c/cli/io/google.py
jccraft-uw: I went with extract_document_id_from_google_url

lib/id3c/cli/io/google.py
jccraft-uw: I updated the comment to
If you export a multi tabbed Google Sheets spreadsheet as a CSV, it's the first tab that will be exported.

jccraft-uw:
> Thanks so much for implementing our first round of feedback, @jccraft-uw . This is looking a lot better!
> 
> I've left a few comments, mostly minor suggestions.
> One thing that doesn't fit anywhere else --
> I'm concerned about how big the `etl_manifest` function has grown in id3c/cli/command/etl/manifest.py. I wonder if we can break apart some of that functionality into smaller functions.

I looked at splitting functionality out of this function into separate functions. It would be validation that would make sense to split out, and it turns out that there aren't that many places where the validation isn't just one line of code. I opted to leave that in the function itself.

lib/id3c/cli/command/manifest.py
jccraft-uw: I don't think we need to throw an error here if neither sample_column nor collection_column is populated because the parse and parse_using_config functions both check that at least one of sample_column and collection_column is set.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -95,55 +95,71 @@ def etl_manifest(*, db: DatabaseSession):
...
+            # 'collection' keys.
+            should_update_identifiers = "sample" in manifest_record.document \
+                and "collection" in manifest_record.document
 
-            if not sample_identifier:
-                LOG.warning(f"Skipping sample with unknown sample barcode «{sample_barcode}»")
-                mark_skipped(db, manifest_record.id)
-                continue
+             # Sample collection date
+            collected_date = manifest_record.document.pop("date", None)
```
joverlee521, line 105: I think we still want to keep `date` in the `sample.details` so that we can easily compare the `collected` column and the aliquot manifest date ingested into `details.date`. There are still existing differences that we need to investigate (see [Trello card](https://trello.com/c/XFOlT3JC/738-investigate-the-discordance-between-samplecollected-and-details%E2%86%92date))

kairstenfay: I agree with Jover here.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -95,55 +95,71 @@ def etl_manifest(*, db: DatabaseSession):
...
-            # identifier set.
+            # Attempt to find barcodes and their related identifiers
+            sample_barcode = manifest_record.document.pop("sample", None)
+            sample_identifier = find_identifier(db, sample_barcode) if sample_barcode else None
             collection_barcode = manifest_record.document.pop("collection", None)
             collection_identifier = find_identifier(db, collection_barcode) if collection_barcode else None
 
+            # Skip a record if it has no sample_identifier and no collection_identifier
+            if not sample_identifier and not collection_identifier:
+                LOG.warning(f"Skipping record «{manifest_record.id}» because it has no known sample barcode or collection barcode")
```
kairstenfay, line 115: Can we use the word "identifier" in this message instead of "barcode"?

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -95,55 +95,71 @@ def etl_manifest(*, db: DatabaseSession):
...
+                mark_skipped(db, manifest_record.id)
+                continue
+
+            # Skip a record if it has a sample barcode but the barcode doesn't match an identifer
```
kairstenfay, line 119: small typo in "identifer"

lib/id3c/cli/command/etl/manifest.py
kairstenfay: Also, can we change the language in this warning message to something slightly clearer like
> because it has neither a known sample identifier nor a collection identifier

?
The current wording suggests a sample could be missing one of the two identifiers and enter this block.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -95,55 +95,71 @@ def etl_manifest(*, db: DatabaseSession):
...
+                    assert sample_identifier.set_name in expected_identifier_sets["samples"], \
+                        (f"Sample identifier found in set «{sample_identifier.set_name}», " +
+                        f"not {expected_identifier_sets['samples']}")
+
+            # Skip a record if the collection identifier is from an unexpected set
+            if collection_identifier and collection_identifier.set_name not in expected_identifier_sets["collections"]:
+                LOG.warning(f"Skipping sample because collection identifier found in set «{collection_identifier.set_name}», not \
+                    {expected_identifier_sets['collections']}")
+                mark_skipped(db, manifest_record.id)
+                continue
```
kairstenfay, line 148: I feel like this barcode skipping block fits better with the code where we arre "Skipping sample with unknown collection barcode «{collection_barcode}»." If it makes sense, could you please move it up there?

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -95,55 +95,71 @@ def etl_manifest(*, db: DatabaseSession):
...
-            # Sample collection date
-            collection_date = manifest_record.document.get("date")
 
             # Upsert sample cooperatively with enrollments ETL routine
             #
-            # The details document was intentionally modified by two pop()s
-            # earlier to remove barcodes that were looked up.  The
-            # rationale is that we want just one clear place in the
+            # The details document was intentionally modified by three pop()s
+            # earlier to remove barcodes that were looked up and the date.
```
kairstenfay, line 154: This language will need to change once you stop the date pop()

lib/id3c/cli/command/manifest.py
```diff
@@ -222,13 +257,23 @@ def parse_using_config(config_file):
             LOG.error(f"Required key «{key}» missing from config {config}")
             raise key from None
 
+        if config.get("sample_column") is None and config.get("collection_column") is None:
+            raise KeyError("The configuration must indicate the sample column, the collection \
+                column, or both.")
```
kairstenfay, line 262: I wonder if there's an edge case where a user may accidentally specify an empty string for sample_column and collection_column, in which case we would not raise an error

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: Sounds good. This will duplicate that data, but it's handy for backwards compatibility.
Fixed.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: I stole your sentence! Fixed.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: Fixed x 2.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: I put the collection check above the sample check.

lib/id3c/cli/command/etl/manifest.py
jccraft-uw: Fixed!

lib/id3c/cli/command/manifest.py
jccraft-uw: Since the YAML reader returns a string I was able to use `not`. Good call out. That was a simple enough change to cover both no entry and an empty string entry.

lib/id3c/cli/command/manifest.py
```diff
@@ -73,26 +87,53 @@ def manifest():
...
+    in the manifest itself.
+    Example: `corrective action`.notnull() and `corrective action`.str.lower().str.startswith("discard")
 
     Manifest records are output to stdout as newline-delimited JSON records.
     You will likely want to redirect stdout to a file.
     """
+    # The --sample_column and --collection_column options are not required, so
+    # check here that at least one of them is provided.
+    if kwargs.get("sample_column") is None and kwargs.get("collection_column") is None:
+        raise ValueError("You must indicate the sample column, the collection column, or both.")
```
tsibley, line 126: Raising a `click.UsageError` instead of `ValueError` here will avoid printing a stacktrace and instead print a nicer message on the command-line consistent with Click's other automatic usage checking. See an example of this in another part of ID3C:

https://github.com/seattleflu/id3c/blob/4a090fe5255205f2668f655c77f35b3ce8bfa63c/lib/id3c/cli/command/redcap_det.py#L101-L102

lib/id3c/cli/command/manifest.py
```diff
@@ -192,30 +246,54 @@ def parse_using_config(config_file):
...
             raise key from None
 
+        if not config.get("sample_column") and not config.get("collection_column"):
+            raise KeyError("The configuration must indicate the sample column, the collection "
+                + "column, or both.")
+
+        reserved_columns_in_extra_columns = find_reserved_columns_used_in_extra_column_list(kwargs["extra_columns"])
+        assert len(reserved_columns_in_extra_columns) == 0, \
+            f"A reserved column name has been configured in the extra_columns setting: {reserved_columns_in_extra_columns}"
+
```
tsibley, line 267: I think it'd be better to put both these two checks and their equivalents in `parse()` above into `_parse()` below instead. Or maybe you considered this but rejected it?

lib/id3c/cli/command/manifest.py
```diff
@@ -445,3 +542,10 @@ def load_ndjson(file: Iterable[str]) -> Iterable:
     Load newline-delimited JSON records from *file*.
     """
     yield from (json.loads(line) for line in file)
+
+
+def find_reserved_columns_used_in_extra_column_list(extra_columns: List[Tuple[str, Union[str, dict]]]):
+    """
+    Find any reserved column names that also appeaer in extra_columns.
+    """
+    return [x for x in extra_columns if str(x[0]).lower() in RESERVED_COLUMNS]
```
tsibley, line 551: This function name feels pretty verbose and specific to me, particularly for an internal function which can benefit from the caller's context being in the same file.

I wonder if this is better expressed in set logic:

```
disallowed_extra_columns = {dst for dst, src in extra_columns} & RESERVED_COLUMNS
```

Is the string casting so the config can do `42: …` instead of `"42": …` in `extra_columns`?

Why the `lower()`? We're not doing case-insensitive matching on this later down the line, are we?

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,81 @@
...
+
+    >>> extract_document_id_from_google_url('s3://my-s3-bucket/filepath.xlsx')
+    """
+    document_id = None
+
+    google_docs_pattern = re.compile(".*//docs.google.com/.*/d/([a-zA-Z0-9-_]+)/?", flags=re.IGNORECASE)
+    google_docs_matches = re.match(google_docs_pattern, url)
+
+    if google_docs_matches and google_docs_matches.lastindex == 1:
+        document_id = google_docs_matches[1]
```
tsibley, line 79: FYI, the regex itself guarantees that if it matches it'll have a match of index 1. No need to check `lastindex`.

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,81 @@
...
+
+    >>> extract_document_id_from_google_url('https://docs.google.com/spreadsheets/d/1BxiMVs0XRA5nFMdKvBdBZjgmUUqptlbs74OgvE2upms/edit#gid=0')
+    '1BxiMVs0XRA5nFMdKvBdBZjgmUUqptlbs74OgvE2upms'
+
+    >>> extract_document_id_from_google_url('s3://my-s3-bucket/filepath.xlsx')
+    """
+    document_id = None
+
+    google_docs_pattern = re.compile(".*//docs.google.com/.*/d/([a-zA-Z0-9-_]+)/?", flags=re.IGNORECASE)
+    google_docs_matches = re.match(google_docs_pattern, url)
```
tsibley, line 76: This is a relatively loose pattern. It'll match things like:

```
https://web.archive.org/web/*//docs.google.com/spreadsheet/d/…
```

While I think that example is unlikely to happen, it's good practice to make regexes as precise as possible up front. I think a combination of `urllib.parse.urlparse()` and a regex would be best here. By way of example, maybe something like:

```python
    url = urlparse(url)

    if url.hostname != "docs.google.com":
        return None

    match = re.search(r"^/[^/]+/d/(?P<document_id>[a-zA-Z0-9-_]+)", url.path, flags = re.IGNORECASE)

    return match["document_id"] if match else None
```

`urlparse()` means the regex doesn't have to deal with the vagaries of URL protocol/schemes and origins (host + port). The change from `.*` to `[^/]+` means that extra path parts aren't slurped up.

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,81 @@
...
+GOOGLE_FORMAT_MIME_MAP = {
+    GoogleDriveExportFormat.CSV: "text/csv",
+    GoogleDriveExportFormat.EXCEL: "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+    GoogleDriveExportFormat.HTML: "text/html",
+    GoogleDriveExportFormat.HTML_ZIPPED: "application/zip",
+    GoogleDriveExportFormat.PDF: "application/pdf",
+    GoogleDriveExportFormat.PLAIN_TEXT: "text/plain",
+    GoogleDriveExportFormat.RICH_TEXT: "application/rtf",
+    GoogleDriveExportFormat.WORD: "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
+}
```
tsibley, line 34: Ok to leave them since they're here now, but as a general comment, it does seem to me like everything but the one we're using (Excel, and maybe CSV) are a YAGNI situation.

lib/id3c/cli/io/google.py
```diff
@@ -0,0 +1,81 @@
...
+def extract_document_id_from_google_url(url: str) -> Optional[str]:
+    """
+    Determines whether the input url is to a Google Documents resource at
+    docs.google.com and if so extracts the document ID. Document types include
+    Google Docs, Google Sheets, and Google Slides.
+
+    >>> extract_document_id_from_google_url('https://docs.google.com/spreadsheets/d/1BxiMVs0XRA5nFMdKvBdBZjgmUUqptlbs74OgvE2upms/edit#gid=0')
+    '1BxiMVs0XRA5nFMdKvBdBZjgmUUqptlbs74OgvE2upms'
+
+    >>> extract_document_id_from_google_url('s3://my-s3-bucket/filepath.xlsx')
```
tsibley, line 71: Really great to have doctests for this sort of function!

tsibley (changes requested):

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -204,17 +225,23 @@ def upsert_sample(db: DatabaseSession,
...
-                   details = coalesce(details, '{}') || %(additional_details)s
+                set {identifiers_update}
+                    collected = coalesce(date_or_null(%(collection_date)s), collected),
+                    details = coalesce(details, {empty_json}) || %(additional_details)s
 
              where sample_id = %(sample_id)s
 
             returning sample_id as id, identifier, collection_identifier, encounter_id
-            """,
+            """.format(identifiers_update = identifiers_update, empty_json = "'{}'"),
```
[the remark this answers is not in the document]
jccraft-uw: This was a good opportunity for me to learn about psycopg2's SQL composition. In this case, it may have been a bit overboard because in here we're not dynamically determining which columns are getting updated .... it's either "include identifier and collection_identifier" or don't. So, the SQL is variable, but not super variable.
I create the first part as its own composable and use it in the main command.

lib/id3c/cli/command/manifest.py
jccraft-uw: That makes good sense. I ended up moving this check into _parse as you suggested below, so this went away.

lib/id3c/cli/command/manifest.py
jccraft-uw: I had the same checks in both "outer" functions so that I could give a tailored error message for the sample_column and collection_column part. Every time I looked at this I thought "Yeah, but you should also do the checks in _parse to be safe". Doing it in all three places felt overboard.
I moved these two checks into _parse. That's the right place for them to be.

lib/id3c/cli/command/manifest.py
jccraft-uw: Python sets functionality is really cool. Thinking of this as the intersection of two sets is so great. I've made the switch.
Also, we don't to lower() anywhere else in here. That was me thinking of case insensitive SQL instead of JSON.

lib/id3c/cli/io/google.py
jccraft-uw: @kairstenfay added those!

lib/id3c/cli/io/google.py
jccraft-uw: Good point!

lib/id3c/cli/io/google.py
jccraft-uw: Your feedback makes sense. I've used the pattern you came up with. I like using urlparse to deal with the URL part.

lib/id3c/cli/io/google.py
jccraft-uw: I thought that if this is a general utility it would be nice to add the common formats.

joverlee521 (approved):

kairstenfay (approved):

PR #161, opened 2020-08-05
Ingest sample.collected date from FHIR Specimen collection

tsibley:
See commit messages for details on the functionality.

In particular, I'd like feedback on the update-if-not-null behaviour that both the manifest and FHIR ETLs now do for `warehouse.sample.collected`. I think it makes sense, but it does introduce an uncertainty about where the date comes, as for any given record it now depends on what ETL received data and processed it most recently. They _should_ both agree, but of course in practice there will probably be times when they don't.

### DIFF
--- a/lib/id3c/cli/command/etl/manifest.py
+++ b/lib/id3c/cli/command/etl/manifest.py
@@ -208,7 +208,7 @@ def upsert_sample(db: DatabaseSession,
             update warehouse.sample
                set identifier = %(identifier)s,
                    collection_identifier = %(collection_identifier)s,
-                   collected = date_or_null(%(collection_date)s),
+                   collected = coalesce(date_or_null(%(collection_date)s), collected),
                    details = coalesce(details, '{}') || %(additional_details)s
 
              where sample_id = %(sample_id)s
--- a/lib/id3c/utils.py
+++ b/lib/id3c/utils.py
@@ -1,6 +1,8 @@
 """
 Utilities.
 """
+from typing import Any, Sequence, Union
+
 
 def format_doc(**kwargs):
     """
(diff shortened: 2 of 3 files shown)


joverlee521:
> I'd like feedback on the update-if-not-null behaviour that both the manifest and FHIR ETLs now do for warehouse.sample.collected

I think the update-if-not-null behavior here makes sense.

> it does introduce an uncertainty about where the date comes, as for any given record it now depends on what ETL received data and processed it most recently

In general, the `collected` field will probably reflect the FHIR collection date since we receive so many DETs from REDCap. The manifest data is usually a one-time upsert because it doesn't get updated as much. 

> They should both agree, but of course in practice there will probably be times when they don't.

Do we want some kind of notification if `sample.collected != sample.details.date`? It could just be a Metabase Pulse at first so we get a general sense of how often they _don't_ agree.

lib/id3c/utils.py
```diff
@@ -14,3 +16,46 @@ def wrap(function):
...
+    >>> getattrpath(obj, "a.b.x")
+    >>> getattrpath(obj, "a.x")
+    >>> getattrpath(obj, "x")
+    >>> getattrpath(obj, ("a", "b", "c"))
+    42
+    """
+    if isinstance(attrpath, str):
+        attrpath = attrpath.split(".")
+
+    assert len(attrpath) != 0
```
kairstenfay, line 53: Could you also say `assert attrpath` here?

This is less of a suggestion and more of my wondering if I'm not considering any edge cases.

kairstenfay:
> > They should both agree, but of course in practice there will probably be times when they don't.
> 
> Do we want some kind of notification if `sample.collected != sample.details.date`? It could just be a Metabase Pulse at first so we get a general sense of how often they _don't_ agree.

I think a Metabase query or pulse is a wise decision for us to discover how discordant the fields may be.

lib/id3c/cli/command/etl/manifest.py
```diff
@@ -208,7 +208,7 @@ def upsert_sample(db: DatabaseSession,
             update warehouse.sample
                set identifier = %(identifier)s,
                    collection_identifier = %(collection_identifier)s,
-                   collected = date_or_null(%(collection_date)s),
+                   collected = coalesce(date_or_null(%(collection_date)s), collected),
```
jccraft-uw, line 211: I'm doing exactly the same thing in PR https://github.com/seattleflu/id3c/pull/160 because the incident report records won't give us collected date.

lib/id3c/utils.py
```diff
@@ -14,3 +16,46 @@ def wrap(function):
...
+    >>> obj.a.b.c = 42
+    >>> getattrpath(obj, "a")
+    {'b': {'c': 42}}
+    >>> getattrpath(obj, "a.b.c")
+    42
+    >>> getattrpath(obj, "a.b.x")
+    >>> getattrpath(obj, "a.x")
+    >>> getattrpath(obj, "x")
+    >>> getattrpath(obj, ("a", "b", "c"))
+    42
```
jccraft-uw, line 48: Those would make great unit tests. Do we have unit tests for utility functions like these?

joverlee521: These are part of the [doctests](https://docs.python.org/3/library/doctest.html) that get run with pytest.

jccraft-uw: Ah, nice. Thanks for the link.

lib/id3c/utils.py
tsibley: `assert attrpath` would do the same thing, but explicitly checking length felt more descriptive to me in this case.

lib/id3c/utils.py
```diff
@@ -14,3 +16,46 @@ def wrap(function):
...
+    if isinstance(attrpath, str):
+        attrpath = attrpath.split(".")
+
+    assert len(attrpath) != 0
+
+    for attr in attrpath:
+        value = getattr(value, attr, None)
+
+        if value is None:
+            return None
```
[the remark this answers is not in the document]
tsibley: My first version used recursion, but I then switched it to iteration. :-)

I considered and prototyped distinguishing between "present but None" and "not present", but we don't need to make that distinction currently so I punted on the additional complexity. Note that in this case at least one of `collectedDateTime` or `collectedPeriod` will be `None` since per-spec they're mutually exclusive.

tsibley:
> Do we want some kind of notification if `sample.collected != sample.details.date`? It could just be a Metabase Pulse at first so we get a general sense of how often they _don't_ agree.

Yeah, this is a reasonable QC query to run. We could definitely make more use of such retrospective QC. I think we'll want to limit it to samples created in ID3C after this and https://github.com/seattleflu/id3c-customizations/pull/126 is deployed.

joverlee521:
[Ingest kit_reg_date for collection_date on lab ror pdf](https://trello.com/c/Xp2jXdtm/701-ingest-kitregdate-for-collectiondate-on-lab-ror-pdf)

PR #162, opened 2020-08-12
Add decade age bins

joverlee521:
Decade age bins to be used in the views for Formative APIs. 
It seems like a common enough need to be included in ID3C core.


trvrb:
Decade age bins have become pretty standard in COVID-19 dashboards / analysis. I think good to have.

joverlee521:
> why not add this inside the `age-bin.sql` file?

To me, it didn't make sense to re-work the existing `age-bin.sql` since these are tables not views.

PR #163, opened 2020-08-31
Support repeating REDCap

joverlee521:
See details in commit messages.

Needs to be deployed with changes in ID3C-customizations (https://github.com/seattleflu/id3c-customizations/pull/134)

### DIFF
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -7,10 +7,11 @@ REDCap.
 import os
 import click
 import logging
+from collections import defaultdict
 from datetime import datetime, timezone
 from functools import wraps
 from textwrap import dedent
-from typing import Callable, Iterable, Optional, Tuple, Dict, List, Any
+from typing import Callable, Iterable, Optional, Tuple, Dict, List, Any, DefaultDict
 from urllib.parse import urljoin
 from id3c.cli.command import with_database_session
 from id3c.cli.redcap import is_complete, Project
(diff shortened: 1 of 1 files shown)


joverlee521:
[See what RedCAP records look like when a repeating instrument is used multiple times. How would we ingest that?](https://trello.com/c/z7eBYmud/743-see-what-redcap-records-look-like-when-a-repeating-instrument-is-used-multiple-times-how-would-we-ingest-that)

kairstenfay:
> This looks really solid, Jover.
> 
> However, when I run `id3c redcap-det generate`, I get the following error:
> 
> ```
> TypeError: __init__() got an unexpected keyword argument 'required_instruments'
> ```

Hmm I think I posted this in the wrong PR. I'm going to post in the related id3c-customizations one.

tsibley (approved):
I was confused by this change in the second commit until I saw the third commit:

```diff
@@ -208,7 +201,7 @@ def command_for_project(name: str,                                                                                                                                         
                             instrument
                                 for instrument
                                 in required_instruments
-                                if not is_complete(instrument, redcap_record)                                                                                                                 
+                                if not is_complete(instrument, redcap_record_instances)                                                                                                       
                         }

                         if incomplete_instruments:
```

I expect that the modified version wouldn't work (right?), so it's a little odd to me to have the commits split up.

joverlee521:
> I expect that the modified version wouldn't work (right?), so it's a little odd to me to have the commits split up.

Yeah, it wouldn't work. I can definitely squash the two commits together.

tsibley (approved):

PR #164, opened 2020-09-10
Add barcode layout for season 3 househould arm

jccraft-uw:
This is a PR for a barcode layout to go with a new identifier set.
The identifier set has already been created. It's called collections-household-general-season-three (I put season three in the name to help [me] distinguish it from the 4 other household collection set identifiers.)

The request in Slack from Roni was:
"is it possible to get 115 sheets of orange Household barcodes? This is for the upcoming, general Household year 3 study."
"these are for this upcoming year's household study, which is no longer 'household observation' but just the household study. We only need single barcodes for this (printed just once) thanks!"

### DIFF
--- a/lib/id3c/labelmaker.py
+++ b/lib/id3c/labelmaker.py
@@ -226,6 +226,12 @@ class SamplesHaarviLayout(LabelLayout):
     copies_per_barcode = 1
     reference = "HAARVI"
 
+class CollectionsHouseholdGeneralLayout(LabelLayout):
+    sku = "LCRY-1100"
+    barcode_type = "HH GENERAL"
+    copies_per_barcode = 1
+    reference = "seattleflu.org"
+
 
 LAYOUTS = {
     "samples": SamplesLayout,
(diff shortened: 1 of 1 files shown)


lib/id3c/labelmaker.py
```diff
@@ -240,6 +246,7 @@ class SamplesHaarviLayout(LabelLayout):
     "collections-household-observation-asymptomatic": CollectionsHouseholdObservationAsymptomaticLayout,
     "collections-household-intervention": CollectionsHouseholdInterventionLayout,
     "collections-household-intervention-asymptomatic": CollectionsHouseholdInterventionAsymptomaticLayout,
+    "collections-household-general-season-three": CollectionsHouseholdGeneralSeasonThreeLayout,
```
joverlee521, line 249: The name of the identifier set and the class are a mouthful 😆 
Maybe we can get remove season three from the name?

jccraft-uw: I added season three in there explicitly because we now have 5 household related identifier sets. I tried to make this easier on myself when I have to find the right identifier set next time. OK?

kairstenfay: I agree with Jover here. I don't think it's necessary to qualify a barcode set with a year. IIUC, the only reason we're adding a new identifier set is because they're dropping the observation vs. intervention distinction. Therefore, `collections-household-general` should be a sufficient distinction.

kairstenfay: Or even, `collections-household`.

jccraft-uw: OK. I lost that battle.

tsibley (approved):

PR #165, opened 2020-09-15
redcap: Support for a conditional logic filter when fetching records

tsibley:
Helpful for reducing the number of fetched records when the record ids
aren't known.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -143,6 +143,7 @@ class Project:
                 ids: List[str] = None,
                 fields: List[str] = None,
                 events: List[str] = None,
+                filter: str = None,
                 raw: bool = False) -> List['Record']:
         """
         Fetch records for this REDCap project.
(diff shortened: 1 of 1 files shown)


kairstenfay (approved):
Good call on adding this.

My only suggestion is to provide an example of the *filter* syntax because IMO the REDCap API for this is kind of quirky.

REDCap's `filterLogic` documentation says:
> filterLogic
String of logic text (e.g., [age] > 30) for filtering the data to be returned by this API method, in which the API will only return the records (or record-events, if a longitudinal project) where the logic evaluates as TRUE. This parameter is blank/null by default unless a value is supplied. Please note that if the filter logic contains any incorrect syntax, the API will respond with an error message. 

It might be useful to mention putting square brackets around the field name, or using a single `=` as the equality operator. Alternatively, we could point users to the REDCap documentation for `filterLogic`.

joverlee521 (approved):

tsibley:
Nod. My reasoning for not including more details was that this code is a pretty thin wrapper around the REDCap API and assumes familiarity with the API itself and REDCap more broadly. There's also quite a lot of syntax that REDCap supports in its conditional logic expressions, enough that re-documenting it all is not feasible. Worse, there's not a single place to link to this doc in REDCap. It's spread out over various sections. :-/

jccraft-uw (approved):
This makes a lot of sense!

PR #168, opened 2020-09-17
Redcap cache

kairstenfay:
Parametrize the cache used in geocoding addresses processed by the REDCap DET ETLs.

Lock the cache file in python rather than relying on the user knowing to lock at runtime with shell's `flock`, etc.

- [ ] Update backoffice cron jobs accordingly


lib/id3c/cli/command/geocode.py
```diff
@@ -31,10 +31,10 @@
 from smartystreets_python_sdk.us_street import Lookup
 from smartystreets_python_sdk.us_extract import Lookup as ExtractLookup
 from id3c.cli import cli
+from id3c.cli.command import pickled_cache
```
kairstenfay, line 34: This edit can be dropped from this commit.

kairstenfay:
[Lock cache file in Python instead of externally with `flock`](https://trello.com/c/sgmbOCvZ/896-lock-cache-file-in-python-instead-of-externally-with-flock)

lib/id3c/cli/command/__init__.py
```diff
@@ -125,7 +126,9 @@ def pickled_cache(filename: str = None) -> TTLCache:
         LOG.info(f"Loading cache from «{filename}»")
         try:
             with open(filename, "rb") as file:
+                lock_file(file)
```
tsibley, line 129: I can't comment on the implementation of `lock_file()` yet, but I think the usage here isn't doing what you expect because it's not being called as a context manager. It should be called as:

```python
with lock_file(file):
    …
```

tsibley:
I think ultimately we may want to switch the geocoding caching from our in-house solution to [diskcache](http://www.grantjenks.com/docs/diskcache/), which I used for the Husky Musher caching. When we first added the caching, I didn't know diskcache existed and, unlike cachetools, didn't uncover it during due diligence.

@joverlee521 asked in Slack:

> What's the difference between the FanoutCache used for Musher and the TTLCache used for geocoding?

and I thought it'd be useful to copy my answer here as well:

> There's a couple differences:
>
> 1. TTLCache comes from cachetools, which implements lower-level cache primitives than FanoutCache from diskcache.
>
> 2. The TTLCache we use is entirely in-memory during execution. On top of that, we layer our own serialization/deserialization to load/save the cache. In contrast, the FanoutCache (diskcache.Cache subclass) is transactionally-safe and is loaded/saved per access. FanoutCache is a further-specialized subclass of diskcache.Cache to avoid waiting on blocking writes.
>
> 3. TTLCache requires a TTL (expiry). Expirations are optional with FanoutCache, and we don't use them.
>
> I found diskcache last night when looking not to reinvent the wheel. It seems like a very nice package that's robust and well-thought out. We could replace our TTLCache usage with it to make the geocoding caching a lot simpler.

PR #169, opened 2020-09-18
labelmaker: Add layout for UW Reopening barcodes

kairstenfay:
Add two new layouts for the UW Reopening (a.k.a. Husky Coronavirus
Testing) study collection barcode sets.

- [ ] Update Barcodes documentation

### DIFF
--- a/lib/id3c/labelmaker.py
+++ b/lib/id3c/labelmaker.py
@@ -239,14 +239,14 @@ class CollectionsUWObservedLayout(LabelLayout):
     sku = "LCRY-1100"
     barcode_type = 'UW OBSERVED'
     copies_per_barcode = 1
-    reference = "washington.edu"
+    reference = "seattleflu.org"
 
 
 class CollectionsUWMailLayout(LabelLayout):
     sku = "LCRY-1100"
     barcode_type = 'UW MAIL'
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Create labelmaker layout for UW reopening barcodes](https://trello.com/c/4wOwUjLg/774-create-labelmaker-layout-for-uw-reopening-barcodes)

joverlee521 (changes requested):
I think the reference should still be `seattleflu.org`?

kairstenfay:
> I think the reference should still be `seattleflu.org`?

I thought so too, but looking at [past Slack history](https://seattle-flu-study.slack.com/archives/C018V0WGGJ0/p1598542694302200?thread_ts=1598478177.269200&cid=C018V0WGGJ0), I wrote a post at some point that suggested uw.edu or washington.edu. I can't recall why but figured there was a reason for it at the time. However, no one commented on it in the thread. I can easily change it.

tsibley:
I think sticking with seattleflu.org makes most sense. The UW affiliation is indicated in the name already. Could compromise with seattleflu.org/uw.

joverlee521 (approved):

PR #170, opened 2020-09-30
etl fhir: Add REDCap source details to `encounter.details`

joverlee521:
~Splits the expected REDCap URL into individual keys within
`encounter.details`.~

See latest commit for details

Should be deployed along with https://github.com/seattleflu/id3c-customizations/pull/148

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -9,6 +9,7 @@ from collections import defaultdict
 from datetime import datetime, timezone
 from typing import Any, List, Dict, Optional, Tuple
 from urllib.parse import unquote
+from urllib.parse import urlparse
 from fhir.resources.bundle import Bundle, BundleEntry
 from fhir.resources.codeableconcept import CodeableConcept
 from fhir.resources.coding import Coding
@@ -481,6 +482,26 @@ def process_encounter_source(encounter: Encounter) -> Any:
     returns the deserialized JSON data.
 
     If the source is any other kind of URI, it will be returned as-is (as a string).
+
+    >>> _ = lambda source: Encounter({"meta":{"source":source},"status":"finished","class":{}})
+
+    >>> process_encounter_source(_('data:application/json,{"foo":"bar"}'))
+    {'foo': 'bar'}
+
+    >>> process_encounter_source(_("data:application/json,%7B%22foo%22%3A%22bar%22%7D"))
+    {'foo': 'bar'}
+
+    >>> process_encounter_source(_("data:application/json;base64,eyJmb28iOiJiYXIifQo="))
+    {'foo': 'bar'}
+
+    >>> process_encounter_source(_("data:text/plain,someplace somewhere"))
+    'data:text/plain,someplace somewhere'
+
+    >>> process_encounter_source(_("https://example.com"))
+    'https://example.com'
+
+    >>> process_encounter_source(_('data:application/json,bogus'))
+    'data:application/json,bogus'
     """
     if not encounter.meta:
         return None
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -8,6 +8,7 @@ import logging
 from collections import defaultdict
 from datetime import datetime, timezone
 from typing import Any, List, Dict, Optional, Tuple
+from urllib.parse import unquote
 from fhir.resources.bundle import Bundle, BundleEntry
 from fhir.resources.codeableconcept import CodeableConcept
 from fhir.resources.coding import Coding
@@ -442,6 +443,7 @@ def process_encounter(db: DatabaseSession, encounter: Encounter,
 
     contained_resources = extract_contained_resources(encounter)
 
+    encounter_source = process_encounter_source(encounter)
     encounter_reason = process_encounter_reason(encounter)
 
     part_of_identifier = None
@@ -455,6 +457,8 @@ def process_encounter(db: DatabaseSession, encounter: Encounter,
     details: Dict[str, Any] = encounter_details({ **related_resources, **contained_resources })
     if patient_language:
         details['language'] = patient_language
+    if encounter_source:
+        details['_provenance'] = encounter_source
     if encounter_reason:
         details['reason'] = encounter_reason
     if part_of_identifier:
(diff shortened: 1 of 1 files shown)


joverlee521:
[Add encounter provenance to `encounter.details`](https://trello.com/c/Rr3xrKQE/828-add-encounter-provenance-to-encounterdetails)

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -455,6 +456,8 @@ def process_encounter(db: DatabaseSession, encounter: Encounter,
     details: Dict[str, Any] = encounter_details({ **related_resources, **contained_resources })
     if patient_language:
         details['language'] = patient_language
+    if encounter_source:
+        details.update(encounter_source)
```
joverlee521, line 460: Should this be under an overall `__provenance` key like we currently have in `sample.details`?

jccraft-uw:
Why not put the encounter metadata (things like REDCap project ID, record ID, repeat instance, etc.) into a new 'metadata' column on the warehouse.encounter table instead of adding it to the 'details' column? Postgres doesn't have great query performance over json documents. Having the metadata stored separately could make for better query times. 
Also, conceptually, the details and metadata are different things. The details give you information about the encounter. The metadata tells you where the data came from.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -469,6 +472,41 @@ def process_encounter(db: DatabaseSession, encounter: Encounter,
...
+    encounter_source = encounter.meta.source
+
+    if not encounter_source:
+        return None
+
+    # Used named group pattern for easy conversion to dictionary after matching
+    # Include letters a-f in `redcap_record_id` for the barcode record id in UW Retros
+    pattern = re.compile(r'https://redcap\.iths\.org' + \
+                         r'(?P<redcap_project_id>/[0-9]+)' + \
+                         r'(?P<redcap_record_id>/[a-z0-9]+)' + \
```
kairstenfay, line 497: Reading your comment above, is this supposed to say `[a-f]` instead of `[a-z]`?

lib/id3c/cli/command/etl/fhir.py
kairstenfay: I think that would be nice for consistency with `sample.details`.

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -469,6 +472,41 @@ def process_encounter(db: DatabaseSession, encounter: Encounter,
...
+                         r'(?P<redcap_event_name>/\w*)?' + \
+                         r'(?P<redcap_repeat_instance>/[0-9]*)?')
```
kairstenfay, line 499: I just approved the related PR in i3c-customizations, but a thought just occurred to me. 

Instead of constructing a string just to deconstruct it in the next step with regex (which could cause unexpected bugs if the exact formula isn't followed for creating a URI string), why not pass along a dict or something we can parse directly for "project_id", "record_id" etc?

lib/id3c/cli/command/etl/fhir.py
joverlee521: Yes! Thank you for catching that!

lib/id3c/cli/command/etl/fhir.py
joverlee521: The `meta.source` seems to be the only appropriate field to hold provenance data and according to FHIR docs it has to be a URI. 

Would it be better to construct the string with query parameters? So the string would be something like:
```
https://redcap.iths.org/record?project_id=12345&record_id=123&event_name=encounter&repeat_instance=3
```

kairstenfay: Ah that figures. 

Hmm, clever idea. I _think_ that would end up being more robust. If there was a missing piece, say someone forgot to enter `event_name`, we would have a null value upon ingest rather than every value in the key-value pairs erroneously being shifted up one key.

kairstenfay: Then again, the regex is probably robust enough that this edge case may not even occur.

tsibley: I agree that it'd be better to use a more standard serialization of our key/value pairs. It will make this ETL code more robust, able to be extended with other keys later, and reduce de-serialization code we have to maintain (e.g. no more regex).

@joverlee521 your suggestion of query params is clever! :-) I wonder about going a step further though and removing the special-casing that even that would require (e.g. around how the URL origin is handled). I think the ideal is that the FHIR ETL routine can deserialize `meta.source` generically and simply pass it into `details`.

My suggestion is a `data:` URI containing JSON, e.g.:

```
data:application/json,{"redcap":{"url":"https://redcap.iths.org","project_id":"...","event_name":"..."}}
```

(The JSON part would be URL-encoded in practice.)

lib/id3c/cli/command/etl/fhir.py
tsibley: The key in `sample.details` uses only a single underscore, `_provenance`, as produced by the `id3c manifest parse…` commands.

tsibley:
@jccraft-uw 

> Why not put the encounter metadata (things like REDCap project ID, record ID, repeat instance, etc.) into a new 'metadata' column on the warehouse.encounter table instead of adding it to the 'details' column? Postgres doesn't have great query performance over json documents. Having the metadata stored separately could make for better query times.
Also, conceptually, the details and metadata are different things. The details give you information about the encounter. The metadata tells you where the data came from.

In our morning meeting on Friday, the discussion of this focused primarily on performance concerns. While I still think that will not be an issue on an indexed jsonb column like `details` for the kind of containment queries I expect we'll use on this provenance data (Pg's performance here is great), I _do_ think that there's a decent semantics-based argument to be made for separating system-maintained information (e.g. metadata driven by ID3C's core FHIR ETL in this case) from user-maintained information (e.g. `details`). There are other situations where this same system-vs-user semantic separation-of-concerns would apply too. That said, from a practical perspective, esp. one where we're both the system (in a sense) and the users, I'm inclined to use the catch-all `details` column we have for this at the moment and consider a more systematic data refactoring of `details` into `metadata` + `details` at a later point.

Happy to talk more about this next week too with your and/or whoever else is interested. I noted your comment while I was doing PR review tonight (working evenings this weekend and taking Monday off) and wanted to get my thoughts down while they were fresh.

tsibley (changes requested):
@joverlee521 Thanks for tackling this! I think using `meta.source` on the Encounter resource makes a lot of sense here. I left all my comments as replies to others' above. Let's work out this core FHIR ETL routine and once that's squared away update the related id3c-customizations PR?

lib/id3c/cli/command/etl/fhir.py
joverlee521: Cool! I didn't know you can do this with URIs! I'll update to use this method instead.

tsibley (approved):
@joverlee521 Thanks for the rework! I thought about some edge cases as I was reviewing and thought it'd be easiest to add failing doctests to demonstrate them. I ended up addressing the failing cases in the process too. Would you review the two commits I pushed and give me any feedback you have?

joverlee521:
>  I thought about some edge cases as I was reviewing and thought it'd be easiest to add failing doctests to demonstrate them. I ended up addressing the failing cases in the process too. Would you review the two commits I pushed and give me any feedback you have?

Thanks for covering the edge cases! The changes look good to me. It definitely didn't click for me that `urlparse` & `urlopen` would work for the data URI 😅

kschwabefry (approved):

jccraft-uw (approved):

tsibley:
If desired, I believe this could be merged and deployed in advance of https://github.com/seattleflu/id3c-customizations/pull/148.

PR #171, opened 2020-10-01
etl redcap-det: Fetch records from REDCap in batches no larger than 5,000

tsibley:
Alleviates presumed strain on the REDCap API in cases when we bump an ETL revision and refetch ~everything from a very large project (as just happened with the Husky Coronavirus Testing project, but could also happen with SCAN).

Batch size to fetch defaults to 5,000, but can be tuned at invocation using the new --fetch-batch-size option.  5,000 is a mostly arbitrary number.  It might not be the best value for any project and is certainly not the best value for all projects or servers.

All records are still fetched up-front and held in memory before the underlying ETL routine is called in a loop.  A future improvement would be to process each batch before fetching the next, but that requires partitioning the set of DETs too.

### DIFF
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -10,6 +10,7 @@ import logging
 from collections import defaultdict
 from datetime import datetime, timezone
 from functools import wraps
+from more_itertools import chunked
 from textwrap import dedent
 from typing import Callable, Iterable, Optional, Tuple, Dict, List, Any, DefaultDict
 from urllib.parse import urljoin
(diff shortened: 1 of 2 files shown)


joverlee521 (approved):
Thank you for the patch! 

In the long term, should the chunk size be dynamic based on which REDCap project we are processing? Assuming the number of repeat instances have an effect on the REDCap API, the chunk size might need to be much smaller for HCT later on in the project.

tsibley:
It's an excellent point that ideal batch size probably does depend on repeating instrument/event usage of each project. I briefly thought about how to automatically dynamically scale it, but saw no obvious answer. Per-project would make sense then as a way to have a manually-tuned value for each.

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -170,8 +171,10 @@ def decorated(*args, db: DatabaseSession, log_output: bool, **kwargs):
                 # Records with repeating instruments or longitudinal
                 # events will have multiple entries in the list.
                 redcap_records: DefaultDict[str, List[dict]] = defaultdict(list)
-                for record in project.records(ids = record_ids, raw = raw_coded_values):
-                    redcap_records[record.id].append(record)
+
+                for chunk in chunked(record_ids, 5000):
```
kairstenfay, line 175: Small request to make this magic number a global variable or comment on it

tsibley: This number is now an option default, where it is self-documenting.

tsibley:
Repushed to add tuning via `--fetch-batch-size` and a little more informational logging to be helpful and tested it locally.

kairstenfay (approved):
Thanks!

joverlee521 (approved):

PR #172, opened 2020-10-04
db session: Provide a default application name

tsibley:
This helps identify connections originating from within ID3C code when
looking at Postgres' pg_stat_activity.

By specifying a default, specific jobs or invocations can still provide
a more usage-specific name via the PGAPPNAME environment variable (e.g.
alongside the credential variables) if desired.

### DIFF
--- a/lib/id3c/db/session.py
+++ b/lib/id3c/db/session.py
@@ -1,6 +1,7 @@
 """
 Database session creation and management.
 """
+import click
 import logging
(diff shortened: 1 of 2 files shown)


jccraft-uw (approved):
Having an app name in there will be handy.

joverlee521 (approved):

kairstenfay (approved):

PR #173, opened 2020-10-08
Google doc metadata

jccraft-uw:
See the commit messages for details.

This PR adds a function to get Google Drive document metadata and then uses that to create the digest for the incident report manifest.

### DIFF
--- a/lib/id3c/cli/io/google.py
+++ b/lib/id3c/cli/io/google.py
@@ -82,7 +82,7 @@ def extract_document_id_from_google_url(url_str: str) -> Optional[str]:
 
     return google_docs_matches["document_id"] if google_docs_matches else None
 
-def get_document_details(document_id) -> dict:
+def get_document_details(document_id: str) -> dict:
     """
     Returns a set of metadata details about a Google Drive document.
     """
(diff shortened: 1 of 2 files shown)


lib/id3c/cli/io/google.py
```diff
@@ -81,3 +81,15 @@ def extract_document_id_from_google_url(url_str: str) -> Optional[str]:
     google_docs_matches = re.match(google_docs_pattern, url.path)
 
     return google_docs_matches["document_id"] if google_docs_matches else None
+
+def get_document_details(document_id) -> dict:
```
kairstenfay, line 85: Could you please add a type hint for `document_id` here?

lib/id3c/cli/io/google.py
```diff
@@ -81,3 +81,15 @@ def extract_document_id_from_google_url(url_str: str) -> Optional[str]:
...
+    """
+    Returns a set of metadata details about a Google Drive document.
+    """
+    drive_service = discovery.build("drive", "v2")
+    metadata = drive_service.files().get(fileId=document_id).execute()
+
+    details = {'version': metadata['version'], 'etag': metadata['etag'], 'modified_date': metadata['modifiedDate'],
+        'modifiying_user': metadata['lastModifyingUserName']}
+
+    return details
```
joverlee521, line 95: Just curious, why rename the fields returned by Google? I feel like it's an extra layer of field names that we would need to keep track of here.

jccraft-uw: I wanted the fields to match our casing convention and went with modifying_user over lastModifyUserName to simplify it.

lib/id3c/cli/io/google.py
jccraft-uw: Added.

joverlee521 (approved):

PR #174, opened 2020-10-14
Add event name and instance ID to REDCap Record

jccraft-uw:
This short change adds event name and repeat instance ID to the REDCap Record object.
I haven't seen anything that indicates that the field names for these vary across projects, so I did not make the field names project attributes like we do for record_id.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -7,7 +7,7 @@ import requests
 from enum import Enum
 from functools import lru_cache
 from operator import itemgetter
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 
 LOG = logging.getLogger(__name__)
@@ -257,11 +257,21 @@ class Record(dict):
     """
     project: Project
     id: str
+    event_name: Optional[str]
+    repeat_instance: Optional[int]
 
     def __init__(self, project: Project, data: Any = {}) -> None:
         super().__init__(data)
         self.project = project
         self.id = self[self.project.record_id_field]
+
+        # These field names are not variable across REDCap projects
+        self.event_name = self.get("redcap_event_name")
+
+        if self.get("redcap_repeat_instance"):
+            self.repeat_instance = int(self["redcap_repeat_instance"])
+        else:
+            self.repeat_instance = None
 
 
 class InstrumentStatus(Enum):
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -254,6 +254,11 @@ class Record(dict):
     Must be constructed with a REDCap :py:class:``Project``, which is stored as
     the ``project`` attribute and used to set the ``id`` attribute storing the
     record's primary id.
+
+    Note that event_name and repeat_instance might not be populated because their
+    corresponding fields (redcap_event_name and redcap_repeat_instance)
+    won't be returned by the API if the request includes the 'fields' parameter
+    but not the record_id field.
     """
     project: Project
     id: str
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -255,28 +255,43 @@ class Record(dict):
     the ``project`` attribute and used to set the ``id`` attribute storing the
     record's primary id.
 
-    Note that the ``event_name`` and ``repeat_instance`` attributes might not
-    be populated because their corresponding fields (``redcap_event_name`` and
-    ``redcap_repeat_instance``) won't be returned by the API if the request
+    Note that the ``event_name``, ``repeat_instance``, and ``repeat_instrument``
+    attributes might not be populated because their corresponding fields
+    (``redcap_event_name``, ``redcap_repeat_instance``, and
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Thanks for updating this!

One thing to note is the `redcap_event_name` and `redcap_repeat_instance` fields may not always be in the returned data from the REDCap API. If the API request includes the `fields` parameter but does not include `record_id` field then these fields are not returned. 
It won't break anything, but might be good to note for future reference in case either field is not available when we expect it to be.

jccraft-uw:
> Thanks for updating this!
> 
> One thing to note is the `redcap_event_name` and `redcap_repeat_instance` fields may not always be in the returned data from the REDCap API. If the API request includes the `fields` parameter but does not include `record_id` field then these fields are not returned.
> It won't break anything, but might be good to note for future reference in case either field is not available when we expect it to be.

That is such a good point! I put that in the documentation for Record.

lib/id3c/cli/redcap.py
```diff
@@ -254,15 +254,30 @@ class Record(dict):
...
     """
     project: Project
     id: str
+    event_name: Optional[str]
+    repeat_instance: Optional[int]
 
     def __init__(self, project: Project, data: Any = {}) -> None:
         super().__init__(data)
         self.project = project
         self.id = self[self.project.record_id_field]
```
joverlee521, line 271: Ah, actually now it's apparent that this will break if the data returned from the API request doesn't include the record's primary id....

Would it be weird to ensure that the record's primary id is _always_ included in the `fields` parameter if the `fields` parameter is provided?

tsibley: I can imagine situations where that would be undesirable (e.g. when the primary id field contains PII and so is intentionally not requested), but I think for our purposes it would be reasonable to always include it. Alternatively, we could instead adjust this code in `Record` to DTRT.

lib/id3c/cli/redcap.py
```diff
@@ -254,15 +254,30 @@ class Record(dict):
     Must be constructed with a REDCap :py:class:``Project``, which is stored as
     the ``project`` attribute and used to set the ``id`` attribute storing the
     record's primary id.
+
+    Note that event_name and repeat_instance might not be populated because their
+    corresponding fields (redcap_event_name and redcap_repeat_instance)
+    won't be returned by the API if the request includes the 'fields' parameter
+    but not the record_id field.
```
tsibley, line 261:
```suggestion
    Note that the ``event_name`` and ``repeat_instance`` attributes might not
    be populated because their corresponding fields (``redcap_event_name`` and
    ``redcap_repeat_instance``) won't be returned by the API if the request
    includes the ``fields`` parameter but not the primary record id field.
```

Tweaks to make this docstring addition 1) conform to the rest of the codebase's use of rST syntax and 2) be more precise about `record_id`.

lib/id3c/cli/redcap.py
jccraft-uw: Would the right thing here mean that Record could return None for id?
It seems like raising an exception would be better since consumers would most likely assume that record.id is always set.

lib/id3c/cli/redcap.py
```diff
@@ -254,15 +254,30 @@ class Record(dict):
...
     """
     project: Project
     id: str
+    event_name: Optional[str]
+    repeat_instance: Optional[int]
```
tsibley, line 266: For completeness, it'd be good to include `repeat_instrument` too. I believe we'll even want this shortly for the childcare ingest…?

tsibley (changes requested):

lib/id3c/cli/redcap.py
joverlee521: Oh that's true, all of our downstream processes depend on the record's primary id to exist.

tsibley: Yes. I'd probably turn the `id` attribute into a dynamic getter so the exception only happens on access. Something like:

```python
@property
def id(self) -> str:
    """
    Returns the record's primary id.

    Raises a :py:class:`RuntimeError` if the primary id field is not available,
    usually because it was not requested from the API.
    """
    try:
        return self[self.project.record_id_field]
    except KeyError as e:
        raise RuntimeError(f"Primary record id field «{self.project.record_id_field}» not available on fetched record") from e
```

lib/id3c/cli/redcap.py
jccraft-uw: Childcare isn't using repeated instruments, but I added repeat_instrument for future use. That's a good idea since we're in here.

lib/id3c/cli/redcap.py
jccraft-uw: Good idea. Don't raise an exception unless the consumer is trying to access id. I've added that in in the latest commit.

kairstenfay (approved):
Thanks all for iterating on this!

lib/id3c/cli/redcap.py
jccraft-uw: @tsibley are you OK with what's in the final commit for this getter?

PR #177, opened 2020-10-21
Incomplete manifest

kairstenfay:
Stop ingesting incomplete rows from the manifest.

The biggest question I have is do we want to stop the entire ingest (e.g. raise an Exception) when we see incomplete rows, or do we just want to drop the incomplete rows from the ingest as I'm currently doing?

I wanted air this question while I continue testing to try to reproduce the original problem and assure that this PR fixes it.

### DIFF
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -484,6 +484,8 @@ def perform_qc(sample_column: str, collection_column: str, barcode_columns: Set[
     """
     parsed_manifest = drop_missing_barcodes(sample_column, collection_column, parsed_manifest)
 
+    parsed_manifest = drop_missing_racks(parsed_manifest)
+
     # Drop any rows that have duplicated barcodes
     parsed_manifest = deduplicate_barcodes(parsed_manifest, barcode_columns)
     return parsed_manifest
@@ -575,6 +577,21 @@ def deduplicate_barcodes(df: pandas.DataFrame, columns: Iterable) -> pandas.Data
             deduplicated = deduplicated.loc[common_idx]
 
     return deduplicated
+
+
+def drop_missing_racks(parsed_manifest: pandas.DataFrame) -> pandas.DataFrame:
+    """
+    Drop rows with no rack information in a given *parsed_manifest*.
+    Rows without rack information are typically incomplete rows that will break
+    our ingest.
+
+    >>> drop_missing_racks(pandas.DataFrame([['aa', 'bb', [1, 2, None]], \
+        ['cc', 'dd', [None, None, None]]], columns=['sample', 'collection', 'racks']))
+      sample collection         racks
+    0     aa         bb  [1, 2, None]
+    """
+    racks_complete = parsed_manifest['racks'].apply(lambda cols: any(cols))
+    return parsed_manifest[racks_complete]
 
 
 def deephash(record):
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -484,10 +484,11 @@ def perform_qc(sample_column: str, collection_column: str, barcode_columns: Set[
     """
     parsed_manifest = drop_missing_barcodes(sample_column, collection_column, parsed_manifest)
 
-    parsed_manifest = drop_missing_racks(parsed_manifest)
-
     # Drop any rows that have duplicated barcodes
     parsed_manifest = deduplicate_barcodes(parsed_manifest, barcode_columns)
+
+    parsed_manifest = drop_missing_racks(parsed_manifest)
+
     return parsed_manifest
 
 
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -25,7 +25,7 @@ from deepdiff import DeepHash
 from hashlib import sha1
 from os import chdir
 from os.path import dirname
-from typing import Iterable, List, Optional, Tuple, Union
+from typing import Iterable, List, Optional, Set, Tuple, Union
 from id3c.cli import cli
 from id3c.cli.io import LocalOrRemoteFile, urlopen
 from id3c.cli.io.google import *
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Update manifest parse logic to not ingest "incomplete" records when we have a large number of null values](https://trello.com/c/CO99f2lz/687-update-manifest-parse-logic-to-not-ingest-incomplete-records-when-we-have-a-large-number-of-null-values)

lib/id3c/cli/command/manifest.py
```diff
@@ -515,6 +579,21 @@ def qc_barcodes(df: pandas.DataFrame, columns: Iterable) -> pandas.DataFrame:
...
+    Drop rows with no rack information in a given *parsed_manifest*.
+    Rows without rack information are typically incomplete rows that will break
+    our ingest.
+
+    >>> drop_missing_racks(pandas.DataFrame([['aa', 'bb', [1, 2, None]], \
+        ['cc', 'dd', [None, None, None]]], columns=['sample', 'collection', 'racks']))
+      sample collection         racks
+    0     aa         bb  [1, 2, None]
+    """
+    racks_complete = parsed_manifest['racks'].apply(lambda cols: any(cols))
```
jccraft-uw, line 593: That's so short and sweet!

jccraft-uw (approved):
This refactoring will make it easy to add more checks/filters in the future if we need to.
Because we won't need to take action when the manifest contains empty racks--because we can silently skip the affected rows and ingest them later when they are fully populated in the manifest--I think skipping the rows and not error'ing out is the way to go.

lib/id3c/cli/command/manifest.py
```diff
@@ -487,10 +476,85 @@ def select_columns(table: pandas.DataFrame, name: str) -> pandas.DataFrame:
...
+    """
+    Perform quality control on the manifest data, dropping rows which violate
+    our standards for complete and accurate data.
+    """
+    parsed_manifest = drop_missing_barcodes(sample_column, collection_column, parsed_manifest)
+
+    parsed_manifest = drop_missing_racks(parsed_manifest)
+
+    # Drop any rows that have duplicated barcodes
+    parsed_manifest = deduplicate_barcodes(parsed_manifest, barcode_columns)
```
joverlee521, line 490: Should drop racks and drop barcode ordering be switched here? If we drop one of the duplicate barcode pair in the `drop_missing_racks` function then we will ingest the other one of the duplicate pair. Does that make sense?

kairstenfay: Ah yes, thank you! That was another question I had but forgot to list at the top of this PR.

I wasn't sure which ordering was best. I took this initial pass thinking maybe the duplicate barcode with an incomplete entry (missing racks) is OK to drop and fine to ingest the other matching barcode with a complete entry. 

But it seems like you disagree. Which one seems preferable? I can see that dropping duplicate barcodes before doing more filtering would be safer.

jccraft-uw: I like to keep the duplicate dropping to the very last step because previous steps might get rid of some of those anyway.

If we have one sample record that has no rack info and another record with the same identifiers and with rack info I think we would want to use that second record.

joverlee521: If the second record is completed at a later date, they are still true duplicates and both should be dropped.

jccraft-uw: We'd get that record on the next run, so I'm OK dropping it on the first run. I'm OK putting the duplication dropper above the rack dropper.

kairstenfay: Is it worth evaluating date to try to pull in additional records? Now that you lay it out like that, @joverlee521 , I think conditionally dropping barcodes first based on date, filtering where racks is not None, and then dropping barcodes again feels a little clunky just in the off chance we might rescue a couple of rows here and there. I'd rather put the onus on the lab to clean their manifest data instead of using obscure logic to ingest most of it is by far the more robust solution. The lab would also appreciate knowing if they have data entry errors (which they'd learn if we don't ingest certain barcodes at all).

kairstenfay: Ah I just saw Chris's comment. Looks like we all agree now.

kairstenfay: I pushed a new commit that deduplicates first.

joverlee521 (approved):
Thank you for refactoring and setting up the way to add future QC. Hopefully the empty racks check will be enough to filter out the incomplete records 🤞 

I do want to comment that the `perform_qc` function might be very specific to SFS and not completely "core". We've been lagging in separating "core" and SFS specific ID3C components so not sure if we want to address this now or continue kicking the can.

kairstenfay:
> I do want to comment that the `perform_qc` function might be very specific to SFS and not completely "core". We've been lagging in separating "core" and SFS specific ID3C components so not sure if we want to address this now or continue kicking the can.

Oh darn. I remember discussing this in the initial design of this task, but I completely forgot once I started writing the code. 
Perhaps we could extend the config to take an argument for a specified set of columns (e.g. "racks") that must not be entirely null?

kairstenfay:
> > I do want to comment that the `perform_qc` function might be very specific to SFS and not completely "core". We've been lagging in separating "core" and SFS specific ID3C components so not sure if we want to address this now or continue kicking the can.
> 
> Oh darn. I remember discussing this in the initial design of this task, but I completely forgot once I started writing the code.
> Perhaps we could extend the config to take an argument for a specified set of columns (e.g. "racks") that must not be entirely null?

I admit I'm leaning towards getting this deployed sooner rather than later given the scope of other tasks on the horizon, but I also don't want to make our future lives unnecessarily more difficult.

joverlee521:
> > Perhaps we could extend the config to take an argument for a specified set of columns (e.g. "racks") that must not be entirely null?

> I admit I'm leaning towards getting this deployed sooner rather than later given the scope of other tasks on the horizon, but I also don't want to make our future lives unnecessarily more difficult.

Yup, I agree it would be nice to get this deployed as soon as possible to prevent the annoying errors. This might a good future improvement to add to our (evergrowing) backlog.

kairstenfay:
Rebased onto master.

tsibley (changes requested):
I don't like hardcoding requirements on `racks` here, and I believe if merged as-is, it will break the manifest parsing we do for UW retrospectives.

In our 1:1, we talked about doing this filtering on `racks` within the specimen-manifest repo's `update` process, e.g. by adding something like:

```
… | jq 'select(.racks | map(select(values)) | any)' | …
```

between `id3c manifest parse` and `id3c manifest diff`.

Benefits of this approach are that it:

1. avoids hardcoding in core
2. keeps things minimal instead of further developing core to provide hooks or more config knobs
3. makes use of the reason why the `id3c manifest …` commands are separate, composable commands up in the first place: to allow for viewing and manipulation of the stream of records at each step.

Note that the racks filter doesn't have to be jq (although it's perfect here): it could be a small Python program (or any other filtering program) instead if that's preferred.

kairstenfay:
Thank you both!

kairstenfay:
Closing in favor of https://github.com/seattleflu/id3c/pull/178 for the refactoring plus test addition work, and in favor of https://github.com/seattleflu/specimen-manifests/pull/20 for the custom empty racks filtering.

PR #179, opened 2020-10-27
etl redcap-det: Add optional --limit on number of received DETs

tsibley:
This will help us manage large numbers of DETs from Husky Coronavirus
Testing (and SCAN) when bumping ETL revisions and reprocessing
everything.  With a limit, subsequent ETL jobs are responsible for
picking up remaining DETs to process.

### DIFF
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -7,6 +7,7 @@ REDCap.
 import os
 import click
 import logging
+import psycopg2.sql as sql
 from collections import defaultdict
 from datetime import datetime, timezone
 from functools import wraps
@@ -89,6 +90,13 @@ def command_for_project(name: str,
     }
 
     def decorator(routine: Callable[..., Optional[dict]]) -> click.Command:
+        @click.option("--det-limit",
+            metavar = "<number>",
+            help    = "Maximum number of received DETs to process in this invocation. "
+                      "If more unprocessed DETs exist, they will be processed on the next invocation of this ETL. "
+                      "Not limited by default.",
+            type    = int)
+
         @click.option("--redcap-api-batch-size",
             metavar = "<number>",
             help    = "Maximum number of records to fetch in one batch from REDCap. "
@@ -104,7 +112,7 @@ def command_for_project(name: str,
         @with_database_session
         @wraps(routine)
 
-        def decorated(*args, db: DatabaseSession, log_output: bool, redcap_api_batch_size: int, **kwargs):
+        def decorated(*args, db: DatabaseSession, log_output: bool, det_limit: int = None, redcap_api_batch_size: int, **kwargs):
             LOG.debug(f"Starting the REDCap DET ETL routine {name}, revision {revision}")
 
             # If the correct environment variables aren't defined, this will
@@ -113,15 +121,23 @@ def command_for_project(name: str,
             api_url = urljoin(redcap_url, "api/")
             api_token = get_redcap_api_token(api_url)
 
+            if det_limit:
+                LOG.debug(f"Processing up to {det_limit:,} pending DETs")
+                limit = sql.Literal(det_limit)
+            else:
+                LOG.debug(f"Processing all pending DETs")
+                limit = sql.SQL("all")
+
             redcap_det = db.cursor(f"redcap-det {name}")
-            redcap_det.execute("""
+            redcap_det.execute(sql.SQL("""
                 select redcap_det_id as id, document
                   from receiving.redcap_det
                  where not processing_log @> %s
                    and document::jsonb @> %s
                  order by id
+                 limit {}
                    for update
-                """, (Json([etl_id]), Json(det_contains)))
+                """).format(limit), (Json([etl_id]), Json(det_contains)))
 
             # First loop of the DETs to determine how to process each one.
             # Uses `first_complete_dets` to keep track of which DET to
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -89,9 +89,9 @@ def command_for_project(name: str,
     }
 
     def decorator(routine: Callable[..., Optional[dict]]) -> click.Command:
-        @click.option("--fetch-batch-size",
+        @click.option("--redcap-api-batch-size",
             metavar = "<number>",
-            help    = "Number of records to fetch in one batch from REDCap. "
+            help    = "Maximum number of records to fetch in one batch from REDCap. "
                       "Multiple batches will be fetched if there are more than this number of records to fetch. "
                       "The default (5,000) is somewhat arbitrary and what will or won't strain the REDCap API depends on both the number of repeating instrument/event instances and the REDCap server itself.",
             default = 5000)
@@ -104,7 +104,7 @@ def command_for_project(name: str,
         @with_database_session
         @wraps(routine)
 
-        def decorated(*args, db: DatabaseSession, log_output: bool, fetch_batch_size: int, **kwargs):
+        def decorated(*args, db: DatabaseSession, log_output: bool, redcap_api_batch_size: int, **kwargs):
             LOG.debug(f"Starting the REDCap DET ETL routine {name}, revision {revision}")
 
             # If the correct environment variables aren't defined, this will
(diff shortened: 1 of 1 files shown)


lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -89,9 +90,16 @@ def command_for_project(name: str,
...
 
     def decorator(routine: Callable[..., Optional[dict]]) -> click.Command:
+        @click.option("--limit",
+            metavar = "<number>",
+            help    = "Maximum number of received DETs to process in this invocation. "
+                      "If more unprocessed DETs exist, they will be processed on the next invocation of this ETL. "
+                      "Not limited by default.",
+            type    = int)
+
         @click.option("--fetch-batch-size",
```
jccraft-uw, line 100: Could we name the two options something more specific? Something like:
--limit > --det_batch_size
--fetch-batch-size > --redcap_batch_size

jccraft-uw (changes requested):
The logic is clean and simple. The request is to name the 2 options so that it's easier to understand what they're affecting.

lib/id3c/cli/command/etl/redcap_det.py
kairstenfay: I'm not sure I follow, @jccraft-uw. 
Both of them relate to REDCap DETs, so the distinction between the suggested "det" and "redcap" prefixes doesn't register with me.

IIUC, we fetch DETs from REDCap, in given batch of size N, up to a limit M.

kairstenfay (approved):

joverlee521 (approved):

lib/id3c/cli/command/etl/redcap_det.py
jccraft-uw: They do both relate to REDCap.
limit : How many records should the job pull from the receiving.redcap_det table?
fetch_batch_size: When querying REDCap for records how many are queried in one batch?

It would make it easier to know what each option controls if we give them a more specific name.

tsibley: @jccraft-uw Fair point, and thank you for considering usability concerns like this! I'm open to using different names here, although I don't much like `--det-batch-size` and `--redcap-batch-size` as alternatives.¹ Let's find other options (har har) that will work for everyone.

Part of my thinking that lead me to `--limit` is that I anticipate we'll eventually want the same option across all `id3c etl` commands, not just those under `id3c etl redcap-det`. This was actually what I went into the codebase intending to do, but then realized that we still hadn't yet refactored our ETLs' SQL select pattern into a common decorator yet. Bummer. The point being though, is that `--limit` is the hard limit of `receiving.*` rows to process—that's only `receiving.redcap_det` for now—and the concept of a "limit" will be understood by anyone familiar with SQL (which is good, since that's exactly what its coupled to here).

When writing the code, I considered the following instead of `--limit`:

- `--batch-size`, but "batch" to me implies that something will be done repeatedly in chunks of that size. This is true for `--fetch-batch-size`, but it isn't true for the functionality I'm adding here.

- `--size` and `--count`, but these felt too vague.

- `-n`, which has CLI tradition on its side for a limiting option, but this felt too terse.

Thinking through this again now, maybe `--receiving-limit` would be more clear?

I agree that the existing `--fetch-batch-size` option isn't as clearly named as I'd like, but I struggled at the time to find alternatives that felt better. I liked the "batch size" part because of the connotations it has for me, as above. I considered `--redcap-batch-size`, but thought that was maybe confusing for some other behaviour since "redcap" could mean the DETs or the records. "Fetch" felt better for that reason, as it implies accessing a remote/external resource for me.

Maybe `--redcap-api-batch-size` would be more self-explanatory without verging on verbosity? This (correctly) suggests a connection with the `REDCAP_API_…` environment variables too.

¹ CLI options are by overwhelming convention hyphen-separated, not underscore separated, so I assume that's what you meant.

jccraft-uw: We configure the cron jobs with the options one time. It doesn't seem like a huge imposition to make longer option names.

kairstenfay: Oh it's certainly not a "huge imposition", but I think it's important to consider the user experience of the tools we write which are for primarily for humans (not cron).

tsibley: I think you're both talking about improving the UI for humans but differ about what that means. To @jccraft-uw, `--limit` is too vague and `--receiving-limit` more clearly indicates its purpose. For @kairstenfay, `--limit` is clear while `--receiving-limit` seems to add an unnecessary and wordy detail.

I'm ambivalent here. `--limit` is what made sense to me originally, but I'm ok to make it `--receiving-limit` instead.

If there's more people want to say here, then I'm open to continuing the conversation for a little more time, but if things are at an impasse, then I don't want to get too bogged down in this decision. If that's the case, then either I can pick one option or support both.

jccraft-uw: I can look at the code if I forget what each option does. I'm OK with short names since that is the team norm.

kairstenfay: I'm mostly ambivalent here, too, and I certainly don't want to create an impasse. I will happily accept whatever the outcome here, whether we change one, both, or neither option names discussed in this thread. 

I just wanted to express my support for @tsibley's naming an option "limit" that's as close to the SQL as possible; but also, I agree with @jccraft-uw  that `--fetch-batch-size` has room for improvement (naming is consistently one of the hardest parts of programming in my experience).

@jccraft-uw If one of us needs to dig into the code to figure out what an option means, then we've created unusable tools. We should try to avoid that at all costs! Would the help message be useful to you? If you run any of our id3c commands with `--help`, you should get a nice description of what the option does. There's a lot of room for verbosity there.

jccraft-uw: It would be nice for the options to be clear enough that one doesn't need to read the help to know what they do.
Again, though, I withdraw my request.

jccraft-uw (approved):

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -113,15 +121,23 @@ def decorated(*args, db: DatabaseSession, log_output: bool, fetch_batch_size: in
...
+            redcap_det.execute(sql.SQL("""
                 select redcap_det_id as id, document
                   from receiving.redcap_det
                  where not processing_log @> %s
                    and document::jsonb @> %s
                  order by id
+                 limit {}
                    for update
-                """, (Json([etl_id]), Json(det_contains)))
+                """).format(limit_count), (Json([etl_id]), Json(det_contains)))
```
kschwabefry, line 140: Is is true that `redcap_det_id` is guaranteed to ascend over time?
I'm making that assumption based on code and comments below, just wanted to ask explicitly, as it seems pretty important to the logic here. If this is called with a limit option that is less than the number of pending DETs, then the `limit` in the query here, coupled with the `order by redcap_det_id`, should get us the earliest DETs which is what we want. 👍

tsibley: Yep. The ETLs intentionally process received records in insert order, and this isn't changed by the addition of `limit` here.

`redcap_det_id` is a generated id column controlled by this sequence:

```
tom@production.db.seattleflu.org/production=> \d receiving.redcap_det_redcap_det_id_seq 
          Sequence "receiving.redcap_det_redcap_det_id_seq"
  Type   │ Start │ Minimum │  Maximum   │ Increment │ Cycles? │ Cache 
─────────┼───────┼─────────┼────────────┼───────────┼─────────┼───────
 integer │     1 │       1 │ 2147483647 │         1 │ no      │     1
Sequence for identity column: receiving.redcap_det.redcap_det_id
```

If there are holes in the actually-used ids (due to aborted transactions or deleted rows), it would be possible to create new rows which would fill those with lower numbers than the sequence's current value. But in practice nothing is doing that.

kschwabefry: Excellent. Thanks Tom!

kschwabefry (approved):

lib/id3c/cli/command/etl/redcap_det.py
tsibley: Ok.

Here's what I've done, as a compromise that I feel good about:

- `--limit` is now `--det-limit`
- `--fetch-batch-size` is now `--redcap-api-batch-size`

I'm going to merge this and if there are further concerns, then we can address those separately.

PR #180, opened 2020-10-27
Configurable cache

kairstenfay:
Here are some commits I did awhile ago as part of a bigger picture to lock the cache file within python (instead of relying on a user to create file lock on the cache at runtime).

These commits alone may be enough to ease up on some of the constraint on our system -- that is, by using different caches for different ETL jobs, we can allow some jobs to run concurrently.

### DIFF
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -19,7 +19,7 @@ from id3c.cli.command import with_database_session
 from id3c.cli.redcap import is_complete, Project
 from id3c.db.session import DatabaseSession
 from id3c.db.datatypes import as_json, Json
-from id3c.cli.command.geocode import pickled_cache
+from id3c.cli.command import pickled_cache
 from . import etl
 
 
--- a/lib/id3c/cli/command/etl/redcap_det.py
+++ b/lib/id3c/cli/command/etl/redcap_det.py
@@ -26,18 +26,6 @@ from . import etl
 LOG = logging.getLogger(__name__)
 
 
-# XXX FIXME: I don't think we should hardcode a cache name like this,
-# particularly with a name that doesn't give any hint as to what uses it or
-# what it contains. The `id3c geocode` command, for instance, explicitly
-# parameterizes the cache file as an option.
-#
-# Going a step further, I don't think @command_for_project should even be
-# providing the "cache" parameter.  What is cached and where it is stored is
-# something specific to each REDCap DET routine, not a global invariant.
-#   -trs, 19 Dec 2019
-CACHE_FILE = 'cache.pickle'
-
-
 @etl.group("redcap-det", help = __doc__)
 def redcap_det():
     pass
@@ -108,11 +96,21 @@ def command_for_project(name: str,
             help        = "Write the output FHIR documents to stdout. You will likely want to redirect this to a file",
             default     = False)
 
+        # XXX FIXME: I don't think @command_for_project should even be
+        # providing the "cache" parameter.  What is cached and where it is stored is
+        # something specific to each REDCap DET routine, not a global invariant.
+        #   -trs, 19 Dec 2019
+        @click.option("--geocoding-cache",
+            metavar = "<cache.pickle>",
+            help = "Local file for caching the results of address geocoding. Geocoding lookups will not be cached otherwise.",
+            required = False,
+            type = click.Path(dir_okay=False, writable=True))
+
         @redcap_det.command(name, **kwargs)
         @with_database_session
         @wraps(routine)
 
-        def decorated(*args, db: DatabaseSession, log_output: bool, det_limit: int = None, redcap_api_batch_size: int, **kwargs):
+        def decorated(*args, db: DatabaseSession, log_output: bool, det_limit: int = None, redcap_api_batch_size: int, geocoding_cache: str = None, **kwargs):
             LOG.debug(f"Starting the REDCap DET ETL routine {name}, revision {revision}")
 
             # If the correct environment variables aren't defined, this will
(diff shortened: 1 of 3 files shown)


kairstenfay:
[Cache handling in REDCap → FHIR generation](https://trello.com/c/69EYeCSn/296-cache-handling-in-redcap-%E2%86%92-fhir-generation)

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -188,7 +186,7 @@ def decorated(*args, db: DatabaseSession, log_output: bool, fetch_batch_size: in
                         redcap_records[record.id].append(record)
 
             # Process all DETs in order of redcap_det_id
-            with pickled_cache(CACHE_FILE) as cache:
+            with pickled_cache(cache_file) as cache:
```
jccraft-uw, line 189: What if cache_file == None here?

kairstenfay:
Oops! It's been so long since I wrote these commits, I forgot there are dependencies in id3c-customizations that need to be updated (w.r.t geocoding). Going to mark this as a draft to reflect that since there is no live id3c-customizations PR yet.

lib/id3c/cli/command/etl/redcap_det.py
kairstenfay: If no cache file is provided, we use an in-memory cache.
The log will output:
```
[2020-10-28 09:40:01] No cache file provided; using transient, in-memory cache.
```

kairstenfay:
The id3c-customizations PR, to be deployed in coordination with this one, is now open for review.

joverlee521 (approved):
Is there going to be a corresponding change in the backoffice repo to use the new `--cache-file` option for all of the redcap-det etls?

kairstenfay:
> Is there going to be a corresponding change in the backoffice repo to use the new `--cache-file` option for all of the redcap-det etls?

Yes. Thank you for prompting me for this.

lib/id3c/cli/command/geocode.py
```diff
@@ -248,53 +245,8 @@ def standardize_address(address_series: pd.Series,
...
-    try:
-        yield cache
-    finally:
-        if filename:
-            with open(filename, "wb") as file:
-                pickle.dump(cache, file)
-
 
-def get_response_from_cache_or_geocoding(address: dict,
+def get_geocoded_address(address: dict,
```
kschwabefry, line 249: `get_response_from_cache_or_geocoding` looks to be imported by various ETLs in id3c-customizations https://github.com/search?q=org%3Aseattleflu+get_response_from_cache_or_geocoding&type=code I think this rename would be a breaking change without a coordinated update in id3c-customizations.

kschwabefry (changes requested):

lib/id3c/cli/command/geocode.py
kschwabefry: Scratch that! Just saw the id3c-customizations PR here: https://github.com/seattleflu/id3c-customizations/pull/160

kschwabefry (approved):

lib/id3c/cli/command/geocode.py
kairstenfay: Sorry, I should have explicitly pointed from this PR to the id3c-customizations one instead of pointing in only one direction.

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -100,11 +88,21 @@ def decorator(routine: Callable[..., Optional[dict]]) -> click.Command:
...
 
+        # XXX FIXME: I don't think @command_for_project should even be
+        # providing the "cache" parameter.  What is cached and where it is stored is
+        # something specific to each REDCap DET routine, not a global invariant.
+        #   -trs, 19 Dec 2019
+        @click.option("--cache-file",
+            metavar = "<cache.pickle>",
+            help = "Local cache file for storing address lookups. Must be specified to cache lookups",
+            required = False,
+            type = click.Path())
```
tsibley, line 99: Good use of `click.Path()`. It'd be nice to override [its defaults](https://click.palletsprojects.com/en/7.x/api/#click.Path) that aren't applicable here:

```
click.Path(dir_okay = False, writable = True)
```

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -100,11 +88,21 @@ def decorator(routine: Callable[..., Optional[dict]]) -> click.Command:
             help        = "Write the output FHIR documents to stdout. You will likely want to redirect this to a file",
             default     = False)
 
+        # XXX FIXME: I don't think @command_for_project should even be
+        # providing the "cache" parameter.  What is cached and where it is stored is
+        # something specific to each REDCap DET routine, not a global invariant.
+        #   -trs, 19 Dec 2019
+        @click.option("--cache-file",
```
tsibley, line 95: Since this is now part of the user interface, let's give this a more descriptive name? Lots of good choices here. Maybe `--geocode-cache`?

lib/id3c/cli/command/etl/redcap_det.py
```diff
@@ -100,11 +88,21 @@ def decorator(routine: Callable[..., Optional[dict]]) -> click.Command:
...
+            metavar = "<cache.pickle>",
+            help = "Local cache file for storing address lookups. Must be specified to cache lookups",
```
tsibley, line 97: My 2¢.

```suggestion
            help = "Local file for caching the results of address geocoding. Geocoding lookups will not be cached otherwise.",
```

tsibley (changes requested):

kairstenfay:
Thanks for your suggestions, @tsibley ! I agree with all of them and thus incorporated each in a new fixup commit.

tsibley (approved):
LGTM, thanks! I rebased this onto the current state of master and applied the fixup commits.

PR #181, opened 2020-10-28
Changes to support our custom command processing the UW priority queue

tsibley:
Pushing a PR for these changes in advance of my PR for the priority queue command, as I believe they're mostly independently reviewable. Ok by me though if folks would rather wait to review both at the same time.

Details for these changes in the commit messages.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -42,7 +42,7 @@ class Project:
         # Check if project details match our expectations
         self._details = self._fetch("project")
 
-        assert self.id == project_id, \
+        assert int(self.id) == int(project_id), \
             f"REDCap API token provided for project {project_id} is actually for project {self.id} ({self.title!r})!"
 
 
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -30,17 +30,23 @@ class Project:
     *project_id* must match the project id returned in the metadata.  This is a
     safety check that the API token is for the intended project, since tokens
     determine the project accessed.
+
+    If the *dry_run* keyword-only argument is set to ``True``, then methods
+    which could modify data in REDCap will pretend to succeed but not actually
+    make API requests.  Read-only methods are unaffected and will return real
+    data.  Defaults to ``False``.
     """
     api_url: str
     api_token: str
     base_url: str
+    dry_run: bool
     _details: dict
     _instruments: List[str] = None
     _events: List[str] = None
     _fields: List[dict] = None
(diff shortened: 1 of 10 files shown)


lib/id3c/cli/redcap.py
```diff
@@ -337,3 +336,40 @@ def is_complete(instrument: str, data: dict) -> bool:
...
+    ('https://redcap.iths.org/api/', 'https://redcap.iths.org/')
+
+    Assumes that the API endpoint is always at ``…/api/`` under the REDCap
+    instance's base URL and vice versa.  This appears to be true for all the
+    instances inspected (redcap.iths.org, redcap.fhcrc.org,
+    redcapdemo.vanderbilt.edu).
+    """
+    url = Url(url)
+
+    if re.search(r'(?<=/)api/?$', url.path):
```
kairstenfay, line 510: Do you not need to escape the `/` characters in this regular expression?

kairstenfay: I only ask because I entered this into an online regex evaluator, and it suggested that the regex was invalid without escaping them.

kairstenfay (approved):
This looks great, thanks, Tom! I love the Url class you added here.

lib/id3c/cli/redcap.py
tsibley: In other languages, `/` are syntax for delimiting a regex literal, the same way `'` are syntax for delimiting a string literal. In those languages, you do have to escape `/` in a pattern, because the `/` would be treated as a delimiter otherwise. For example, in JS, you'd write this code something like:

```js
if (url.path.match(/(?<=\/)api\/?$/)) {
```

Some languages even let you use different delimiters for regex literals to avoid having to verbosely escape your `/` if you have a lot of them in your pattern.

Even in those languages, you can often also provide your regex patterns as a string, but then they're not regex literals, e.g. in JS:

```js
if (url.path.match('(?<=/)api/?$')) {
```

However, Python doesn't use `/` to delimit regex literals, because it doesn't support regex literals to begin with. You have to provide your regex as a string, as I do here with `r'…'`.

kairstenfay: Thanks, Tom, for laying this out. I appreciate it.

joverlee521 (approved):

tsibley:
Added 4 commits to the end of this branch to support the synthetic DET generation in https://github.com/seattleflu/id3c-customizations/pull/165.

joverlee521 (approved):
Thanks for refactoring the DET creation, makes a lot of sense to move it to the REDCap module!

PR #182, opened 2020-10-30
Accommodate missing metadata fields for Google doc

jccraft-uw:
Don't assume that every field in the metadata of a Google doc will
always exist. In production we just had an occurrence of the
`lastModifyingUserName` not coming back in the details.

### DIFF
--- a/lib/id3c/cli/io/google.py
+++ b/lib/id3c/cli/io/google.py
@@ -89,7 +89,7 @@ def get_document_details(document_id: str) -> dict:
     drive_service = discovery.build("drive", "v2")
     metadata = drive_service.files().get(fileId=document_id).execute()
 
-    details = {'version': metadata['version'], 'etag': metadata['etag'], 'modified_date': metadata['modifiedDate'],
-        'modifiying_user': metadata['lastModifyingUserName']}
+    details = {'version': metadata.get('version'), 'etag': metadata.get('etag'), 'modified_date': metadata.get('modifiedDate'),
+        'modifiying_user': metadata.get('lastModifyingUserName')}
 
     return details


jccraft-uw:
This is a minor fix, and the tests pass, so I'm going to merge and deploy it.

lib/id3c/cli/io/google.py
```diff
@@ -89,7 +89,7 @@ def get_document_details(document_id: str) -> dict:
     drive_service = discovery.build("drive", "v2")
     metadata = drive_service.files().get(fileId=document_id).execute()
 
-    details = {'version': metadata['version'], 'etag': metadata['etag'], 'modified_date': metadata['modifiedDate'],
-        'modifiying_user': metadata['lastModifyingUserName']}
+    details = {'version': metadata.get('version'), 'etag': metadata.get('etag'), 'modified_date': metadata.get('modifiedDate'),
+        'modifiying_user': metadata.get('lastModifyingUserName')}
 
     return details
```
tsibley, line 95: There's only one caller of this function, and that caller only uses the ETag, which it requires exist. This suggests to me that this function could/should just return the ETag value directly, without using `.get(…)`, and not bother even trying to access the other keys of `metadata`. As it stands, if the ETag goes missing, this commit will convert a `KeyError: etag` to `AttributeError: 'NoneType' object has no attribute 'encode'`, which seems less useful on the whole.

jccraft-uw: OK. Changing that now.

jccraft-uw: It's in a new PR. https://github.com/seattleflu/id3c/pull/183

PR #184, opened 2020-10-30
Pin smartystreets-python-sdk at 4.6.1

jccraft-uw:
With 4.7.0 deployed we get this error:
"No module named 'smartystreets_python_sdk.us_reverse_geo"

I was able to run the uw-reopening ETL locally with this version pinned. The lat/log lookup succeeded.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ setup(
         "pyyaml",
         "requests",
         "s3fs",
-        "smartystreets-python-sdk >= 4.0.1",
+        "smartystreets-python-sdk >= 4.0.1, <= 4.6.1",
         "xlrd",
 
         # We use pkg_resources, which (confusingly) is provided by setuptools.


setup.py
```diff
@@ -75,7 +75,7 @@
         "pyyaml",
         "requests",
         "s3fs",
-        "smartystreets-python-sdk >= 4.0.1",
+        "smartystreets-python-sdk <= 4.6.1",
```
tsibley, line 78: While unlikely to be an issue in our environment, this isn't the right pin, since it doesn't preserve the minimum version req. I believe `>= 4.0.1, <= 4.6.1` is the combination of the two.

jccraft-uw: I just force pushed. Does that look right now?

tsibley: LGTM. Please test it, of course.

jccraft-uw:
Deployed.

PR #186, opened 2020-11-06
Barcode work

jccraft-uw:
In my attempt to rebase to master and shuffle recent changes to the same files, I messed up this branch. I'm closing this PR in favor of https://github.com/seattleflu/id3c/pull/189 . It includes feedback from this one.

The commits in this PR make barcode minting about 5x faster than the current minting process. These changes do not reduce the number of minting failures, however. Please see commit messages for details.

Some background about the hamming_distance function: We don't want to put into use a new barcode if it is too similar to an existing barcode. We use the hamming_distance function to measure how different two barcodes are. The function compares the characters in the first barcode and the second barcode; it compares by position ... does barcode1[5] == barcode2[5], for example. The function returns the number of characters that are different. 
Example: select hamming_distance('64ef4b77', '64eb4bd7')   ... this returns 2. 2 of the characters are different between the two barcodes.


jccraft-uw:
[Make minting barcodes faster](https://trello.com/c/MGNkalP4/878-make-minting-barcodes-faster)

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
```diff
@@ -44,8 +46,8 @@ create or replace function warehouse.identifier_barcode_distance_check() returns
                         TG_OP, NEW.barcode),
 
                 detail = format(
-                    'Barcode %L is closer than %s substitutions away from existing barcodes %s',
-                        NEW.barcode, minimum_distance, conflicting_barcodes),
+                    'Barcode %L is closer than %s substitutions away from an existing barcode',
+                        NEW.barcode, minimum_distance),
```
kschwabefry, line 58: It could still be useful to include first_conflict here as an example of what caused the failure.

kschwabefry (commented):
Thanks for doing this Chris, it's exciting to see the timing improvements.

What does the timing look like if we change the slices width from 3 to 2? 
That would allow us to further inspect hamming distance between a barcodes pair like (AAAAAAAA, AA0AA0AA) which wouldn't currently make it past the slice width 3 filtering, but do have a hamming distance of 2, below our current set threshold of 3.

schema/sqitch.plan
```diff
@@ -200,4 +200,6 @@ roles/materialized-view-refresher/grants [roles/materialized-view-refresher/crea
 shipping/age-bin-decade [shipping/schema types/intervalrange] 2020-08-12T22:33:55Z Jover Lee <joverlee@fredhutch.org> # Add table representation of decade age bins
 @2020-08-12 2020-08-12T22:50:50Z Jover Lee <joverlee@fredhutch.org> # Schema as of 12 August 2020
 
+functions/barcode_slices 2020-11-05T17:50:20Z Chris Craft <jccraft@uw.edu> # Add function to create slices out of input text
+warehouse/identifier/slices 2020-11-05T20:05:29Z Chris Craft <jccraft@uw.edu> # Add slices column to warehouse.identifier table
 warehouse/identifier/triggers/barcode-distance-check [warehouse/identifier/triggers/barcode-distance-check@2020-08-12] 2020-11-02T20:56:03Z Chris Craft <jccraft@uw.edu> # Making barcode distance check optimizations
```
joverlee521: FYI, in the past we've tried to match the order between git commits and the sqitch.plan. I can see it being very confusing in the future to see these out of order in the git history.

jccraft-uw: Good point. I combined these into one commit instead of updating the same .sql file across two commits.

jccraft-uw:
> Thanks for doing this Chris, it's exciting to see the timing improvements.
> 
> What does the timing look like if we change the slices width from 3 to 2?
> That would allow us to further inspect hamming distance between a barcodes pair like (AAAAAAAA, AA0AA0AA) which wouldn't currently make it past the slice width 3 filtering, but do have a hamming distance of 2, below our current set threshold of 3.

Good point, @kschwabefry . The "slice" method does require the barcodes to have at least slice_width consecutive characters in common before the hamming distance is checked. When I switched the slice_width to 2, the minting went from 9 to 13 minutes.
Elapsed time: 13 minutes, Minted 1275 identifiers in 2340 tries (1065 retries). 

So that's
No slicing: 59 minutes
slice_width = 3: 9 minutes
slice_width = 2: 13 minutes

Let's talk about this on Monday.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: Added back in fixup commit.

tsibley:
> That would allow us to further inspect hamming distance between a barcodes pair like (AAAAAAAA, AA0AA0AA) which wouldn't currently make it past the slice width 3 filtering, but do have a hamming distance of 2, below our current set threshold of 3.

It's really important that the comparison set reduction criteria doesn't exclude rows which would break the barcode distance invariant.

jccraft-uw:
> > That would allow us to further inspect hamming distance between a barcodes pair like (AAAAAAAA, AA0AA0AA) which wouldn't currently make it past the slice width 3 filtering, but do have a hamming distance of 2, below our current set threshold of 3.
> 
> It's really important that the comparison set reduction criteria doesn't exclude rows which would break the barcode distance invariant. I want to make sure we fully understand the potential for that before any set reduction criteria is deployed.

In fixups I've changed the slice_width from 3 to 2. In one test run last night I see no barcodes that fail the hamming distance < 3 check and do not intersect with the conflicting barcode's slices. That was with a set of 1,000 barcodes that I generated. I'm running now with 3,000. I'm also looking at other pg mechanisms for determining similarity.

jccraft-uw:
> > > That would allow us to further inspect hamming distance between a barcodes pair like (AAAAAAAA, AA0AA0AA) which wouldn't currently make it past the slice width 3 filtering, but do have a hamming distance of 2, below our current set threshold of 3.
> > 
> > 
> > It's really important that the comparison set reduction criteria doesn't exclude rows which would break the barcode distance invariant. I want to make sure we fully understand the potential for that before any set reduction criteria is deployed.
> 
> In fixups I've changed the slice_width from 3 to 2. In one test run last night I see no barcodes that fail the hamming distance < 3 check and do not intersect with the conflicting barcode's slices. That was with a set of 1,000 barcodes that I generated. I'm running now with 3,000. I'm also looking at other pg mechanisms for determining similarity.

For my second test I created a table with 3,000 barcodes (last 8 characters of a new UUID). I have my slice_width set to 2.
This query yielded no results:
select o.barcode, t.barcode, hamming_distance(o.barcode, t.barcode)
from test_barcodes t
cross join warehouse.identifier o
where o.barcode <> t.barcode
and not t.slices && o.slices
and hamming_distance(o.barcode, t.barcode) < 3

...no records for the query of records having no intersection of slices yet having a hamming_distance < 3. 
That's encouraging to show that with slice_width = 2 we are not missing records we'd want to filter out with the hamming_distance.

jccraft-uw:
@seattleflu/id3c-devs , how to speed up the hamming distance check is quite an engineering challenge. See https://dba.stackexchange.com/questions/72134/fast-hamming-distance-queries-in-postgres for one developer's custom postgres indexing extension. 

This article https://stackoverflow.com/questions/9606492/hamming-distance-similarity-searches-in-a-database/47487949#47487949 was my inspiration for the "slices" implementation in this PR. As I noted in one of the replies above, with slice_width set to 2 I did not find (comparing 3,000 barcodes against my local warehouse.identifier table) any where hamming distance < 3 but the slices did not intersect. That plus the perf improvement that it provides (without slices 59 minutes, with slice width = 2, 13 minutes) make me like this approach. The cherry on top is that this approach is easy to explain/understand. 

I've been exploring postgress trigram ideas documented at https://www.postgresql.org/docs/9.6/pgtrgm.html. I'm finding the "distance" between barcodes with `barcode1 <-> barcode2 `. I've looked a the distance compared to the hamming distance between sets of barcodes, and I've come up with the criterion of 0.81 .... use the pre-filter of distance < 0.81 before checking the hamming distance. I haven't run a perf test yet with this implementation, but in a rudimentary query, using distance was faster than doing the slices check. However, what I really don't like is that I picked 0.81 out of a set of barcode runs that I made. That *seems* to be the sweet spot for identifying hamming distance < 3. It seems good, but I don't have confidence to say that this threshold would *never* let barcodes through the have too small of a hamming distance. 

I put the data from my test runs up at https://docs.google.com/spreadsheets/d/1VWtTV_b6hyIsM8npRfGQ_hNMsyamXNy7U-jLyI7kRGg

Minting barcodes is taking a lot of time these days, and we're finding that we can't do much querying while minting is happening. I'd like to proceed with the slices approach (using slice width of 2). It may not be the best or be-all-end-all solution, but it's something for now.

tsibley:
_(We talked about this in standup this morning, but I wanted to briefly repeat myself here for team visibility/future reference.)_

@jccraft-uw I dug into the slices approach yesterday and tinkered with it to get a better sense of its properties and behaviour. I now understand it as a sort of a modified n-gram (or k-mer) approach to use ordered sets instead of unordered sets. I concur that with a slice width of 2, as suggested by @kschwabefry, the slice overlap check should not filter out too many potential rows for distance comparison. It'd be nice to move forward with these slices! (I have just a few minor comments on the implementation which I'll submit in a review.)

> The cherry on top is that this approach is easy to explain/understand.

I think "easy" is relative here. :-) It wasn't clear to me until I spent some time tinkering with it. I think it'd be nice for ourselves in the future and others looking at ID3C to write up this slices approach as part of ID3C's own documention. Would be good to include examples of how the comparison works, the rationale for _why_ it works, its inspiration, its relationship to n-grams/k-mers, and whatever else seems fit.

In addition, I just opened #187 with some smaller (but I think still meaningful) optimizations for barcode minting speed.

tsibley:
@jccraft-uw 

:+1: for rebasing the slices work on top of https://github.com/seattleflu/id3c/pull/187 now that its merged, as you mentioned during standup this morning.

A few additional minor comments on the slices implementation currently in this PR that I'll mention now to maybe save a rebase or two:

- Parallelism: As previously mentioned in discussion, I think it might be nicer to mark `barcode_slices()` parallel-safe and instead control parallelism at the connection/session level with `set max_parallel_workers_per_gather to …` or similar. However, this may negatively affect performance because it'd disable parallelism for `hamming_distance()` too. It'd be nice to understand more what's going on here and what's pathologic, even if we keep `barcode_slices()` marked as parallel unsafe. This might be as minimal as looking at the query plan under various conditions to see the differences.

- Documenting the slices approach, as described in my previous comment, would be really nice. I don't consider this a blocker to merge/deploy, but if it's punted, I fear that it will never happen. :-)

- I noticed that there are hard tabs in the `barcode_slices()` implementation. An easy oversight, but please replace these with four-space tabs to follow the codebase convention.

- The commit message contains a lot of useful explanation and rationale for the changes, which is really nice to see!

jccraft-uw:
> @jccraft-uw
> 
>  for rebasing the slices work on top of #187 now that its merged, as you mentioned during standup this morning.
> 
> A few additional minor comments on the slices implementation currently in this PR that I'll mention now to maybe save a rebase or two:
> 
> * Parallelism: As previously mentioned in discussion, I think it might be nicer to mark `barcode_slices()` parallel-safe and instead control parallelism at the connection/session level with `set max_parallel_workers_per_gather to …` or similar. However, this may negatively affect performance because it'd disable parallelism for `hamming_distance()` too. It'd be nice to understand more what's going on here and what's pathologic, even if we keep `barcode_slices()` marked as parallel unsafe. This might be as minimal as looking at the query plan under various conditions to see the differences.
> * Documenting the slices approach, as described in my previous comment, would be really nice. I don't consider this a blocker to merge/deploy, but if it's punted, I fear that it will never happen. :-)
> * I noticed that there are hard tabs in the `barcode_slices()` implementation. An easy oversight, but please replace these with four-space tabs to follow the codebase convention.
> * The commit message contains a lot of useful explanation and rationale for the changes, which is really nice to see!

Yeah, I destroyed this branch trying to rebase and failing and then trying to start fresh from master. So, see https://github.com/seattleflu/id3c/pull/189 instead.

The parallelism issue was only an issue in the one-time update to populate slices for all warehouse.identifier rows. In that step I've added SET max_parallel_workers_per_gather = 0; 
Now when I run that locally (several times) my machine does not fall over during this step.

I also turned tabs into spaces in the SQL files.

jccraft-uw:
@tsibley , as for documenting the slices approach in the ID3C repo itself, where should I put that ... in the main README, or could I add a README to lib/id3c/cli/command ?
The main README is a very polished overview of the whole project, so the slices level of details doesn't seem to fit there.

tsibley:
> The parallelism issue was only an issue in the one-time update to populate slices for all warehouse.identifier rows. In that step I've added SET max_parallel_workers_per_gather = 0;
Now when I run that locally (several times) my machine does not fall over during this step.

Awesome, thanks!

> as for documenting the slices approach in the ID3C repo itself, where should I put that ... in the main README, or could I add a README to lib/id3c/cli/command ?
> The main README is a very polished overview of the whole project, so the slices level of details doesn't seem to fit there.

I agree the README doesn't seem like the right place. How about as a new document just for the slices stuff in _docs/_ or _dev/docs/_?

jccraft-uw:
> @jccraft-uw
> 
> > The parallelism issue was only an issue in the one-time update to populate slices for all warehouse.identifier rows. In that step I've added SET max_parallel_workers_per_gather = 0;
> > Now when I run that locally (several times) my machine does not fall over during this step.
> 
> Awesome, thanks!
> 
> > as for documenting the slices approach in the ID3C repo itself, where should I put that ... in the main README, or could I add a README to lib/id3c/cli/command ?
> > The main README is a very polished overview of the whole project, so the slices level of details doesn't seem to fit there.
> 
> I agree the README doesn't seem like the right place. How about as a new document just for the slices stuff in _docs/_ or _dev/doc/_?

dev/doc worked great! I put a document there in https://github.com/seattleflu/id3c/pull/189

PR #187, opened 2020-11-12
Optimize barcode distance check

tsibley:
Three independent optimizations to the barcode distance check that in combination reduce the the time spent per identifier from 2.25s to 1.24s (on my local computer over 3 trials of 100 barcodes each).

image

Details in the commit messages. The "limit 1" optimization also exists amongst the larger (and more impactful) "slices" work in #186.


schema/deploy/functions/hamming_distance.sql
```diff
@@ -28,6 +28,34 @@ create or replace function public.hamming_distance(a text, b text)
     immutable
     parallel safe;
 
+create or replace function public.hamming_distance_lte(a text, b text, threshold integer)
```
jccraft-uw, line 31: Because this will not return the actual hamming distance between the two strings if the actual is more than the threshold, it could be better to return a boolean .... it's a flag ... "yes or no, was the hamming distance between the two values greater than the threshold?. If someone wanted the actual hamming distance they could use the original function.

select hamming_distance_lte('AAAAAAAA', 'BBBBBBBB', 3)  returns 4 when the actual hamming distance is 8. 
We really just need to know that yes, this was above threshold.

tsibley: For our distance check, yes, we only need a boolean. For general querying, though, there are times when I've wanted to see the distance of close-enough barcodes. Returning the distance helps in that case because you can compute the distance just once instead of twice.

I agree that the boolean comparison feels slightly clearer than the threshold comparison:

    hamming_distance_lte(a, b, d) <= d
    hamming_distance_lte(a, b, d) is true

…but not clearer enough for me to feel it offsets the double-run downside. I doubt there is a speed difference between the two comparisons.

There's precedent for returning the thresholded distance with the similar [`levenshtein_less_equal()` function](https://www.postgresql.org/docs/10/fuzzystrmatch.html#id-1.11.7.25.6).

tsibley:
For anyone curious, I uploaded the [benchmark data](https://gist.github.com/tsibley/4a63dc48562262508da88d4962b2d116) I collected. For the timings I claim, I averaged over trials 2, 3, and 4. Trial 1 was possibly impacted by unrelated load on my computer during the benchmark, and I wasn't confident in its independence.

schema/deploy/functions/hamming_distance.sql
jccraft-uw: OK. I was coming from the simplicity and clearness perspective.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
```diff
@@ -27,15 +27,16 @@ create or replace function warehouse.identifier_barcode_distance_check() returns
...
         lock table warehouse.identifier in exclusive mode;
 
-        select array_agg(barcode) into strict conflicting_barcodes from (
+        select barcode into conflicting_barcode from (
             select barcode
               from warehouse.identifier
              where hamming_distance_ci(barcode, NEW.barcode) < minimum_distance
             except
             select old_barcode
+             limit 1
```
tsibley: Hmm, I'm now wondering if this is actually correct or not. The `limit 1` may not work as intended depending on how `except` is applied. Looking into this more.

jccraft-uw: I had the same thought, and I saw it working. Let me know what you find.

schema/deploy/functions/hamming_distance.sql
```diff
@@ -39,10 +67,27 @@ create or replace function public.hamming_distance_ci(a text, b text)
     immutable
     parallel safe;
 
+create or replace function public.hamming_distance_ci_lte(a text, b text, threshold integer)
```
jccraft-uw, line 70: Do we use the "ci" version of the function anywhere? If not it should be removed to follow your mantra of don't add it unless it's needed.

lib/id3c/db/__init__.py
```diff
@@ -60,49 +62,53 @@ def mint_identifiers(session: DatabaseSession, name: str, n: int) -> Any:
...
+            failures[m] += 1
+
+            if failures[m] > max_consecutive_failures:
+                LOG.error("Too many consecutive failures, aborting")
+                raise TooManyFailuresError(failures[m])
+            else:
+                LOG.debug("Retrying")
+
+    duration = datetime.now() - started
+    rate = duration.total_seconds() / n
```
jccraft-uw, line 107: When I see 'rate' I think barcodes per second, so I think n / duration.total_seconds().
A variable name like secs_per_barcode or something like that would match better what you're capturing.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
tsibley: Ok, it does work as intended (assuming this is a valid test):

```sql
tom@[local]/tom=# values ('a'), ('b'), ('c') except values ('b') limit 1;
 column1 
─────────
 a
(1 row)

tom@[local]/tom=# values ('a'), ('b'), ('c') except values ('a') limit 1;
 column1 
─────────
 c
(1 row)

tom@[local]/tom=# values ('a'), ('b'), ('c') except values ('b') limit 1;
 column1 
─────────
 a
(1 row)

tom@[local]/tom=# values ('a'), ('b'), ('c') except values ('c') limit 1;
 column1 
─────────
 a
(1 row)

tom@[local]/tom=# values ('x'), ('x'), ('x') except values ('x') limit 1;
 column1 
─────────
(0 rows)

```

lib/id3c/db/__init__.py
tsibley: Yeah, that's fair. I had it as a proper rate initially but then switched to time per barcode and forgot to update the variable name. I'll fix this up.

schema/deploy/functions/hamming_distance.sql
tsibley: Good point. We don't, AFAICT by searching across our repos and within Metabase. I'll remove both "ci" variants.

tsibley: Done in the repush. `hamming_distance_ci_lte` is never added and `hamming_distance_ci` is removed when it's lone callsite is removed.

lib/id3c/db/__init__.py
tsibley: Done in the repush, and both rates are now logged. Look good?

schema/deploy/functions/hamming_distance.sql
jccraft-uw: That's awesome! Thanks for doing that.

lib/id3c/db/__init__.py
jccraft-uw: That's cool that you're logging both. Both will be handy.

jccraft-uw (approved):
Thanks a lot for getting these optimizations in.

tsibley:
Thanks for the review! I'm looking forward to the much larger jump in performance with your slices approach.

PR #188, opened 2020-11-16
Receiving indexes

tsibley:
See commit messages for all the deets.

After deploy, the following should be manually run to clean up my rogue index:

```
drop index receiving.redcap_det_document_jsonb_idx__trs_20200323;
```


jccraft-uw (approved):
These make sense. The index on receiving.redcap_det (first commit) matches the structure of the one we have in prod, and the processing log indicies in commit #2 make sense.
I sqitch deploy'ed and reverted a couple of times. All works nicely.

tsibley:
@jccraft-uw :+1: thanks for the eyes.

Realizing now that I also meant to note that the `processing_log` indexes match the one receiving table where `processing_log` is already indexed as intended: `receiving.enrollment`.

PR #189, opened 2020-11-17
Add barcode minting optimizations

jccraft-uw:
I wrecked the branch for the original PR https://github.com/seattleflu/id3c/pull/186
This PR incorporates the feedback.

Create a new column `slices` on the warehouse.identifier table
to hold 2-character-wide slices of the barcode text. These values are
used to find similar barcodes before running the hamming distance
function.

The changes are:
1. Add the `barcode_slices` function to create the slices array
from the input barcode. See the note in the deploy script regarding
parallelism.

2. Add the `slices` column to the warehouse.identifier table. Update
existing rows to populate `slices` using the `barcodes_slices` function.

3. Update the barcode-distance-check trigger function that is
used when warehouse.identifier rows are created or updated. Add this line
to leverage the slices column before calling the hamming_distance_lte
function: `where slices && NEW.slices`
That is, only call `hamming_distance_lte` on rows that have barcode slices
that intersect with the candidate slices. Previously, `hamming_distance_lte`
was called on all the barcodes in the table.

A compromise I made was to use this same trigger function to populate a
new or updated row's `slices` column as well as to use it to look for
conflicts. I did this to avoid adding the overhead of a new trigger
function that would only set this value. Philosophically it's odd
for a function with "check" in its name to change data in the row being
checked. I hesitated to change the name given our desire to keep names
short.

I couldn't add the `slices` population to the `barcode-default-from-uuid`
function because it is used only for inserts into the warehouse.identifier
table. I needed to ensure that `slices` get updated when rows get updated
as well.

See the schema/verify/functions/barcode_slices.sql file for examples
of what the slices look like.

Minting barcodes is about 4x faster (across multiple tests) with these
changes applied. In the most recent test of minting 1275
barcodes here were the times and the number of conflicting barcodes
encountered. Note that the number of conflicts is about the same
between the runs, indicating that the conflict detection logic is
comparable.

Before changes:
Elapsed time: 59 minutes, Minted 1275 identifiers in 2285 tries (1010 retries)

With changes:
Elapsed time: 15 minutes, Minted 1275 identifiers in 2444 tries (1169 retries)


jccraft-uw:
[Make minting barcodes faster](https://trello.com/c/MGNkalP4/878-make-minting-barcodes-faster)

schema/deploy/warehouse/identifier/slices.sql
```diff
@@ -0,0 +1,13 @@
+-- Deploy seattleflu/schema:warehouse/identifier/slices to pg
+
+begin;
+
+alter table warehouse.identifier
+add column if not exists slices text[];
```
tsibley, line 6: I would remove the `if not exists` here (and the similar `if exists` in the revert). My tationale is that if the column exists when this change is applied, then something has gone wrong, and we'd want to know about it as an error condition, not as a (silent-ish) success.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
```diff
@@ -23,6 +24,11 @@ create or replace function warehouse.identifier_barcode_distance_check() returns
         raise debug 'Checking new barcode "%" is at least % substitutions away from existing barcodes (except old barcode "%")',
             NEW.barcode, minimum_distance, old_barcode;
 
+        -- Populate the row's array of barcode slices to compare with existing rows.
+        -- This greatly reduces the number of barcodes against which the
+        -- `hamming_distance_lte` function is applied.
+        NEW.slices = public.barcode_slices(NEW.barcode);
```
tsibley, line 30: Do we need to think about the value of `OLD.slices` here?

I guess the current behaviour would be that you can never set `slices` yourself, either on insert or update; it's always overwritten by this assignment. Is that right?

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
```diff
@@ -31,8 +37,9 @@ create or replace function warehouse.identifier_barcode_distance_check() returns
 
         select barcode into conflicting_barcode from (
             select barcode
-              from warehouse.identifier
-             where hamming_distance_lte(lower(barcode), new_barcode, minimum_distance-1) < minimum_distance
+            from warehouse.identifier
+            where slices && NEW.slices
+            and hamming_distance_lte(lower(barcode), new_barcode, minimum_distance-1) < minimum_distance
```
tsibley, line 42: Minor nit (which is not merge-blocking): this diff changes the formatting unnecessarily. As a general rule, it's nice to "when in Rome…" by staying consistent with existing formatting in a file, since it minimizes diffs and avoids mixed styles within a single file.

schema/sqitch.plan
```diff
@@ -214,3 +214,7 @@ receiving/presence-absence/indexes/processing-log [receiving/presence-absence] 2
 receiving/redcap-det/indexes/processing-log [receiving/redcap-det] 2020-11-16T20:54:38Z Thomas Sibley <tsibley@fredhutch.org> # Index receiving.redcap_det.processing_log column
 receiving/sequence-read-set/indexes/processing-log [receiving/sequence-read-set] 2020-11-16T20:54:41Z Thomas Sibley <tsibley@fredhutch.org> # Index receiving.sequence_read_set.processing_log column
 @2020-11-17 2020-11-17T18:24:24Z Thomas Sibley <tsibley@fredhutch.org> # Schema as of 17 Nov 2020
+
+functions/barcode_slices 2020-11-17T18:31:45Z Chris Craft <jccraft@uw.edu> # Add function to create slices out of input text
+warehouse/identifier/slices 2020-11-17T18:32:16Z Chris Craft <jccraft@uw.edu> # Add slices column to warehouse.identifier table
+warehouse/identifier/triggers/barcode-distance-check [warehouse/identifier/triggers/barcode-distance-check@2020-11-17] 2020-11-17T18:36:26Z Chris Craft <jccraft@uw.edu> # Add slices check to barcode-distance-check
```
tsibley, line 220: As a dot the i's and cross the t's nit: this last Sqitch change has a dependency on the one before it, `warehouse/identifier/slices`, but that's not declared in the plan. You can just add it manually to the dep list.

schema/deploy/warehouse/identifier/slices.sql
tsibley, line 6: Should `slices` be not null so that if `barcode_slices()` returns null for any reason we don't insert a bad row?

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
```diff
@@ -31,8 +37,9 @@ create or replace function warehouse.identifier_barcode_distance_check() returns
 
         select barcode into conflicting_barcode from (
             select barcode
-              from warehouse.identifier
-             where hamming_distance_lte(lower(barcode), new_barcode, minimum_distance-1) < minimum_distance
+            from warehouse.identifier
+            where slices && NEW.slices
```
tsibley, line 41: If `slices && NEW.slices` returns null because one of them is null, I think this query fails in the wrong direction (potentially doesn't compare distance to anything). Seems like it might be best to guard against that?

dev/doc/01-barcode-slices.md
```diff
@@ -0,0 +1,52 @@
...
+barcode2: a1eeeeee has slices {1__a1,2__1e,3__ee,4__ee,5__ee,6__ee,7__ee}
+The two sets of slices do intersect.
+Because the sets of slices intersect, we should compute the hamming distance.
+The hamming distance is 6.
+
+There is room for improvement here. With only one slice that intersects, we
+could avoid the hamming distance check. However, without an extension, postgres
+does not have a function for returning the members or the number of members
+of two arrays that intersect; the only built in function returns true/false
+indicating whether there is an intersection.
```
tsibley, line 52: Nice observation. This is likely because set intersection is what tables and joins are for, which points to a future path for slices as an auxiliary table that's joined against (i.e. effectively an index instantiated as a explicit table).

dev/doc/01-barcode-slices.md
```diff
@@ -0,0 +1,52 @@
...
+The two sets of slices do intersect.
+Because the sets of slices intersect, we should compute the hamming distance.
+The hamming distance is 2.
+
+(Three)
+barcode1: a1b2c3d4 has slices {1__a1,2__1b,3__b2,4__2c,5__c3,6__3d,7__d4}
+barcode2: a1eeeeee has slices {1__a1,2__1e,3__ee,4__ee,5__ee,6__ee,7__ee}
+The two sets of slices do intersect.
+Because the sets of slices intersect, we should compute the hamming distance.
+The hamming distance is 6.
```
tsibley, line 46: FWIW, this doesn't render well in Markdown: https://github.com/seattleflu/id3c/blob/a03ca5dd4970ca00783b46feddc91f4042a462cf/dev/doc/01-barcode-slices.md

tsibley (commented):
Looks pretty good, thanks for reworking this! I especially appreciate the dev doc; I think it'll be valuable in the future to have that concise, clear explanation.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: I realize that but could not in clear conscious leave the lines unaligned since I was touching one of them.

schema/deploy/warehouse/identifier/slices.sql
jccraft-uw: Agreed. I've changed that in the fixup commit.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: Thanks for seeing this. In the current state someone could update slices with whatever value they want.  I'm now trying various approaches to prevent someone from updating slices in a way that leaves its value not matching barcode_slices(barcode). I really loved the idea of making this a function index instead of a column. (That's such a cool pg feature.) Then we'd have no column to worry about. But pg doesn't use that index for the intersection operation. It uses the index if I say an explicit `where barcode_slices(barcode) = '{blah'` but not `where barcode_slices(barcode) && barcode_slices(NEW.barcode)`. 
Preventing violating updates would be a similar problem if we stored the slices in a separate table; we'd have to make sure that someone couldn't update that table wrong. 
This is not resolved in any commit yet. I'm still tinkering for the cleanest approach.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: I've never seen a FROM indented more than a SELECT in SQL anywhere, but I've restored this back to how it was formatted in the fixup commit.

schema/sqitch.plan
jccraft-uw: Gotcha. I've added that in the fixup. I also added a required to warehouse/identifier/slices.sql ... it needs the function.

schema/deploy/warehouse/identifier/slices.sql
jccraft-uw: Might as well make it not null! I've added that in the fixup .... set it to no null after populating existing rows.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: True true. I've added the null checks in the fixup commit.

dev/doc/01-barcode-slices.md
jccraft-uw: I've added *'s to the examples in the fixup commit.

schema/deploy/warehouse/identifier/slices.sql
jccraft-uw: This is moot now that we're not storing slices in its own column. The "move to index" fixup commit has the new implementation.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: Thanks for the discussion this morning. Using a functional index instead of having a `slices` column in the table gets rid of the operational issues we'd have to deal with around people updating the slices column. And minting is lots faster with the index. I'll add that info as a comment into this PR. When it's time to merge I'll update the commit message with the new numbers.

schema/sqitch.plan
jccraft-uw: In the new index implementation both the index and the barcode-distance-check depend on the function. I've added that in for those 2.

schema/deploy/warehouse/identifier/slices.sql
jccraft-uw: Moot now with the index (instead of `slices` column) implementation.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: See what you think about what I've done in the 'move to index' fixup commit. I'm handling the case where barcode_slices(barcode) is an empty array. When I looked for `barcode_slices(barcode) is null` this caused pg to not use the index.
The barcode_slices function returns null if the input text is null. If the barcode is null in an identifier row we don't need to check the hamming distance against it.

jccraft-uw:
Here's the run time comparison using a GIN index and not a dedicated `slices` column. (I'll update the commit message with this info before I merge.)

Code as in Master:
Minted 1275 identifiers in 2577 tries (1302 retries) over 0:46:07.046137 (2.17 s/identifier = 0.46 identifiers/s)

With the GIN index in this PR ('move to index' fixup commit):
Minted 1275 identifiers in 2634 tries (1359 retries) over 0:02:02.502329 (0.10 s/identifier = 10.41 identifiers/s)

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
tsibley: Ah, I swear it's a style used by people other than me. :-) In any case, it's one [prevailing style within ID3C](https://github.com/search?q=select+OR+where++repo%3Aseattleflu%2Fid3c+language%3APython+language%3APython&type=Code&ref=advsearch&l=Python&l=Python).

schema/sqitch.plan
tsibley: Ah, I see the informational `-- requires: …` comments you added. Those are good, but are informational only (e.g. for humans only). Sorry, I should have been clearer: the place that matters is in _schema/sqitch.plan_ within the square brackets, i.e. the line I made my original review comment on.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
tsibley: Hmm. So the new condition is:

```
(barcode_slices(barcode) = '{}' or barcode_slices(barcode) && barcode_slices(NEW.barcode))
```

but I don't think this covers the null edge case I raised.

1. It doesn't catch cases where `barcode_slices(NEW.barcode)` is null, which could happen for a variety of reasons. The whole expression would evaluate as null in that case and thus skip the distance calc.

2. The `barcode_slices()` function never returns an empty array: it either returns null or an array with values. So `barcode_slices(barcode) = '{}'` will always be false.

My suggestion is to do this instead:

```
barcode_slices(barcode) && barcode_slices(NEW.barcode) is not false
```

which will test Hamming distance if the slices overlap or if (for whatever reason) either the LHS or RHS is null. I expect this still uses the index, but it would be good to double check that empirically.

Looking at this again right now, I'm also realizing that `barcode_slices()` likely wants to be case-insensitive as well, otherwise barcodes may be inappropriately skipped due to case differences alone. My suggestion is to switch the input and output types of `barcode_slices()` from `text` to `citext`, as the function is specifically intended for `warehouse.identifier.barcode` which is `citext`. (There are perf. reasons why the Hamming distance functions don't use `citext`, but since `barcode_slices()` is pre-computed by the index, those same reasons don't apply.)

schema/sqitch.plan
jccraft-uw: Ah, sorry about that. I've added the dependencies in the sqitch.plan too.

schema/deploy/warehouse/identifier/triggers/barcode-distance-check.sql
jccraft-uw: I've changed the function to take/return citext. I also renamed the index to end in "_idx" as is customary.

I couldn't come up with a way to make `barcode_slices(barcode) && barcode_slices(NEW.barcode) is not false` use the index. What I did instead was to make two code paths .... if the slices on the new barcode is null don't use the intersection shortcut; if slices on the new barcode is not null do use the shortcut. This means duplicating most of the logic, but it lets the happy path use the index. It also lets us add other conditions to the unhappy path if we need them.

This does not cover the case when barcode_slices(barcode) is null for some rows in the table. This would happen if the barcode column itself is null. In that case it's fine not to compare against it. 
Do you see other scenarios when barcode_slices(barcode) could be null?

jccraft-uw: Meh, I don't like what I added. I couldn't use what you suggested `barcode_slices(barcode) && barcode_slices(NEW.barcode) is not false` because that doesn't use the index, so I tried to cover one scenario. That seems like a lot of copied code for little value.

So, what are the scenarios that would lead to null slices? If the barcode is null then the slices will be null for sure. What other scenarios are we wanting to look for?

tsibley: Ah, thanks for testing it out. Sorry the suggestion wasn't workable. I think a "happy path" and "grudging path" are ok, but I agree it'd be nice to avoid if we can since duplicate code paths allow for the possibility of bugs due to drift, which is bad for something like unique identifier control. On the other hand, it's nice to cover all edge cases in important code like this so that they can't bite us later.

`barcode_slices()` returns null if the input text is a) null or b) shorter than the slice width (currently 2). (a) should be prevented by the `not null` constraint on `warehouse.identifier.barcode` and (b) should be prevented by the `length = 8` check constraint on `warehouse.identifier.barcode`. But as both of those constraints are independent of the distance check (e.g. could be removed without the distance check knowing), it seems most safe for the distance check to not rely on them. However, this may not be the most pragmatic choice given other considerations... but also read on.

One option to avoid the duplication but also handle the null edge cases is to revert the `if new_slices is null` conditional and change the query criteria in the trigger to

```
-- Use barcode slices to reduce the comparison set for pairwise
-- distance calculations.  If slices can't be computed/are missing for
-- either side, then always check the distance out of an abundance of
-- precaution.
(barcode_slices(barcode) && new_slices or barcode_slices(barcode) is null or new_slices is null)
```

The `is null` check can then be made fast by adding an additional btree index, e.g.:

```
create index identifier_barcode_slices_btree_idx on warehouse.identifier using btree (barcode_slices(barcode));
```

The new index will take up additional space, of course, but on my local instance it's only a tiny ~64MB for the whole thing. To reduce the size, we could even further target the index on the `is null` expression:

```
create index identifier_barcode_slices_is_null_idx on warehouse.identifier using btree ((barcode_slices(barcode) is null));
```

which is only ~10MB.

<details><summary>Example query plan (with the first index above)</summary>

```
2020-12-08 12:35:49.247 PST [27379] tom@seattleflu LOG:  duration: 54.470 ms  plan:
        Query Text: select barcode                          from (
                    select barcode
                      from warehouse.identifier
                     where (barcode_slices(barcode) && new_slices or barcode_slices(barcode) is null or new_slices is null)
                       and hamming_distance_lte(lower(barcode), new_barcode, minimum_distance-1) < minimum_distance
                    except
                    select old_barcode
                     limit 1
                ) x  
        Subquery Scan on x  (cost=1372.32..1374.05 rows=1 width=32)
          ->  Limit  (cost=1372.32..1374.04 rows=1 width=36)
                ->  HashSetOp Except  (cost=1372.32..12061.48 rows=6225 width=36)
                      ->  Append  (cost=1372.32..12045.92 rows=6226 width=36)
                            ->  Subquery Scan on "*SELECT* 1"  (cost=1372.32..12014.77 rows=6225 width=13)
                                  ->  Gather  (cost=1372.32..11952.52 rows=6225 width=9)
                                        Workers Planned: 2
                                        ->  Parallel Bitmap Heap Scan on identifier  (cost=372.32..10330.02 rows=2594 width=9)
                                              Recheck Cond: ((barcode_slices(barcode) && '{1__a5,2__5d,3__de,4__e7,5__79,6__9e,7__ef}'::citext[]) OR (barcode_slices(barcode) IS NULL))
                                              Filter: (hamming_distance_lte(lower((barcode)::text), 'a5de79ef'::text, 2) < 3)
                                              ->  BitmapOr  (cost=372.32..372.32 rows=18755 width=0)
                                                    ->  Bitmap Index Scan on identifier_barcode_slices_idx  (cost=0.00..182.85 rows=16380 width=0)
                                                          Index Cond: (barcode_slices(barcode) && '{1__a5,2__5d,3__de,4__e7,5__79,6__9e,7__ef}'::citext[])
                                                    ->  Bitmap Index Scan on __tmp_test  (cost=0.00..186.36 rows=2375 width=0)
                                                          Index Cond: (barcode_slices(barcode) IS NULL)
                            ->  Subquery Scan on "*SELECT* 2"  (cost=0.00..0.02 rows=1 width=36)
                                  ->  Result  (cost=0.00..0.01 rows=1 width=32)
```

</details>

jccraft-uw: @tsibley I went with adding the smaller index to capture whether the slices is null.
With that index created, all was working great except for the `or new_slices is null` part. With that check, the operation was slow again. Ah, it looks like PG has issues with `limit 1`. See https://stackoverflow.com/questions/21385555/postgresql-query-very-slow-with-limit-1/27237698 and maybe https://www.postgresql.org/message-id/flat/CA%2BU5nMLbXfUT9cWDHJ3tpxjC3bTWqizBKqTwDgzebCB5bAGCgg%40mail.gmail.com.

I ended up reverting the change we made a few weeks ago that added `limit 1`. By reverted I mean that I copy and pasted from an earlier commit that had this in there. I see it working correctly:
```ERROR:  INSERT of (barcode='5220aa94') excluded by minimum substitution distance check
DETAIL:  Barcode '5220aa94' is closer than 3 substitutions away from existing barcodes {5b20aa97,5626aa94,52b03a94,522f1a94}
HINT:  If the INSERT relied on default values, retrying it will use a new value and possibly succeed
CONTEXT:  PL/pgSQL function warehouse.identifier_barcode_distance_check() line 38 at RAISE
SQL state: 23P01```

I've updated the commit message with all the changes and the results from my runs today.
I think we're looking good with those two null checks in there working fast.

tsibley: Thanks for digging into this. The pathologic query planner behaviour possible around small limits not using indexes is really interesting and useful to know about.

tsibley (approved):
Thanks for putting in all the time to dig into this and follow through with testing all the variations. This will be a really nice, robust improvement to an important core part of the system. Please feel free to merge and deploy this at will.

There's some small formatting sloppiness in the changes to the trigger that increase noise in the diff (e.g. lines with unnecessary changes), but I don't think they're enough to warrant requesting changes. I mention them here only because I think reducing diff noise is best practice and something to keep in mind. Using partial `git revert`s (e.g. of 8d27a7928b43e7b0665eada7fcf2b5579f67ff99) can help with this in cases of rolling back changes from a previous commit.

jccraft-uw:
> Thanks for putting in all the time to dig into this and follow through with testing all the variations. This will be a really nice, robust improvement to an important core part of the system. Please feel free to merge and deploy this at will.
> 
> There's some small formatting sloppiness in the changes to the trigger that increase noise in the diff (e.g. lines with unnecessary changes), but I don't think they're enough to warrant requesting changes. I mention them here only because I think reducing diff noise is best practice and something to keep in mind. Using partial `git revert`s (e.g. of [8d27a79](https://github.com/seattleflu/id3c/commit/8d27a7928b43e7b0665eada7fcf2b5579f67ff99)) can help with this in cases of rolling back changes from a previous commit.

I removed a stray empty line that I introduced. It's now time to merge and deploy !!!!

jccraft-uw:
Deployed !

PR #192, opened 2020-12-01
etl presence-absence: skip samples that have been marked as failed

joverlee521:
Each sample has a top level key "sampleFailed" that is marked True if
the entire sample has been marked as failed. We need to inspect this
top level key because the individual target results do not reflect this
failure status.

ID3C currently does not model failed presence/absence results,
so skip all results if the sample is failed at this top level.

### DIFF
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -102,6 +102,12 @@ def etl_presence_absence(*, db: DatabaseSession):
                     LOG.info(f"Skipping sample «{received_sample['sampleId']}» without SFS barcode")
                     continue
 
+                # Don't go any further if the sample is marked as Failed
+                sample_failed = received_sample.get("sampleFailed")
+                if sample_failed is True:
+                    LOG.info(f"Skipping sample «{received_sample_barcode}» that has been failed")
+                    continue
+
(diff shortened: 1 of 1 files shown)


joverlee521:
[Deal with failed OpenArray sample results](https://trello.com/c/BVDUdZND/913-deal-with-failed-openarray-sample-results)

jccraft-uw (approved):

joverlee521:
After this PR is merged, we should be able to clean up the `warehouse.presence_absence` table with the following:
```sql
begin;

delete from warehouse.presence_absence
where presence_absence_id in (
    select
        presence_absence.presence_absence_id
    from (
        select
            concat_ws('/', 'NWGC', s."sampleId", t."geneTarget", s."chip") as potential_identifier
        from receiving.presence_absence,
        json_to_recordset(document -> 'samples') as s("sampleFailed" bool, "sampleId" text, "chip" text, "targetResults" json),
        json_to_recordset("targetResults") as t("geneTarget" text)
        where "sampleFailed" is true
    ) as failed_samples
    join warehouse.presence_absence on presence_absence.identifier = failed_samples.potential_identifier
);

```
(This currently matches 16482 presence_absence records).

tsibley (approved):
LGTM by inspection.

kschwabefry (approved):

PR #194, opened 2020-12-23
Labelmaker/smaller clia labels

kairstenfay:
Support smaller CLIA barcode layouts as an option. See commits for details

### DIFF
--- a/lib/id3c/cli/command/identifier.py
+++ b/lib/id3c/cli/command/identifier.py
@@ -46,6 +46,11 @@ def identifier():
     metavar = "<file.pdf>",
     type = click.File("wb"))
 
+@click.option("--layout",
+    metavar = "<layout>",
+    default = 'default',
+    help = "Specify a desired label layout.")
+
 @click.option("--quiet", "-q",
     help = "Suppress printing of new identifiers to stdout",
     is_flag = True,
@@ -56,7 +61,8 @@ def identifier():
     is_flag = True,
     flag_value = True)
 
-def mint(set_name, count, *, labels, quiet, dry_run):
+
+def mint(set_name, count, *, labels, layout, quiet, dry_run):
     """
     Mint new identifiers and make barcode labels.
 
@@ -69,6 +75,9 @@ def mint(set_name, count, *, labels, quiet, dry_run):
     using the Lab Labels¹ instance <https://backoffice.seattleflu.org/labels/>.
     An alternative instance may be used by setting the LABEL_API environment
     variable to the instance URL.
+
+    If --layout is requested, the printable barcode labels will use the given
+    version of the layout, if available.
 
     ¹ https://github.com/MullinsLab/Lab-Labels
     """
@@ -86,8 +95,8 @@ def mint(set_name, count, *, labels, quiet, dry_run):
             print(identifier.barcode, identifier.uuid, sep = "\t")
 
     if labels:
-        layout = labelmaker.layout_identifiers(set_name, minted)
-        pdf = labelmaker.generate_pdf(layout)
+        label_layout = labelmaker.layout_identifiers(set_name, minted, layout)
+        pdf = labelmaker.generate_pdf(label_layout)
         labels.write(pdf)
 
 
@@ -97,7 +106,13 @@ def mint(set_name, count, *, labels, quiet, dry_run):
     metavar = "<file.pdf>",
     type = click.File("wb"))
 
-def labels(filename):
+@click.option("--layout",
+    metavar = "<layout>",
+    default = 'default',
+    help = "Specify a desired label layout.")
+
+
+def labels(filename, layout: str='default'):
     """
     Make barcode labels for an existing batch of identifiers.
 
@@ -108,6 +123,9 @@ def labels(filename):
 
     The batch of identifiers to make labels for is selected interactively based
     on the identifier set and time of original generation.
+
+    If --layout is requested, the printable barcode labels will use the given
+    version of the layout, if available.
 
     ¹ https://github.com/MullinsLab/Lab-Labels
     """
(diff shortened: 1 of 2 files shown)


kairstenfay:
[Make smaller CLIA barcode labels](https://trello.com/c/UTqZlvxW/906-make-smaller-clia-barcode-labels)

lib/id3c/cli/command/identifier.py
```diff
@@ -46,6 +46,11 @@ def identifier():
     metavar = "<file.pdf>",
     type = click.File("wb"))
 
+@click.option("--small",
```
jccraft-uw, line 49: Could we make this more generic, for example, make the option "--layout" to choose which layout to use? This use case has us going with the small layout, but we may have other scenarios that have us use multiple layouts for a single collection set. If --layout is not specified use the default.

lib/id3c/labelmaker.py
```diff
@@ -208,10 +208,28 @@ class CollectionsScanKiosksLayout(LabelLayout):
 
 
 class CollectionsCliaComplianceLayout(LabelLayout):
```
jccraft-uw, line 210: If we go more generic we would have a second layout class for Clia.

lib/id3c/cli/command/identifier.py
kairstenfay: Interesting. Were you thinking `--layout` would be the Cryo layout specification, something like `--layout LCRY-2380`? In that approach, we'd also need to specify options like `--no-barcode-text` or `--no-barcode-name`. I'm not sure how useful those latter options would be.

lib/id3c/labelmaker.py
kairstenfay: How were you imagining invoking a second layout class?

jccraft-uw: I was thinking we would pass in for the --layout option the name of the layout class itself.

lib/id3c/cli/command/identifier.py
kschwabefry: A hybrid approach could be to use `--layout small` (or `compact`? `minified`?) as the option to toggle both the label size and whether text is included. `--layout` could be expanded to support other layouts in the future.

lib/id3c/cli/command/identifier.py
```diff
@@ -70,6 +76,9 @@ def mint(set_name, count, *, labels, quiet, dry_run):
     An alternative instance may be used by setting the LABEL_API environment
     variable to the instance URL.
 
+    If --small is requested, the printable barcode labels will use the "small"
+    version of the layout, if available.
```
kschwabefry, line 80: To be more upfront and clear with the user, I would be inclined to have it throw a `notimplemented` if this option is requested where its not supported yet rather than silently use the normal label.

lib/id3c/labelmaker.py
kairstenfay: I have some reservations about evaluating a string literally to instantiate a python class. I would personally prefer to create multiple, callable layouts within a barcode class (e.g. small, compact, or minified) as suggested by @kschwabefry in a different thread in this PR.

lib/id3c/cli/command/identifier.py
kairstenfay: I like the hybrid approach of declaring a `--layout` type with a common name (like small/compact/minified etc.). I also see how creating a `--layout` option could more easily be expanded to support other layouts in the future, but I'm concerned about the "YAGNI" philosophy @tsibley often raises. How likely are we to have to support future layouts?

kschwabefry: I would weigh two factors: 
- YAGNI and the likeliness of future alternate layouts: I don't know of specific plans (but it reasonably could happen)
- the difficulty to change it later: in this case, refactoring it later seems doable, and we are the users of this CLI which gives us a little more flexibility.

Based on the above understanding, `--small` is ok with me for now, with the assumption we would likely refactor it if we do need another alternate layout later. (A comment to that effect, either in the code or in the commit message, may be a useful reminder of that intention)

Since he raised the point initially, I want to give @jccraft-uw a chance to weigh back in here as well.

jccraft-uw: Thanks for taking a look, @kschwabefry . In the interest of getting this out sooner than later, I agree with going with `--small`. If we need additional layouts in the future we can refactor the code then.

kairstenfay: What about an option where we offer `--layout` with a default value of `default`, and then this PR implements a `--layout` value of `small`?

And I'll implement Kristen's other suggestion to raise a NotImplementedError if `small` or some other non-supported layout is specified but there's nothing implemented for that class.

jccraft-uw: `--layout` sounds really good. We're specifying which layout we want. For now, only CLIA would have a layout other than `default`, `small`.
This is what I had in mind--specify the layout that you want when you're minting.

kschwabefry: Sounds like a good resolution. Thanks all.

kairstenfay: Thanks both for your feedback. Please take a look at the latest commit. The only thing I'm not sure of is if we want to maintain the layout choices in two places: one in the Click option in the `identifier` command, and the other in the set of implemented layouts for each barcode layout in `labelmaker`.

lib/id3c/cli/command/identifier.py
kairstenfay: This is now implemented.

lib/id3c/cli/command/identifier.py
jccraft-uw: I would not add the possible options in the Click option because it would take work to keep that up to date if we add layouts in labelmaker.

kairstenfay:
Ready for re-review.

I just re-pushed to fix a bug that was breaking the CI builds. I also removed the `Click.choice` type on the `identifier` command's `--layout` option so we don't have to maintain the set of layouts in multiple places (only `labelmaker`).

cc @jccraft-uw

lib/id3c/labelmaker.py
jccraft-uw: Going with the --layout option is good.

jccraft-uw (approved):
The KitFab team will appreciate this work!

jccraft-uw (approved):
I didn't try this out locally, but the code changes look good.

kairstenfay:
I just re-pushed to make the wording in the `identifier` doctstrings about `--layout` more generic (rather than describing usage of the "small" option).

kschwabefry (approved):

PR #195, opened 2020-12-29
etl presence-absence: Check target clinicalStatus

kairstenfay:
Update the presence-absence ETL to check a target's clinical status
(`clinicalStatus`).

Per @joverlee521:
    Individual targets of the OpenArray will be CLIA validated (~mid
    January 2021?). We will be receiving target.clinicalStatus that
    indicates whether or not a result is "Clia" or "Research". We need
    to ingest this key for individual results so that we can return
    SARS-CoV-2 OpenArray results when it gets validated.

    Release plan:
    1. Samplify send OA results with sample.assayType = "Research" (no
    changes to ID3C ETL required)
    2. Samplify adds target.clinicalStatus to the JSONs for both OA and
    qPCR
    3. ID3C updates ETL to ingest target.clinicalStatus instead of the
    sample.assayType
    4. Samplify removes the sample.assayType key

    Ideally all steps are completed before the OA SARS-CoV-2 is
    validated so that the results can be returned to participants.

### DIFF
--- a/lib/id3c/cli/command/etl/presence_absence.py
+++ b/lib/id3c/cli/command/etl/presence_absence.py
@@ -120,6 +120,10 @@ def etl_presence_absence(*, db: DatabaseSession):
                 extraction_date = received_sample.get("extractionDate")
                 assay_name = received_sample.get("assayName")
                 assay_date = received_sample.get("assayDate")
+                # The assayType field will be removed after Samplify starts
+                # sending us OpenArray results with target.clinicalStatus.
+                #
+                # kfay, 28 Dec 2020
                 assay_type = received_sample.get("assayType")
 
                 # Guard against empty chip values
@@ -313,6 +317,12 @@ def presence_absence_details(document: dict,
     Raises `AssertionError` if we find an unknown *assay_name* or unknown
     *assay_type*.
     """
+    # Look for clinical status -- historically, this detail was stored at the
+    # top level, but with the new clinical validation of OpenArray results, we
+    # may have a mix of clinial and research results in one upload. Fall back to
+    # the assayType if clinicalStatus is not available for this target.
+    assay_type = document.get('clinicalStatus') or assay_type
+
     device = None
 
     if assay_name:
@@ -324,9 +334,19 @@ def presence_absence_details(document: dict,
     if assay_type:
         assert assay_type in {'Clia', 'Research'}, f"Found unknown assay type «{assay_type}»"
     else:
-        # Assumes anything with 4 wellResults is "Clia" and everything else
-        # "Research" assays
-        assay_type = 'Clia' if len(document['wellResults']) == 4 else 'Research'
+        # 7 April 2020 was the date we first started ingesting `assay_type`.
+        if extraction_date < '2020-04-07':
+            # Assumes anything with 4 wellResults is "Clia" and everything else
+            # "Research" assays
+            assay_type = 'Clia' if len(document['wellResults']) == 4 else 'Research'
+
+        # The "4 well assumption" used for CLIA results may not apply to
+        # OpenArray results. If no assay type is given, default to 'Research' to
+        # avoid accidentally reporting a non-clinical result to participants.
+        else:
+            LOG.warning("No assay type found for target. Defaulting to «Research».")
+            assay_type = 'Research'
+
 
     return {
         "device": device,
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Update presence/absence ETL to check target clinicalStatus](https://trello.com/c/N6iLC1HU/925-update-presence-absence-etl-to-check-target-clinicalstatus)

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -175,6 +179,13 @@ def etl_presence_absence(*, db: DatabaseSession):
                     else:
                         identifier = f"NWGC/{received_sample_id}/{target.identifier}"
 
+                    # Look for clinical status -- historically, this detail was
+                    # stored at the top level, but with the new clinical
+                    # validation of OpenArray results, we may have a mix of
+                    # clinial and research results in one upload. Fall back to
+                    # the assayType if clinicalStatus is not available.
+                    clinical_status = test_result.get('clinicalStatus') or assay_type
```
joverlee521, line 187: It might be easier for future debugging if the `assay_type` value gets determined in a single place. Could this line be moved to within `presence_absence_details()`?

joverlee521 (commented):
Thanks for making the change @kairstenfay! Just a couple suggestions:

I think we might want to add a date cutoff for the `assay_type` fallback within `presence_absence_details()` since the 4 well assumption for CLIA results may not apply for the OA results. 


Specifically talking about this if/else statement within `presence_absence_details()`: 
```python
if assay_type:
        assert assay_type in {'Clia', 'Research'}, f"Found unknown assay type «{assay_type}»"
else:
        # Assumes anything with 4 wellResults is "Clia" and everything else
        # "Research" assays
        assay_type = 'Clia' if len(document['wellResults']) == 4 else 'Research'
```
Not entirely sure what the best date cutoff would be. We could use the date when `assay_type` was first ingested back in [April](https://github.com/seattleflu/id3c/pull/141) assuming that all TaqmanQPCR results we've received since then have the `assayType` key in the JSON.

kairstenfay:
> Thanks for making the change @kairstenfay! Just a couple suggestions:
> 
> I think we might want to add a date cutoff for the `assay_type` fallback within `presence_absence_details()` since the 4 well assumption for CLIA results may not apply for the OA results.
> 
> Specifically talking about this if/else statement within `presence_absence_details()`:
> 
> ```python
> if assay_type:
>         assert assay_type in {'Clia', 'Research'}, f"Found unknown assay type «{assay_type}»"
> else:
>         # Assumes anything with 4 wellResults is "Clia" and everything else
>         # "Research" assays
>         assay_type = 'Clia' if len(document['wellResults']) == 4 else 'Research'
> ```
> 
> Not entirely sure what the best date cutoff would be. We could use the date when `assay_type` was first ingested back in [April](https://github.com/seattleflu/id3c/pull/141) assuming that all TaqmanQPCR results we've received since then have the `assayType` key in the JSON.

Thanks for this feedback. without an `assay_type` fallback for OA results (assuming we implement a date cutoff), would we want to populate the `assay_type` field at all or leave it empty?

joverlee521:
> Thanks for this feedback. without an assay_type fallback for OA results (assuming we implement a date cutoff), would we want to populate the assay_type field at all or leave it empty?

If future results don't have `assayType` or `clinicalStatus` I think we can either throw an error or log a warning about it and default to "Research" so that we never accidentally report a non-clinical result to participants.

kairstenfay:
> > Thanks for this feedback. without an assay_type fallback for OA results (assuming we implement a date cutoff), would we want to populate the assay_type field at all or leave it empty?
> 
> If future results don't have `assayType` or `clinicalStatus` I think we can either throw an error or log a warning about it and default to "Research" so that we never accidentally report a non-clinical result to participants.

It looks like we still receive a lot of results with no `assayType` after 2020-04-07. Please check my latest version of the commit for the if/else block I added to make sure I understood you correctly. I am doing a run locally right now and it's pretty noisy.

joverlee521:
> It looks like we still receive a lot of results with no `assayType` after 2020-04-07. Please check my latest version of the commit for the if/else block I added to make sure I understood you correctly. I am doing a run locally right now and it's pretty noisy.

Ah right, it's noisy because the OpenArray results currently do not have `assayType`. Hmm, the date cutoff might have to be the future date when we first receive OpenArray results with the `assayType`/`clinicalStatus`.

kairstenfay:
@joverlee521 Is this PR only on hold because we don't know the date of when we receive the new OpenArray results? And once we know that date, we can replace `2020-04-07` in the filter with it and merge?

kschwabefry (approved):

jccraft-uw (approved):

joverlee521:
Deployed

PR #197, opened 2021-01-08
REDCap DET generation

tsibley:
Switches `id3c redcap-det generate` to use the new-style env vars for API credentials and significantly reduces the amount of data transferred over the wire in the API requests.

See commit messages for details.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -405,7 +405,7 @@ def is_complete(instrument: str, data: dict) -> bool:
     >>> is_complete("test", {}) is None
     True
     """
-    instrument_complete_field = data.get(f"{instrument}_complete")
+    instrument_complete_field = data.get(completion_status_field(instrument))
 
     if instrument_complete_field is None:
         return None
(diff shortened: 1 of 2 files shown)


tsibley:
I authored the bulk of these commits as a wip a month ago and was reminded to come back to them today.

lib/id3c/cli/redcap.py
```diff
@@ -417,6 +417,28 @@ def is_complete(instrument: str, data: dict) -> bool:
...
+    #     single underscore.
+    #   - Maybe: removing leading numbers?
+    #
+    # The full set of transformations REDCap applies aren't entirely clear to
+    # me at the moment, so I'm punting for now.  The caller must provide the
+    # internal names.
+    #   -trs, 8 Jan 2020
+    return f"{instrument}_complete"
+
+
```
kairstenfay, line 441: I am a little unclear on the added value of this `completion_status_field` and what you hope to one day achieve in the TODO statement. Would you mind elaborating, please?

tsibley: Sure.

The reason for the function is to centralize and abstract away the transformation, instead of leaving it increasingly sprinkled throughout the code, so that it's easier to update later. Yes, the transformation is trivial right now, but, for example, the `--instrument` option for `id3c redcap-det generate` takes user-provided values and those may not be valid without normalization, which this function could do.

Is this abstraction necessary right now? Nope. But it was ~no extra work, and IMO accomplishes two important things:

1. Increases ease of future changes by centralizing details
2. Improves readability, because now the code talks about _meaningful concepts_ like "completion status" instead of being a list of operations (like "append `_complete` to the name of this instrument")

kairstenfay: Ah, I see. When you mentioned normalizing `instrument`, for some reason I was thinking that was the field name in REDCap rather than the user provided value. Also, your listed reasons make sense to me. Thanks!

kairstenfay (approved):

lib/id3c/cli/redcap.py
```diff
@@ -410,6 +417,28 @@ def is_complete(instrument: str, data: dict) -> bool:
...
+    # XXX TODO: It would be good to normalize *instrument* here, including:
+    #
+    #   - Lowercasing
+    #   - Replacing runs of whitespace and/or non-alphanumerics (?) with a
+    #     single underscore.
+    #   - Maybe: removing leading numbers?
+    #
+    # The full set of transformations REDCap applies aren't entirely clear to
+    # me at the moment, so I'm punting for now.  The caller must provide the
+    # internal names.
```
kschwabefry, line 437: That's fair. All I see in the redcap docs is the replacement of whitespace with underscore, but it would make sense to run some tests to see if there are other changes.

kschwabefry (approved):

lib/id3c/cli/redcap.py
tsibley: Yeah, I did run some tests and enumerated my findings in the preceding lines, but I didn't want to invest more time once it became clear that the full behaviour isn't documented by REDCap.

PR #199, opened 2021-01-12
redcap: Support paged fetching of records

tsibley:
For very large projects, such as our Husky Coronavirus Testing project,
paging allows record resultsets to fit within the memory/timeout
constraints of the API server.  REDCap seems to spool the whole
resultset (into memory?) before sending the response.  When constraints
are exceeded, we see HTTP 500 errors from the API.

---

This will enable fixing the UW EH&S export, by adding, e.g., `page_size = 5000` to the relevant `project.records(…)` call.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -9,7 +9,7 @@ import requests
 from enum import Enum
 from functools import lru_cache
 from operator import itemgetter
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 from .utils import running_command_name
 from ..url import Url
 
@@ -155,7 +155,9 @@ class Project:
         fields ``redcap_event_name``, ``redcap_repeat_instrument``, and
         ``redcap_repeat_instance``.
         """
-        return self.records(ids = [record_id], raw = raw)
+        # This is always a List[Record] because we don't pass page_size, but
+        # mypy (0.790) can't figure that out.
+        return self.records(ids = [record_id], raw = raw) # type: ignore
 
 
     def records(self, *,
@@ -165,7 +167,8 @@ class Project:
                 fields: List[str] = None,
                 events: List[str] = None,
                 filter: str = None,
-                raw: bool = False) -> List['Record']:
+                raw: bool = False,
+                page_size: int = None) -> Union[List['Record'], Iterator['Record']]:
         """
         Fetch records for this REDCap project.
 
@@ -196,6 +199,17 @@ class Project:
         The optional *raw* parameter controls if numeric/coded values are
         returned for multiple choice fields.  When false (the default), string
         labels are returned.
+
+        The optional *page_size* parameter, when set, enables paged fetching of
+        records by their primary record id (i.e. :attr:`.record_id_field`).
+        *page_size* can only be used with REDCap projects that use record
+        auto-numbering.  Each fetch will include a maximum of *page_size*
+        records, although may contain many more result rows for records with
+        repeating instruments/events).  The last fetch may contain results
+        for more than *page_size* records because it uses an unrestricted upper
+        bound in order to catch anything created since the start of the
+        pagination process.  When *page_size* is provided, an iterator, as
+        opposed to a list, is returned.
         """
         parameters = {
             'type': 'flat',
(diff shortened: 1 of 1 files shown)


lib/id3c/cli/redcap.py
```diff
@@ -196,6 +197,18 @@ def records(self, *,
         The optional *raw* parameter controls if numeric/coded values are
         returned for multiple choice fields.  When false (the default), string
         labels are returned.
+
+        The optional *page_size* parameter, when set, enables paged fetching of
```
jccraft-uw, line 201: I think you're saying that in order to use the `page_size` parameter, the REDCap project must use for its record_id a value that is numeric and incrementing. A suggestion to make that clearer:

The optional *page_size* parameter, when set, enables paged fetching of records by their primary record id (i.e. :attr:`.record_id_field`). This parameter can be used only with REDCap projects that use a record_id that is an incrementing integer.

I dropped that auto-incremented part since we don't care how that record_id integer gets generated.

tsibley: Thanks. I've adapted your wording to clarify further and repushed.

lib/id3c/cli/redcap.py
```diff
@@ -225,7 +238,43 @@ def records(self, *,
...
+            return self._fetch_records_paged(parameters, page_size)
+        else:
+            return list(self._fetch_records(parameters))
+
+
+    def _fetch_records_paged(self, parameters: dict, page_size: int) -> Iterator['Record']:
+        assert bool(self._details["record_autonumbering_enabled"]), \
+            "Record auto-numbering must be enabled to use page_size parameter"
+
+        next_record_id = self._fetch("generateNextRecordName")
```
jccraft-uw, line 251: I didn't know where the "generateNextRecordName" came from, so I did some Googling. Ah, this is a string that the REDCap API understands. Could you put a comment above this line with something like the following

"Generate Next Record Name (content=generateNextRecordName) - To be used by projects with record auto-numbering enabled, this method exports the next potential record ID for a project. It generates the next record name by determining the current maximum numerical record ID and then incrementing it by one."

I see this on https://research.fredhutch.org/cds/en/redcap-version-history.html

kschwabefry (approved):
LGTM. We may get an empty last page of results if the next record number is at a page boundary, but this should be handled fine. Thanks for putting the note about page size vs expected incoming rate of records. Looks like we'd have a minimum buffer of one page_size.

jccraft-uw (approved):

lib/id3c/cli/redcap.py
kschwabefry: IIRC, the tough part about linking to our official redcap docs is that the link is version dependent and break once theres a new version released e.g. https://redcap.iths.org/redcap_v10.5.2/API/help.php?content=exp_next_id will soon break

jccraft-uw: Sure. I was suggesting putting in a comment to say what generateNextRecordName does.

tsibley: I will add a comment.

I didn't think one was necessary since `self._fetch(…)` is used elsewhere as well, without similar comments.

lib/id3c/cli/redcap.py
```diff
@@ -155,7 +155,7 @@ def record(self, record_id: str, *, raw: bool = False) -> List['Record']:
         fields ``redcap_event_name``, ``redcap_repeat_instrument``, and
         ``redcap_repeat_instance``.
         """
-        return self.records(ids = [record_id], raw = raw)
+        return self.records(ids = [record_id], raw = raw) # type: ignore
```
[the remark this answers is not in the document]
tsibley: Ah, I'll add a reason for the comment.

tsibley:
Repushed to address review comments. Thanks for the :eyes:!

While I was thinking about it, I realized we can fetch the last page without an upper bound to catch everything created since the start of pagination. This avoids the "new records created too fast with too small a page size" problem (albeit one that's not very likely).

lib/id3c/cli/redcap.py
jccraft-uw: Thanks.

lib/id3c/cli/redcap.py
jccraft-uw: Thanks

lib/id3c/cli/redcap.py
```diff
@@ -225,7 +239,47 @@ def records(self, *,
...
+        assert bool(self._details["record_autonumbering_enabled"]), \
+            "Record auto-numbering must be enabled to use page_size parameter"
+
+        # Query the current maximum record id + 1 for the project.  Due to the
+        # absence of transactions, this is a hideously unsafe API design choice
+        # on REDCap's part, but it's fine for our purposes.
+        next_record_id = self._fetch("generateNextRecordName")
+
+        pages = [
+            (lower, lower + page_size if lower + page_size < next_record_id else None)
```
jccraft-uw, line 258: Yes, indeed. Not setting an upper range for that last batch makes total sense.

lib/id3c/cli/redcap.py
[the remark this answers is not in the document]
kschwabefry: Aha, its a mypy thing. Thanks!

kschwabefry (approved):

PR #200, opened 2021-01-12
Remove retry limit for barcode minting

jccraft-uw:
This commit removes the logic in the `mint_identifiers` function that throws
an error if a maximum number of minting retries are hit. When we hit that
currently, we simply re-run the job. With this maximum retry limit removed,
the job will just keep retrying until it gets a valid barcode.

### DIFF
--- a/lib/id3c/db/__init__.py
+++ b/lib/id3c/db/__init__.py
@@ -32,35 +32,16 @@ class IdentifierSetNotFoundError(IdentifierMintingError):
         return f"No identifier set named «{self.name}»"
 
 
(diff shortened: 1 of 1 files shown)


jccraft-uw:
[Barcode minting: remove the retry condition that stops the job](https://trello.com/c/A1Jw7qe6/951-barcode-minting-remove-the-retry-condition-that-stops-the-job)

kschwabefry (approved):

tsibley (approved):

jccraft-uw:
Deployed

PR #201, opened 2021-01-14
redcap: Support different date formats when updating records

tsibley:
Allows the import into the UW EH&S reporting project to use this method
instead of rolling its own API call.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -289,7 +289,7 @@ class Project:
         return (Record(self, r) for r in self._fetch("record", parameters))
 
 
-    def update_records(self, records: List[Dict[str, str]]) -> int:
+    def update_records(self, records: List[Dict[str, str]], date_format: str = "YMD") -> int:
         """
         Update existing *records* in this REDCap project.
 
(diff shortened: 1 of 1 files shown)


jccraft-uw (approved):

lib/id3c/cli/redcap.py
```diff
@@ -289,7 +289,7 @@ def _fetch_records(self, parameters: dict) -> Iterator['Record']:
         return (Record(self, r) for r in self._fetch("record", parameters))
 
 
-    def update_records(self, records: List[Dict[str, str]]) -> int:
+    def update_records(self, records: List[Dict[str, str]], date_format: str = "YMD") -> int:
```
jccraft-uw, line 292: This is where an enum could be handy, to know what the possible values are.

tsibley: Nod. I know there's a range of preferences on when to use enums, varying between "always" and "never". I'm somewhere in between; where exactly varies between languages too. An enum seemed like overkill to me here (as opposed to say, `InstrumentStatus`, which felt appropriate when I added it), but I haven't tried to articulate why. We can discuss more in Slack if you want.

As far as knowing what the possible values are: they're documented in the method's docstring and also validated at runtime with an assert. An enum _would_ add a small amount of compile-time safety.

kschwabefry, line 292: The EHS pipeline does create new records in a redcap project. Will update_records() work for that use case? The comments call out that this is for updating existing records and "not suitable for creating new records in projects that use auto-numbered record ids.", but that project uses barcodes as record ids so that exclusion may not apply.

tsibley: I believe it is suitable because auto-numbered ids aren't in use, but have not tested it directly yet because I'm not sure how to do that without doing it against the production project.

I documented the auto-numbered ids caveat because at the time I added `update_records()` I didn't want to think about the edge cases there (with regard to `forceAutoNumbering: false` and other things). I started too, but then decided that was scope creep I didn't want to deal with at the time.

kschwabefry (approved):

kairstenfay (approved):

PR #202, opened 2021-01-20
Redcap dal

kairstenfay:
Add additional methods to our redcap `Project` class to support different functionality we use in the [backoffice](https://github.com/seattleflu/backoffice) repo.

Methods include
* report
* metadata
* update_metadata
* users
* update_users

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -420,6 +420,52 @@ class Project:
         return self._fetch('user', {})
 
 
+    def update_users(self, users: List[Dict[str, Any]]) -> int:
+        """
+        Update existing *users* in this REDCap project.
+
+        *users* must be an iterable of :py:class:``dict``s mapping REDCap
+        user metadata to values.  The username, at a minimum, must be included.
+        All keys must be strings. From the REDCap API documentation:
+
+        > All values should be numerical with the exception of username,
+        > expiration, data_access_group, and forms.
+
+        Any value provided for a field, including the empty string, will
+        overwrite any existing value. Missing attributes, according to the
+        REDCap API docs, are handled by provisioning a user with:
+
+        > the minimum privileges (typically 0=No Access) for the
+        > attribute/privilege. However, if an existing user's privileges are
+        > being modified in the API request, then any attributes not provided
+        > will not be modified from their current value but only the attributes
+        > provided in the request will be modified.
+
+        Returns a count of the number of users updated, as reported by REDCap.
+        """
+        expected_count = len(users)
+
+        parameters = {
+            'data': json.dumps(users)
+        }
+
+        if not self.dry_run:
+            LOG.debug(f"Updating {expected_count:,} REDCap users for {self}")
+            result = self._fetch("user", parameters)
+
+            updated_count = result
+        else:
+            LOG.debug(f"Pretending to update {expected_count:,} REDCap users for {self} (dry run)")
+            updated_count = expected_count
+
+        assert expected_count == updated_count, \
+            "Expected vs. actual users updated do not match: {expected_count:,} != {updated_count:,}"
+
+        LOG.debug(f"Updated {updated_count:,} REDCap users for {self}")
+
+        return updated_count
+
+
     def _fetch(self, content: str, parameters: Dict[str, str] = {}, *, format: str = "json") -> Any:
         """
         Fetch REDCap *content* with a POST request to the REDCap API.
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -413,6 +413,13 @@ class Project:
         return updated_count
 
 
+    def users(self) -> List[Dict[str, Any]]:
+        """
+        Fetch the REDCap project's users.
+        """
+        return self._fetch('user', {})
+
+
     def _fetch(self, content: str, parameters: Dict[str, str] = {}, *, format: str = "json") -> Any:
         """
         Fetch REDCap *content* with a POST request to the REDCap API.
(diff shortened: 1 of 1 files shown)


kairstenfay:
[Move custom POSTs to REDCap API in random backoffice Python scripts to the id3c data access layer](https://trello.com/c/G8HPR9mZ/961-move-custom-posts-to-redcap-api-in-random-backoffice-python-scripts-to-the-id3c-data-access-layer)

lib/id3c/cli/redcap.py
```diff
@@ -361,6 +361,9 @@ def _fetch(self, content: str, parameters: Dict[str, str] = {}, *, format: str =
         }
 
         response = requests.post(self.api_url, data=data, headers=headers)
+        if response.status_code != 200:
+            LOG.debug(response.json())
+
```
tsibley, line 488: FWIW, last night I wrote something for the same reasons, but with a different approach, in 86f060a.

kairstenfay: Awesome! I'll drop this commit, then.

jccraft-uw (approved):
It will be so great to have these functions (I'm excited about `report`) in our core library.

lib/id3c/cli/redcap.py
```diff
@@ -362,6 +362,13 @@ def report(self, report_id: str, raw: bool = False) -> List[Dict[str, Any]]:
         return self._fetch('report', parameters)
 
 
+    def metadata(self) -> List[Dict[str, str]]:
+        """
+        Fetch the REDCap project metadata.
+        """
+        return self._fetch('metadata', {})
+
+
```
tsibley, line 371: There is an existing `fields` property which provides this, and I think "fields" is a better/more precise name than "metadata".

lib/id3c/cli/redcap.py
```diff
@@ -369,6 +369,50 @@ def metadata(self) -> List[Dict[str, str]]:
...
+        overwrite any existing value. From the REDCap API Documentation:
+
+        > Because of this method's destructive nature, it is only available for
+        > use for projects in Development status.
+
+        Returns a count of the number of records updated, as reported by
+        REDCap.
+        """
+        parameters = {
+            'data': json.dumps(metadata),
```
tsibley, line 391: This should use `as_json()` from our `id3c.json` module to avoid the same pitfalls that 8f3eaf2953c43c09c0043d96218234b6d177cb84 corrected.

lib/id3c/cli/redcap.py
```diff
@@ -369,6 +369,50 @@ def metadata(self) -> List[Dict[str, str]]:
         return self._fetch('metadata', {})
 
 
+    def update_metadata(self, metadata: Dict[str, str]) -> int:
```
tsibley, line 372: Given the existing `.fields` property, I think this should be renamed to `.update_fields()`.

lib/id3c/cli/redcap.py
```diff
@@ -420,6 +420,52 @@ def users(self) -> List[Dict[str, Any]]:
...
+        > being modified in the API request, then any attributes not provided
+        > will not be modified from their current value but only the attributes
+        > provided in the request will be modified.
+
+        Returns a count of the number of users updated, as reported by REDCap.
+        """
+        expected_count = len(users)
+
+        parameters = {
+            'data': json.dumps(users)
```
tsibley, line 449: Ditto re: `as_json()`.

lib/id3c/cli/redcap.py
```diff
@@ -369,6 +369,50 @@ def metadata(self) -> List[Dict[str, str]]:
...
+            'returnContent': 'count',
+        }
+
+        expected_count = len(metadata)
+
+        if not self.dry_run:
+            LOG.debug(f"Updating {expected_count:,} REDCap metadata for {self}")
+            result = self._fetch("metadata", parameters)
+
+            updated_count = result
```
tsibley, line 403: Can you confirm that you tested that the API response is a bare number and not `{"count":123}` like the response for record updating?

lib/id3c/cli/redcap.py
```diff
@@ -420,6 +420,52 @@ def users(self) -> List[Dict[str, Any]]:
...
+
+        parameters = {
+            'data': json.dumps(users)
+        }
+
+        if not self.dry_run:
+            LOG.debug(f"Updating {expected_count:,} REDCap users for {self}")
+            result = self._fetch("user", parameters)
+
+            updated_count = result
```
tsibley, line 456: Ditto re: confirming the API response.

tsibley (commented):
Thanks for this work! :+1: for not delaying merge after the first review. I just did a post-merge review and noted some changes I think should be made.

lib/id3c/cli/redcap.py
kairstenfay: Oops! I overlooked that entirely. Agree on our name being more precise. I'll drop this method.

lib/id3c/cli/redcap.py
kairstenfay: Ah, thank you!

lib/id3c/cli/redcap.py
kairstenfay: Confirmed. Apparently the REDCap API is inconsistent regarding what it means when we specify `{return: "count"}` in the request.

lib/id3c/cli/redcap.py
kairstenfay: Also confirmed.

kairstenfay:
Thanks @tsibley ! I opened up a new PR, #209, to address this feedback.

PR #205, opened 2021-01-26
Update labelmaker with school testing labels

jccraft-uw:
We now have two collection sets for the schools testing study. Use the same naming convention used for the UW Reopening study--"home" and "observed".

I minted and made labels locally. "SCHOOL TESTING OBSERVED <barcode>" does just fit onto the current label. Nothing gets truncated for home or observed.


kschwabefry (approved):

jccraft-uw:
Deployed.

PR #211, opened 2021-02-27
Rework barcode slices to resolve db restore issue

jccraft-uw:
Since the introduction of the barcode slices functionality, restoring
the production database has been failing with errors that look like:
`pg_restore: [archiver (db)] Error from TOC entry 6314; 1259 4123860
INDEX identifier_barcode_slices_idx postgres
pg_restore: [archiver (db)] could not execute query: ERROR:
type "citext[]" does not exist
LINE 5:         slices citext[];'

The following link showed a similar issue, and the author figured out that
when pg_restore creates functional indexes on tables in a schema different
from the schema in which the function is defined, that it fails
to find the function. His solution, which works for our situation, is
to explictly prefix the function's schema when it is used in the indexes.
https://www.postgresonline.com/journal/archives/121-Restore-of-functional-indexes-gotcha.html

In the funtction script itself I also prefixed the schema (public)
where citext is defined, just for safe measure.

In the sqitch plan I added a tag that I forgot to add as part of the
barcodes slices deployment a couple of months ago.


jccraft-uw:
[Fix citext-dependent indices on the testing db instnace](https://trello.com/c/q72HNmBK/958-fix-citext-dependent-indices-on-the-testing-db-instnace)

tsibley:
I'm a little confused by the resolution here.

To restate things as I understand them to make sure we're on the same page: the problem is finding the `citext[]` type during function execution during creation of the functional index, not finding the function itself.

Given that, does using the schema-qualified name of the function in the index resolve the type resolution on its own, without the additional schema-qualification of the types inside the function?

Or is the real fix schema-qualifying the types inside the function?

jccraft-uw:
> I'm a little confused by the resolution here.
> 
> To restate things as I understand them to make sure we're on the same page: the problem is finding the `citext[]` type during function execution during creation of the functional index, not finding the function itself.
> 
> Given that, does using the schema-qualified name of the function in the index resolve the type resolution on its own, without the additional schema-qualification of the types inside the function?
> 
> Or is the real fix schema-qualifying the types inside the function?

I didn't try the fixes separately because the round trip to test (dumping the database and then restoring it) takes a long time.

jccraft-uw:
> > > I'm a little confused by the resolution here.
> > > To restate things as I understand them to make sure we're on the same page: the problem is finding the `citext[]` type during function execution during creation of the functional index, not finding the function itself.
> > > Given that, does using the schema-qualified name of the function in the index resolve the type resolution on its own, without the additional schema-qualification of the types inside the function?
> > > Or is the real fix schema-qualifying the types inside the function?
> > 
> > 
> > I didn't try the fixes separately because the round trip to test (dumping the database and then restoring it) takes a long time.
> 
> But it WOULD be nice to know! I'm going to try without the change to the function itself. I suspect the real fix was in the index creation.

@tsibley , alas it is explicitly specifying the public schema in the function that solves this problem. Changing the indexes did not have an impact. I've removed the index changes from this PR.

tsibley:
Thanks for digging!

jccraft-uw:
Deployed.

PR #212, opened 2021-03-02
Add collection sets for new studies to Manifest ETL

jccraft-uw:
We forgot to add the collection sets for the new studies to the
manifest ETL. As a result, samples with them are being skipped.


jccraft-uw:
Deployed.

PR #213, opened 2021-03-10
redcap: Include the project in two log lines which were missing it

tsibley:
Most lines already included the project info.

This is helpful when reading logs for a process that accesses multiple
REDCap projects.


jccraft-uw (approved):
It's super helpful to log the project ID!

PR #214, opened 2021-03-23
Use supportsAllDrives when getting Google metadata

jccraft-uw:
In the `files().get` call to get the metadata for a Google Drive document
specify supportsAllDrives=True. SFS has moved content into a shared
Google Drive, and without this parameter set to True we get a 404
error.

From https://developers.google.com/drive/api/v2/reference/files/get:
"Whether the requesting application supports both My Drives and shared
drives. (Default: false)"

### DIFF
--- a/lib/id3c/cli/io/google.py
+++ b/lib/id3c/cli/io/google.py
@@ -87,6 +87,6 @@ def get_document_etag(document_id: str) -> str:
     Returns the etag of a Google Drive document.
     """
     drive_service = discovery.build("drive", "v2")
-    metadata = drive_service.files().get(fileId=document_id).execute()
+    metadata = drive_service.files().get(fileId=document_id, supportsAllDrives=True).execute()
 
     return metadata["etag"]


kschwabefry (approved):

tsibley:
This looks fine to me! Thanks for debugging it. What an option. I was curious to read more about this option/the API and [ran across this](https://developers.google.com/resources/api-libraries/documentation/drive/v3/python/latest/drive_v3.files.html):

> supportsTeamDrives: boolean, Deprecated use supportsAllDrives instead.
>
> supportsAllDrives: boolean, Deprecated - Whether the requesting application supports both My Drives and shared drives. This parameter will only be effective until June 1, 2020. Afterwards all applications are assumed to support shared drives.

:thinking:

tsibley (approved):

jccraft-uw:
> This looks fine to me! Thanks for debugging it. What an option. I was curious to read more about this option/the API and [ran across this](https://developers.google.com/resources/api-libraries/documentation/drive/v3/python/latest/drive_v3.files.html):
> 
> > supportsTeamDrives: boolean, Deprecated use supportsAllDrives instead.
> > supportsAllDrives: boolean, Deprecated - Whether the requesting application supports both My Drives and shared drives. This parameter will only be effective until June 1, 2020. Afterwards all applications are assumed to support shared drives.

I came across that June 1, 2020 date too. Since our code doesn't work without the parameter and adding it in causes it to work, it looks like that change hasn't automatically come to us. We are not pinning a specific version of `google-api-python-client` in ID3C.

tsibley:
Nod, yeah, I would suspect that doc is just wrong or out of date. It looks auto-generated in the javadoc style, and the other docs I can find don't mention anything about deprecation.

jccraft-uw:
Deployed.

PR #217, opened 2021-03-29
Add APIJsonDecodeError to log the response text

jccraft-uw:
When fetching REDCap records, raise the custom `APIJsonDecodeError`
error so that we log the text that is failing JSON decoding. This will
help us continue the investigation with ITHS.

With the response context set to b'abc' to simulate the error:
Currently we see logged:
Exception has occurred: JSONDecodeError
Expecting value: line 1 column 1 (char 0)

With this commit we will see:
Exception has occurred: APIJsonDecodeError
Error: 'Expecting value' Text to decode: 'abc'

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -493,7 +493,14 @@ class Project:
         except requests.HTTPError as e:
             raise APIError(response = response) from e
 
-        return response.json() if format == "json" else response.text
+        if format == "json":
+            try:
+                return response.json()
+            except json.decoder.JSONDecodeError as e:
+                raise APIJsonDecodeError(msg = e.msg, doc = e.doc, pos = e.pos)
+
+        else:
+            return response.text
 
 
     def __repr__(self) -> str:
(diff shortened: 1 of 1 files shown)


jccraft-uw:
[Investigate intermittent redcap api response issues](https://trello.com/c/hH2Htc4b/1034-investigate-intermittent-redcap-api-response-issues)

lib/id3c/cli/redcap.py
```diff
@@ -795,3 +802,13 @@ def __str__(self):
...
+
+
+class APIJsonDecodeError(json.decoder.JSONDecodeError):
+    """
+    Error class for bad JSON responses from the REDCap API.
+
+    Includes useful details of the error in the stringification.
+    """
+    def __str__(self):
+        return f"Error: {self.msg!r} Text to decode: {self.doc!r}"
```
tsibley, line 814: I think it'd be nicer/more informative to keep the existing detailed stringification behaviour (e.g. call `super().__str__()`) but extend it with the original doc (or better yet, a length-limited substring of the original doc centered around the error position).

lib/id3c/cli/redcap.py
```diff
@@ -795,3 +802,13 @@ def __str__(self):
         # to see things exactly as they were, with minimal post-processing,
         # when troubleshooting.
         return f"{self.response.status_code} {self.response.reason}: {self.response.text}"
+
+
+class APIJsonDecodeError(json.decoder.JSONDecodeError):
```
tsibley, line 807: I'd suggest matching our name (and casing) of the exception with the stdlib, using just `class JSONDecodeError` here. Any future catching of it can be scoped by module (e.g. `id3c.cli.redcap.JSONDecodeError`). This also suggests to me that maybe we should add this "enhanced" `JSONDecodeError` to the `id3c.json` library instead…

tsibley:
I implemented my suggestions, with a bit of generalization and caution around PII leakage, as #218.

PR #218, opened 2021-04-02
Better JSON decoding errors

tsibley:
I added context to JSON decoding errors for better troubleshooting of intermittent REDCap issues. Inspired by #217 but generalized a bit.

### DIFF
--- a/lib/id3c/utils.py
+++ b/lib/id3c/utils.py
@@ -94,3 +94,64 @@ def shorten(text, length, placeholder):
         return text[0:length - len(placeholder)] + placeholder
     else:
         return text
+
+
+def contextualize_char(text, idx, context = 10):
+    """
+    Marks the *idx* char in *text* and snips out a surrounding amount of
+    *context*.
+
+    Avoids making a copy of *text* before snipping, in case *text* is very
+    large.
+
+    >>> contextualize_char('hello world', 0, context = 4)
+    '▸▸▸h◂◂◂ello…'
+    >>> contextualize_char('hello world', 5, context = 3)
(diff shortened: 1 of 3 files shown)


lib/id3c/utils.py
```diff
@@ -94,3 +94,64 @@ def shorten(text, length, placeholder):
         return text[0:length - len(placeholder)] + placeholder
     else:
         return text
+
+
+def contextualize_char(text, idx, context = 10):
```
kschwabefry, line 99: Just commenting, I'll be curious to see if using the default here (up to) 10 characters on either side ends up being enough to debug these. I think trying to limit leakage of potential phi makes sense though, so this seems like a good starting point.

kschwabefry (approved):
Looks great! This extra logging info will hopefully be really interesting debugging the JSONDecodeErrors with redcap api responses.

kschwabefry:
[Investigate intermittent redcap api response issues](https://trello.com/c/hH2Htc4b/1034-investigate-intermittent-redcap-api-response-issues)

lib/id3c/utils.py
tsibley: Yeah, we'll see! I think 10 chars of surrounding context will be sufficient for many actual JSON encoding errors, but I don't really expect actual encoding errors that often.

More often I expect that the response body isn't JSON at all (even if it is supposed to be) and that this will result in an error at position 0, which triggers the first branch of this conditional:

https://github.com/seattleflu/id3c/blob/74611ac5c11fd1be93ab60ce276b4cf4395f23d1/lib/id3c/json.py#L96-L100

tsibley:
Deployed.

PR #222, opened 2021-04-16
identifier/labelmaker: add --copies-per-barcode option

joverlee521:
When minting barcodes or printing labels, use the `--copies-per-barcode`
option to specify how many copies per barcode should be included in the
printable barcode labels.

NOTE: This is a naive overwrite of the default number. It does not take
into account the specific layout for better UX that may require blanks.
For example, changing the copies per barcode for the "samples" barcodes
can result in a confusing layout in the PDF.


kschwabefry (approved):
Thanks for adding the choice restriction too, that looks great.

PR #223, opened 2021-04-29
Migrate tests from Travis CI to GitHub Actions

tsibley:
We ran out of free trial credits on Travis CI back in January.  Rather
than apply for a one-time grant of open-source credits (and have to
re-apply in the future when they run out again, whenever that might be),
move to GitHub Actions where this isn't an issue.  We're already using
GitHub for repository hosting and PRs, so using it for CI too makes
sense and reduces our service sprawl.

I based this CI workflow on the one I setup and honed for the Nextstrain
CLI repo a month or two ago.


tsibley:
The Travis config ran on two major versions of Python, and so when converting to GitHub Actions, I expanded that to include two more recent major versions as well. However, since we run the tests within Pipenv, they always run under 3.6 because that's what our Pipfile declares. This means we were running twice as many Travis jobs as necessary, since both ended up using 3.6. I realized this because on GitHub, the environment doesn't contain 3.6 unless we specifically install it during setup, so all the non-3.6 jobs failed immediately when Pipenv couldn't find a 3.6 install.

I think we want the tests to run under multiple Python versions, while still keeping the Pipenv locked to a single version for development/deploy, so I've marked this draft and will come back to it to address how that all works together. The tests will run outside of Pipenv once I move the dev deps out of Pipfile and into setup.py.

tsibley:
Ok, it's working now, with the addition of a few more commits. I'll address id3c-customizations next, which should be easier now that I've worked out the kinks here.

It'll be nice to have CI working again.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - pull_request
+
+jobs:
+  test:
+    name: python=${{ matrix.python }} os=${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
```
kschwabefry, line 14: It looks like backoffice is running ubuntu 18.04. What would you think about adding that into the matrix here?

tsibley: Sure, that's reasonable. I don't expect this OS version difference to impact our test results very much, but it's definitely possible. I've pushed an additional commit here and on https://github.com/seattleflu/id3c-customizations/pull/211.

kschwabefry: Right on. Thanks Thomas!

kschwabefry (approved):

PR #226, opened 2021-06-08
Improve JSON decoding errors

tsibley:
Small improvements after noticing an oddity in our alerts.

Full details of changes in commit messages.

### DIFF
--- a/lib/id3c/utils.py
+++ b/lib/id3c/utils.py
@@ -68,15 +68,11 @@ def shorten(text, length, placeholder):
 
     The maximum *length* must be longer than the length of the *placeholder*.
 
-    The right-hand end of *text* is stripped of whitespace.
-
     Behaviour is slightly different than :py:func:`textwrap.shorten` which is
     intended for shortening sentences and works at the word, not character,
     level.
 
     >>> shorten("foobar", 6, "...")
-    'foobar'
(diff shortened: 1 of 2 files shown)


tsibley:
Rebased onto latest `master` to pickup unrelated fixes for CI.

kschwabefry (approved):
Looks great! 👍 for breaking up commits and adding lots of doc tests along the way!

lib/id3c/utils.py
```diff
@@ -90,6 +90,29 @@ def shorten(text, length, placeholder):
         return text
 
 
+def shorten_left(text, length, placeholder):
```
joverlee521, line 93: Out of curiosity, what's the thought process behind making this a new function rather than adding a parameter to `shorten()`?

tsibley: Good question. I don't recall considering that when writing this (although the commit's author date was a month ago, so memory might be faulty). I think my default was to have a separate parallel function because the core of the function is substantially different, but I do see now that the supporting code is identical. We could adapt it if you prefer!

joverlee521: Definitely fine to keep as is, I have no preference 🤷

tsibley:
Deployed.

PR #228, opened 2021-06-14
geocode: Avoid extracting addresses from empty or whitespace-only strings

tsibley:
SmartyStreets Python API throws an error when given an empty or
whitespace-only string.  Even if it didn't, there's no point in making a
request for an empty address.

Fixes an error we observed in production in the ETL for metadata about
UW residual samples.

### DIFF
--- a/lib/id3c/cli/command/geocode.py
+++ b/lib/id3c/cli/command/geocode.py
@@ -281,7 +281,7 @@ def geocode_address(address: dict) -> dict:
         STREET_CLIENT = smartystreets_client_builder().build_us_street_api_client()
 
     lookup = us_street_lookup(address)
-    if not lookup.street:
+    if lookup.street is None or not lookup.street.strip():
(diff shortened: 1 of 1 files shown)


joverlee521 (changes requested):
Thanks for looking into this @tsibley! 

It's definitely good to have this check in place within `extract_address()` but I think we can go even further upstream to prevent extra SmartyStreet lookups. Within `geocode_address()` we already check if the address contains a `street`, so I think we just need to add `strip()` to catch the whitespace-only strings:

```diff
 
     lookup = us_street_lookup(address)
-    if not lookup.street:
+    if not lookup.street.strip():
         LOG.warning(f"Missing street address; can't geocode")
         return None
 
```

tsibley:
That's fair. I've pushed another commit adjusting that check in `geocode_address()`.

joverlee521 (approved):

PR #230, opened 2021-06-21
Web API for identifiers and identifier sets

tsibley:
Refresh of [a long-standing WIP](https://github.com/seattleflu/id3c/compare/wip/identifier-api) from a couple years ago to add an identifier API. It had slightly different but overlapping motivations at the time. I refreshed it on the latest code and added an identifier lookup endpoint which could be used for validation.

Marked as draft because I expect we'll want some further consideration first. We don't have to use this as-is, but I figured it could be a starting point for discussion about what we do want there with much of the plumbing done.

Commit messages contain details.

### DIFF
--- a/lib/id3c/api/routes.py
+++ b/lib/id3c/api/routes.py
@@ -194,18 +194,30 @@ def get_identifier_set(name, *, session):
 
 
 @api_v1.route("/warehouse/identifier-sets/<name>", methods = ['PUT'])
+@content_types_accepted(["application/x-www-form-urlencoded", "multipart/form-data", None])
+@check_content_length
 @authenticated_datastore_session_required
 def put_identifier_set(name, *, session):
     """
     Make a new identifier set.
 
     PUT /warehouse/identifier-sets/*name* to create the set if it doesn't yet
-    exist.  201 Created is returned when the set is created, 204 No Content if
-    the set already existed.
+    exist.
+
+    If a *description* form parameter is provided, its value is set/updated in
+    the database.
+
+    201 Created is returned when the set is created or updated, 204 No
+    Content if the set already existed.
     """
     LOG.debug(f"Making identifier set «{name}»")
 
-    new_set = datastore.make_identifier_set(session, name)
+    fields = {}
+
+    if "description" in request.form:
+        fields["description"] = request.form["description"]
+
+    new_set = datastore.make_identifier_set(session, name, **fields)
 
     return "", 201 if new_set else 204
 
--- a/lib/id3c/api/routes.py
+++ b/lib/id3c/api/routes.py
@@ -4,7 +4,7 @@ API route definitions.
 import json
 import logging
 import pkg_resources
-from flask import Blueprint, request, send_file
+from flask import Blueprint, jsonify, request, send_file
 from . import datastore
 from .utils.routes import authenticated_datastore_session_required, content_types_accepted, check_content_length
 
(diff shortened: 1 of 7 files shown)


lib/id3c/api/routes.py
```diff
@@ -145,6 +146,99 @@ def receive_fhir(*, session):
...
+    metadata fields.
+    """
+    LOG.debug(f"Fetching identifier set «{name}»")
+
+    set = datastore.fetch_identifier_set(session, name)
+
+    return jsonify(set._asdict())
+
+
+@api_v1.route("/warehouse/identifier-sets/<name>", methods = ['PUT'])
```
geomapdev, line 197: Is there a reason you chose PUT instead of POST method here? Also, would it be beneficial to also be able to set the description value for the new record?

lib/id3c/api/routes.py
```diff
@@ -145,6 +146,99 @@ def receive_fhir(*, session):
...
+    the set already existed.
+    """
+    LOG.debug(f"Making identifier set «{name}»")
+
+    new_set = datastore.make_identifier_set(session, name)
+
+    return "", 201 if new_set else 204
+
+
+@api_v1.route("/warehouse/identifier-sets/<name>/identifiers", methods = ['POST'])
```
geomapdev, line 214: Is it worth considering setting an upper limit to n value here?

geomapdev (commented):
I tested each of these endpoint successfully. Looks like a useful set! Just a few questions.

lib/id3c/api/routes.py
tsibley: Good question. In my experience:

PUT is usually used for idempotent operations acting directly on a specific resource (e.g. one with a unique id, in this case `<name>`).

POST is usually used for non-idempotent operations acting on a collection resource, to take arbitrary action (e.g. launch a job) or create new records where you don't know the unique id in advance (e.g. auto-numbered records). Responses to POSTs are often 302 Redirects to the newly-created specific resource.

There are lots of discussions about this, of course. One reasonable example is https://restfulapi.net/rest-put-vs-post/.

It would indeed probably be beneficial to allow a description to be set. I didn't think it necessary as a first pass (and it wasn't in my old wip), but it wouldn't be hard to add. I'll add it.

lib/id3c/api/routes.py
tsibley: Ah, good call. It's always good to consider what limits are appropriate. I'm not sure what reasonable bound would be though… Here's the current summary stats of batch size for all we've minted so far:

```
min	     1
mode	 1,040
max	11,900
```

Stepping back from `n`, I think what a limit would really seek to prevent is minting reqs that run "forever". We can't guarantee this just through `n` alone (although very small `n` might approximate it well enough), so maybe the more appropriate thing would be an explicit time limit. There is some existing limit on request/response duration already at the web server level (both at uWSGI and Apache, I believe), so that may provide enough of a time limit if the corresponding database session is terminated when the request is. But I'm not sure that's the case.

This also means that we may not be able to use this API as-is to mint large batches of identifiers without increasing existing request/response duration limits, possibly to unreasonable levels. An async API would obviate that issue, but then require a minting job queue and polling of job status by the client. (We've talked about this in the past in related convos.)

@seattleflu/core-devs thoughts?

lib/id3c/api/routes.py
geomapdev: Ok, that explanation makes sense. I came across that same page, and the last entry in the comparison table was throwing me off:
Generally, in practice, always use PUT for UPDATE operations. | Always use POST for CREATE operations.
-- | --

But your explanation matches up well with [this answer](https://stackoverflow.com/questions/630453/what-is-the-difference-between-post-and-put-in-http) on SO, so thanks for clearing that up for me!

tsibley: Added support for "description" with additional commits.

lib/id3c/api/routes.py
geomapdev: I think it makes sense to drop it or if possible just disable that route until there's a use case for it. I see that the `id3c identifier mint` command is used when generating barcode labels, but I'm not familiar enough with all of the workflows to know which ones might need this functionality.

tsibley: Yeah, I agree, reasonable to drop it from this PR. I'd move it to another branch and push that up as a wip, to potentially be picked up by someone else in the future. Would love input from others here too before I do.

tsibley: Repushed dropping the minting endpoint.

davereinhart (approved):
Looks good. After this is merged, I'll make some minor changes to this code as part of PR #234 to include the new required `use` column in the `identifier_set` table.

tsibley:
Sounds like a plan, thanks!

tsibley:
Deployed.

PR #232, opened 2021-06-23
Pinning click to version 7.x

geomapdev:
After prior deployment that updated click to version 8.0.1, ETLs started
encountering issues. They continued to run without error, but changes were
rolled back inexplicably. This is possibly related to change in click
flag_value typing [#1886](https://github.com/pallets/click/issues/1886).

Confirmed with local testing that rolling back to version 7.1.2 resolved
the problem. Pinning click to 7.x while investigating further and making
required code changes to upgrade to 8.x.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,10 +18,10 @@
     "default": {
         "aiobotocore": {
             "hashes": [
-                "sha256:81890d270b1f948ffd218e8bab11e235bea272840ea8b1b9e0aef1954c6cec9e"
+                "sha256:8ecee55346651e0f4cbda883e3e16cfe11460b8d7adcc08d0017cbb867636ae1"
             ],
             "markers": "python_version >= '3.6'",
-            "version": "==1.3.0"
+            "version": "==1.3.1"
         },
         "aiohttp": {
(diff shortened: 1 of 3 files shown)


joverlee521:
I believe the failed mypy tests can be ignored by adding click to the `mypy.ini` file:
```
[mypy-click]
ignore_missing_imports = True
```
We've done this in the past for other packages as well (see https://github.com/seattleflu/id3c/commit/b92279722f4a2252ad42baf3c648a3b9c1107121)

kschwabefry (approved):
Thanks Dave! Great to have this fix.
I might suggest rebasing to combine the commits https://github.com/seattleflu/id3c/pull/232/commits/4f01d627bc8f14c0b56c9e066260a11a1e7d08c4 and https://github.com/seattleflu/id3c/pull/232/commits/01f0ecb0c6a48ef4d11b4bfbdfc82e0df2cef232 (if not all 3 commits in this PR) because pinning the other two packages was just an intermediate state of the code as part of the development process.

joverlee521 (approved):
Echoing @kschwabefry's suggestion to rebase for a cleaner git history.

PR #233, opened 2021-06-30
Add pyproject.toml to .gitignore

geomapdev:
A pyproject.toml file is auto-generated when using pipenv>=2020.11.15. Adding
it to .gitignore out of an abundance of caution, to avoid having it deployed until
production is upgraded to pipenv>=2020.11.15 but allowing us to use those versions
of pipenv locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ __pycache__/
 /.mypy_cache/
 /lib/id3c.egg-info/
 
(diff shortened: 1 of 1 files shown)


geomapdev:
This may not need much in the way of testing, other than updating to pipenv>=2020.11.15 locally then run `pipenv sync` and look for the pyproject.toml file to be generated.

kschwabefry (approved):

PR #234, opened 2021-07-02
Add use column to identifier_set

davereinhart:
Identifier type (sample, collection, clia, etc.) is missing from the data
model, currenly only  encoded through identifier_set naming convention. Adding
it to identifier_set makes it more explicitly defined which will simplify
implementations of API and ETL functions and allow for future migration away
from using multiple identifier columns in sample table.

### DIFF
--- a/lib/id3c/api/datastore.py
+++ b/lib/id3c/api/datastore.py
@@ -187,8 +187,8 @@ def fetch_identifier(session: DatabaseSession, id: str) -> Any:
 
     *id* may be a full UUID or shortened barcode.
 
-    Returns a named tuple with ``uuid``, ``barcode``, ``generated``, and
-    ``set`` attributes.  If the identifier doesn't exist, raises a
+    Returns a named tuple with ``uuid``, ``barcode``, ``generated``, ``set``,
+    and ``use`` attributes.  If the identifier doesn't exist, raises a
     :class:`~werkzeug.exceptions.NotFound` exception.
     """
     try:
@@ -199,7 +199,7 @@ def fetch_identifier(session: DatabaseSession, id: str) -> Any:
 
     with session:
         identifier = session.fetch_row(f"""
-            select uuid, barcode, generated, identifier_set.name as set
+            select uuid, barcode, generated, identifier_set.name as set, identifier_set.use
               from warehouse.identifier
               join warehouse.identifier_set using (identifier_set_id)
              where {id_field} = %s
@@ -218,12 +218,12 @@ def fetch_identifier_sets(session: DatabaseSession) -> Any:
     """
     Fetch all identifier sets from the backing database using *session*.
 
-    Returns a list of named tuples with ``name`` and ``description``
+    Returns a list of named tuples with ``name``, ``description``, and ``use``
     attributes.
     """
     with session, session.cursor() as cursor:
         cursor.execute("""
-            select name, description
+            select name, description, use
               from warehouse.identifier_set
             """)
 
@@ -236,13 +236,13 @@ def fetch_identifier_set(session: DatabaseSession, name: str) -> Any:
     """
     Fetch the identifier set *name* from the backing database using *session*.
 
-    Returns a named tuple with ``name`` and ``description`` attributes.  If the
-    set doesn't exist, raises a :class:`~werkzeug.exceptions.NotFound`
+    Returns a named tuple with ``name``, ``description``, and ``use`` attributes.
+    If the set doesn't exist, raises a :class:`~werkzeug.exceptions.NotFound`
     exception.
     """
     with session:
         set = session.fetch_row("""
-            select name, description
+            select name, description, use
               from warehouse.identifier_set
              where name = %s
             """, (name,))
@@ -259,35 +259,96 @@ def fetch_identifier_set(session: DatabaseSession, name: str) -> Any:
 def make_identifier_set(session: DatabaseSession, name: str, **fields) -> bool:
     """
     Create a new identifier set *name* in the backing database using *session*
-    if it doesn't already exist.
+    if it doesn't already exist, or update if it does exist.
 
-    If *description* is provided as a keyword argument, its value is
-    set/updated in the database.
+    If *use* and/or *description* are provided as keyword arguments, their values
+    are set in the database. Becuase *use* is a required field in the target table,
+    if it is not provided as a keyword argument the query will attempt to retrieve
+    its value from an existing record.
 
     Returns ``True`` if the set was created or updated and ``False`` if it
-    already existed.
+    already existed and was not updated.
+
+    Raises a :class:`BadRequestDatabaseError` exception if the database reports a
+    `constraint` error and a :class:`Forbidden` exception if the database reports a
+    `permission denied` error.
     """
     with session, session.cursor() as cursor:
-        # If I expected to have additional columns in the future, I'd build up
-        # the SQL dynamically, but I don't, so a simple conditional is enough.
-        #   -trs, 1 July 2021
-        if "description" in fields:
-            cursor.execute("""
-                insert into warehouse.identifier_set (name, description)
-                    values (%s, nullif(%s, ''))
-                    on conflict (name) do update
-                        set description = excluded.description
-                """, (name, fields["description"]))
-
+        if "use" in fields and "description" in fields:
+            try:
+                cursor.execute("""
+                    insert into warehouse.identifier_set (name, use, description)
+                        values (%s, %s, nullif(%s, ''))
+                        on conflict (name) do update
+                            set use = excluded.use,
+                                description = excluded.description
+                            where identifier_set.use <> excluded.use
+                                or coalesce(identifier_set.description,'') <> coalesce(excluded.description,'')
+                    """, (name, fields["use"], fields["description"]))
+            except (DataError, IntegrityError) as error:
+                raise BadRequestDatabaseError(error) from None
+        elif "use" in fields:
+            try:
+                cursor.execute("""
+                    insert into warehouse.identifier_set (name, use)
+                        values (%s, %s)
+                        on conflict (name) do update
+                            set use = excluded.use
+                            where identifier_set.use <> excluded.use
+                    """, (name, fields["use"]))
+            except (DataError, IntegrityError) as error:
+                raise BadRequestDatabaseError(error) from None
+        elif "description" in fields:
+            try:
+                cursor.execute("""
+                    insert into warehouse.identifier_set (name, use, description)
+                        select s.name, t.use, s.description 
+                        from (values(%s, nullif(%s,''))) s(name, description)
+                        left join (
+                            select name, use 
+                            FROM warehouse.identifier_set WHERE name = %s
+                        ) t using (name)
+                        on conflict (name) do update
+                            set use = excluded.use, description = excluded.description
+                            where identifier_set.use <> excluded.use
+                            or coalesce(identifier_set.description,'') <> coalesce(excluded.description,'')
+                    """, (name, fields["description"], name))
+            except (DataError, IntegrityError) as error:
+                raise BadRequestDatabaseError(error) from None
         else:
-            cursor.execute("""
-                insert into warehouse.identifier_set (name)
-                    values (%s)
-                    on conflict (name) do nothing
-                """, (name,))
+            try:
+                cursor.execute("""
+                    insert into warehouse.identifier_set (name, use)
+                        select s.name, t.use
+                        from (values(%s)) s(name)
+                        left join (
+                            select name, use 
+                            FROM warehouse.identifier_set WHERE name = %s
+                        ) t using (name)
+                        on conflict (name) do update
+                            set use = excluded.use
+                            where identifier_set.use <> excluded.use
+                    """, (name, name))
+            except (DataError, IntegrityError) as error:
+                raise BadRequestDatabaseError(error) from None
 
         return cursor.rowcount == 1
 
+@export
+@catch_permission_denied
+def fetch_identifier_set_uses(session: DatabaseSession) -> Any:
+    """
+    Fetch all identifier set uses from the backing database using *session*.
+
+    Returns a list of named tuples with ``use`` and ``description`` attributes.
+    """
+    with session, session.cursor() as cursor:
+        cursor.execute("""
+            select use, description
+              from warehouse.identifier_set_use
+            """)
+
+        return list(cursor)
 
 @export
 class BadRequestDatabaseError(BadRequest):
--- a/lib/id3c/api/routes.py
+++ b/lib/id3c/api/routes.py
@@ -201,26 +201,39 @@ def put_identifier_set(name, *, session):
     """
     Make a new identifier set.
 
-    PUT /warehouse/identifier-sets/*name* to create the set if it doesn't yet
-    exist.
+    PUT /warehouse/identifier-sets/*name* to create the set if it doesn't yet exist.
 
-    If a *description* form parameter is provided, its value is set/updated in
-    the database.
+    For new sets, *use* form parameter is required. For existing sets, if *use* form parameter
+    is provided, its value is updated in the database. Valid *use* values can be found via
+    GET /warehouse/identifier-set-uses. 
+    
+    If a *description* form parameter is provided, its value is set/updated in the database.
 
     201 Created is returned when the set is created or updated, 204 No
-    Content if the set already existed.
+    Content if the set with *name* already existed and was not updated.
     """
     LOG.debug(f"Making identifier set «{name}»")
 
-    fields = {}
-
-    if "description" in request.form:
-        fields["description"] = request.form["description"]
-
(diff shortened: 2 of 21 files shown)


schema/deploy/warehouse/identifier/identifier-type.sql
```diff
@@ -0,0 +1,12 @@
+-- Deploy seattleflu/schema:warehouse/identifier/identifier-type to pg
+-- requires: warehouse/identifier
+
+begin;
+
+alter table warehouse.identifier_set
+    add column identifier_type text;
```
tsibley, line 7: Do we want to make the type of this column more restricted than `text`? I think we might. Much of the ID3C core code, not just our customizations, makes a distinction between sample and collection, for instance, so it feels like that should be formalized conceptually by the schema.

If we do want to narrow down the type, Pg's domain types are a good option for this sort of value where a vocabulary table is not necessary (i.e. because it's not intended to be customized). There's an example of using domain types for `warehouse.individual.sex`.

(If it was intended and expected to be customized or extensible, then a new vocabulary table would be more appropriate. An example of this in ID3C is the `warehouse.encounter_location_relation` table.)

---

Similarly, should we make this column `not null`? Probably, IMO. I think we can exhaustively assign our existing identifier sets. This will require a bit of coordination by making it `not null` in a separate sqitch change, so we can interleave with a manual `update`.

---

The `id3c identifier set create` command should also be able to set this column so we can do so when making new sets later. This is particularly important if the column is not null, but also to avoid needing manual `insert` or `update` statements in addition to the CLI command.

---

Let's index this column, as it seems very likely we'll be doing filtering/searching based on it. (I was reminded to consider this by the [schema development guidelines](https://github.com/seattleflu/id3c/blob/master/README.md#guidelines) I wrote long ago. :-)

--- 

Separately, the repetition of the table name in the column names is a pattern avoided in ID3C's schema (with the sole exception of generated serial pks). Let's name the column just `type` or `kind` or `use`? (I like `use` because it's more specific and less overloaded than `type`, but this is mostly aesthetics.)

tsibley (changes requested):

schema/deploy/warehouse/identifier/identifier-type.sql
geomapdev: Good suggestions, I'll get started on these! And thanks for the link to the guidelines, those are helpful.

I thought a new 1:M table might be overkill, but if the list of identifier types is likely to grow then that seems like a good option. It would also allow us to use an integer value instead of text, which should perform better for search/filter and even more so with indexing.

I wasn't sure how to implement `NOT NULL` without the required values in place, but a separate sqitch change makes sense.

The changes needed on `id3c identifier set create` is what I was getting at on the call earlier... whether that should be included in this PR or a later one. But with `NOT NULL` in place it should definitely be included here to avoid breaking it. That will also be the case for the new api function you have in PR #230, [put_identifier_set](https://github.com/seattleflu/id3c/pull/230/files#diff-4423d40718dcdaa9d1a971ec4f113f118e302460a71bc7c3118086d770817d49R201) so merging these two branches into master is going to take some coordination.

I avoided using `type` for the reason you mentioned but I wanted to keep it as part of the column name since that's how it was being referred to in other conversations. If `use` seems intuitive enough I can go with that.

davereinhart: After some testing I realized an integer id was not the way to go here. There's too much overhead with the extra join needed to query identifiers when the `use` text value is in a separate table.  I ended up using `warehouse.encounter_location_relation` as a template as @tsibley suggested, using `citext` as the PK which eliminates the need for a join.

I also tested adding an index on the new `use` column of `warehouse.identifier_set` but saw no improvement when querying identifiers by `use` type (using `EXPLAIN ANALYZE` method). I was able to reduce cost of the same queries by approximately 50% by adding a b-tree index on `warehouse.identifier.identifier_set_id`.

davereinhart:
To test this branch now requires multiple steps. After `git pull`:

1. Run the first set of sqitch changes on your local database (adds the new column and vocab table):
`sqitch deploy dev @2021-07-03`
2. Run the following queries to assign `use` values to all records in `warehouse.identifier_set`:
  ```
  UPDATE warehouse.identifier_set SET use='kit' WHERE name LIKE 'kits-%';
  UPDATE warehouse.identifier_set SET use='test-strip' WHERE name LIKE 'test-strips-%';
  UPDATE warehouse.identifier_set SET use='clia' WHERE name LIKE 'collections-clia-%';
  UPDATE warehouse.identifier_set SET use='collection' WHERE name LIKE 'collections-%' AND name NOT LIKE 'collections-clia-%';
  UPDATE warehouse.identifier_set SET use='sample' WHERE name='samples' OR name LIKE 'samples-%';
  ```
3. Run the final sqitch changes to add `not null` to the `use` column and add the new `identifier_set_id` index:
`sqitch deploy dev`

schema/deploy/warehouse/identifier-set-use/data.sql
```diff
@@ -0,0 +1,17 @@
...
+
+begin;
+
+insert into warehouse.identifier_set_use (use, description)
+    values
+        ('sample', 'Sample ID'),
+        ('collection', 'Collection ID'),
+        ('clia', 'CLIA ID'),
+        ('kit','Test kit ID'),
+        ('test-strip','Test strip ID')
```
tsibley, line 12: I'd suggest updating these descriptions to be a bit more useful, a la the table/column comments in our schema or the existing identifier set descriptions. I think future folks will appreciate it.

tsibley (changes requested):
Overall this looks great, and the commit messages are nicely descriptive.

Aside from the comments below, I also think it might be helpful to include each identifier set's use in the `id3c identifier set ls` output as a third column between the existing two columns.

schema/deploy/warehouse/identifier-set-use.sql
```diff
@@ -0,0 +1,21 @@
...
+
+comment on table warehouse.identifier_set_use is
+    'Controlled vocabulary for warehouse.identifier_set.use';
+
+comment on column warehouse.identifier_set_use.use is
+    'A standard identifier use type, e.g. sample, collection, clia';
+
+comment on column warehouse.encounter_location_relation.priority is
+    'A plain text description of this identifier use type';
+
```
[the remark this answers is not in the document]
davereinhart: Yes they were! Fixed.

schema/deploy/warehouse/identifier-set-use/data.sql
davereinhart: Yes, I think you're right. Updated them to be more "descriptive".

davereinhart:
Waiting for PR #230 to be merged to make additional changes (adding `use` column) to the new API as part of this PR.

davereinhart:
Rebased and made changes to API endpoints that were added with PR #230. Ready for review, in particular looking closely at the changes to the `warehouse/identifier-sets` PUT endpoint. The inclusion of a new required `use` field raises some questions about how to handle existing updates to existing records. In the end, I decided to only allow `use` to be set on creation and to only allow updates to `description` value on `name` conflicts if the incoming `use` value matches the existing record.

schema/deploy/warehouse/identifier-set-use/data.sql
tsibley: Thanks!

lib/id3c/api/routes.py
```diff
@@ -193,34 +193,49 @@ def get_identifier_set(name, *, session):
     return jsonify(set._asdict())
 
 
-@api_v1.route("/warehouse/identifier-sets/<name>", methods = ['PUT'])
+@api_v1.route("/warehouse/identifier-sets/<name>/<use>", methods = ['PUT'])
```
tsibley, line 196: I'd suggest not putting `use` into the URL here, as it's not part of the resources' PK and doesn't fit into the URL/resource hierarchy. Instead, `use` can be passed like `description` as a parameter. The param will obviously be required on creation, but could be allowed to be omitted for updates (for example, of just `description`). (The client can then choose if its ok with assuming pre-existence or not.)

lib/id3c/api/datastore.py
```diff
@@ -256,38 +256,55 @@ def fetch_identifier_set(session: DatabaseSession, name: str) -> Any:
...
 
         else:
             cursor.execute("""
-                insert into warehouse.identifier_set (name)
-                    values (%s)
+                insert into warehouse.identifier_set (name, use)
+                    values (%s, %s)
                     on conflict (name) do nothing
-                """, (name,))
+                """, (name, use))
```
tsibley, line 289: I'm on the fence about whether its best to make `use` a set-once attribute or leave it up to usage/convention. I see the upside of not letting an identifier set change uses (that's potentially confusing if the identifiers have already been used for that purpose!), but outside of this API, there's no restriction on updating it, so it's a little weird to enforce it here but not elsewhere. My thinking is leaning towards not requiring it to be set-once.

If we _do_ require it to be set-once, then I think a 409 Conflict response should be returned when the value passed in doesn't match the database (during both creation and update). Otherwise a client doesn't get useful information about the state of the resource and has to conditionalize on a fetch of the identifier set instead of just making one idempotent PUT request. I also think that if we require it set-once in the API, we should strongly consider propagating that restriction deeper into the database itself.

tsibley (changes requested):

lib/id3c/api/datastore.py
davereinhart: That makes sense. I didn't consider that the API is already locked down to the equivalent database permissions, so currently only postgres user or role members will be able to use this endpoint. So I agree we can allow updates to `use` here.

lib/id3c/api/routes.py
davereinhart: Ok, I agree it makes more sense to pass it as a form parameter. I'll update this shortly.

tsibley:
Ah, it'd also be good to update _lib/id3c/api/static/index.html_ with the new and updated endpoints.

lib/id3c/api/routes.py
davereinhart: This is close to working as intended, with the exception of allowing for updating `description` for an existing `identifier_set` without updating its `use`. Because `use` is a required field in the table, it's also required for any insert statement. 

I'm generating the sql statement dynamically now based on what form parameters are sent, so I'm not sure how to handle the case where only `description` is being updated. I started trying to switch to an update statement in those cases but that creates additional challenge of how to handle when `name` is not found (in which case it should attempt an insert and throw an error because `use` is not provided).

davereinhart: UPDATE: commit f0b5ca162cabc9ec4d8e9b4f96dd6b0140298397 seems to be working the way it needs to. It required a bit more logic and code plus an additional select query but now handles all possible combinations of `use` and `description` form parameters.

lib/id3c/api/datastore.py
```diff
@@ -259,35 +259,67 @@ def fetch_identifier_set(session: DatabaseSession, name: str) -> Any:
...
+        conflict_where_str = " or ".join([("identifier_set." + k + " <> excluded." + k) for k in fields.keys()])
+
+        sql_str = "insert into warehouse.identifier_set ({}) \
+                    values ({}) \
+                    on conflict (name) do update set {} \
+                    where {}".format(columns_str, values_str, conflict_update_str, conflict_where_str)
+
+        sql_str = sql_str.replace("''","null")
+        try:
+            cursor.execute(sql_str)
```
tsibley, line 302: This introduces a SQL injection vulnerability by unsafely handling user input, specifically, the submitted field values.

The field names are risky, but technically ok for now, as our code is the sole provider of those. However, it's better practice for a function like this that accepts an arbitrary field list to treat it like untrusted input and handle accordingly. The reason is that the calling code is safe for now, but may change to not be safe in the future, and it won't be obvious at that point that a security vulnerability is being introduced.

Safe construction of a dynamic SQL statement can be done with the functions and classes provided by the [psycopg2.sql module](https://www.psycopg.org/docs/sql.html). There are several examples in ID3C and our other code bases.

Separately, I think this function might be a bit easier to read by introducing some conditional branching and using multiple formatted SQL statements (as it was before) instead of trying to fit everything into one.

tsibley (changes requested):

lib/id3c/api/datastore.py
davereinhart: Thanks for pointing that out! I had switched to using `str.format(...)` in an attempt to generate a single SQL statement dynamically, but I can see why that's a bad idea. The link you provided has the information I needed.

I refactored again using a conditional statement with multiple SQL statements, but stuck with the existing pattern of hard coding SQL identifiers and passing query arguments, which should make it safe, more readable, and consistent with the other `datastore.py` functions.

The case where no `use` value is passed ended up requiring an additional select query to check if the record exists. If only the `description` is being updated on an existing record, switching to an update query made the most sense. The other challenge was keeping the function idempotent and returning a 204 code if nothing was changed. This required additional where clauses on `conflict (name) do update` and use of `coalesce` to compare null `description` values.

davereinhart:
Repushed with a minor change to include the new `identifier_set.use` column in the `find_identifier` db utility function.

lib/id3c/api/datastore.py
```diff
@@ -218,12 +218,12 @@ def fetch_identifier_sets(session: DatabaseSession) -> Any:
...
     Fetch all identifier sets from the backing database using *session*.
 
-    Returns a list of named tuples with ``name`` and ``description``
+    Returns a list of named tuples with ``name``, ``description``, and ``use``
     attributes.
     """
     with session, session.cursor() as cursor:
         cursor.execute("""
-            select name, description
+            select name, description, use
```
kschwabefry, line 226: Would we ever want use to be a list of values, if an identifier set had multiple `use`s?

kschwabefry (commented):
Looks really good. Left a few comments for discussion below.

lib/id3c/api/static/index.html
```diff
@@ -91,7 +91,12 @@ <h3 class="code">GET /v1/warehouse/identifier-sets/<em>&lt;name&gt;</em></h3>
 
     <h3 class="code">PUT /v1/warehouse/identifier-sets/<em>&lt;set&gt;</em></h3>
     <p>Make a new identifier set <em>name</em>.  Idempotent if the set already
-    exists.  An optional <em>description</em> parameter may be provided.
+    exists.  Optional <em>description</em> parameter may be provided. A <em>use</em>
+    parameter is required for new sets and is optional if only updating
+    <em>description</em> on an existing set.
```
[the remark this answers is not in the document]
davereinhart: I think it's still idempotent in the sense that it if you call it multiple times with the same arguments, it will not update the record on each call. If something changed (either a new record, or existing record where use or description changed) it updates the record and returns a 200 code, but if everything is the same as the existing record it returns a 204 (No Content)

lib/id3c/api/datastore.py
```diff
@@ -259,35 +259,96 @@ def fetch_identifier_set(session: DatabaseSession, name: str) -> Any:
...
     Create a new identifier set *name* in the backing database using *session*
-    if it doesn't already exist.
+    if it doesn't already exist, or update if it does exist.
 
-    If *description* is provided as a keyword argument, its value is
-    set/updated in the database.
+    If *use* and/or *description* are provided as keyword arguments, their values
+    are set in the database. Becuase *use* is a required field in the target table,
+    if it is not provided as a keyword argument the query will attempt to retrieve
+    its value from an existing record.
```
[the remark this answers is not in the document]
davereinhart: That would have been my first inclination, to create separate endpoints for creates and updates, but I was following the pattern already established for the optional `description` argument here:
https://github.com/seattleflu/id3c/blob/ca2d7f127b8e86818a469ba18b74aeef5e05844d/lib/id3c/api/datastore.py#L264-L268

There was some previous discussion on this PR about including `use` as a keyword argument, but it isn't showing up any more because I marked it as resolved (I'm inclined to _not_ do that going forward)

lib/id3c/api/datastore.py
davereinhart: Good question, I'm not really sure. Would this be the case for `tiny swabs` set?  I believe all of the other sets have a single use so I'm not sure if that's just an outlier or not. There's a related slack thread in the software channel from early July about the use type for each existing set.

kschwabefry: It can get a little murky. For tiny swabs, they are both collection ids and aliquot ids - but I'm not sure we'd want to bring concept of aliquot ids into our data model (or leave that to the LIMS). 
From the [slack thread](https://seattle-flu-study.slack.com/archives/C0173B4S1LY/p1625586692016000) it does seem like the existing cases of a identifier set mapping to multiple uses are due to data errors. 

Perhaps let's keep this enforced as a single value and update later if required.

kschwabefry (approved):

lib/id3c/api/static/index.html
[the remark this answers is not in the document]
kschwabefry: I think youre right. 👍

davereinhart:
Deployed

PR #236, opened 2021-07-14
Tests: Check that our Sqitch changes use transactions properly

tsibley:
Deploy and revert scripts must use a final commit and no other premature
commits.  Verify scripts must use a final rollback and no other
premature rollbacks.  All scripts must start with begin.

This test was written to catch future instances of actual mistake made
in our shipping/views change in the id3c-customizations repo.  The
possibility for the same mistake exists in this repo too.

### DIFF
--- a/tests/sqitch-transactions.py
+++ b/tests/sqitch-transactions.py
@@ -1,0 +1,34 @@
+import pytest
+from pathlib import Path
+from sqlparse import split, parse
+from sqlparse.tokens import Keyword
+
+topdir = Path(__file__).resolve().parent.parent
+
+sql_files = list(sorted(topdir.glob("schema/**/*.sql")))
+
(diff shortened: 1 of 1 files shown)


kschwabefry (approved):
Right on, thanks for adding a test for this.
Thinking out loud, maybe we could do something similar to check that the sqitch plan has a tag.

tsibley:
> Thinking out loud, maybe we could do something similar to check that the sqitch plan has a tag.

Yes, absolutely, with one caveat:

Since we don't want to require that every commit introduces a tag, we'd want to skip this test on branches and only run it on `master`, _including_ on our CI jobs which run against the merge of `master` and the PR branch (but not the parallel CI job which runs on the tip of the PR branch).

PR #238, opened 2021-07-16
Tests: Check that on master the Sqitch plan always ends with a tag

tsibley:
Tags are how Sqitch enables reworking of changes (such as our
shipping/views change) and are intended to mark the point when a schema
is "released" and thus existing change scripts can't be edited in place;
instead adding a new change or rework is necessary.

In our case, we "release" our schema when we deploy it to production,
and we typically deploy to production on every merge to master.  This
means we very rarely have periods where a tag isn't necessary, so
require one via our CI tests.

### DIFF
--- a/tests/sqitch-plan.py
+++ b/tests/sqitch-plan.py
@@ -1,0 +1,44 @@
+import pytest
+import re
+from os import environ
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


tsibley:
Prompted by @kschwabefry's suggestion in https://github.com/seattleflu/id3c/pull/236#pullrequestreview-708576558.

kschwabefry (approved):

PR #240, opened 2021-07-19
geocode: Support specifying explicit SmartyStreets licenses to use

tsibley:
This is sometimes necessary if more than one license/plan is associated
with an account.  The default behaviour is unchanged.  Requires a
version bump on the SmartyStreets SDK dep, as the license handling was
only added in 4.6.0.

I made this change while making other changes to locally test
SmartyStreets' new "enhanced" lookups.  This knob can be necessary to
resolve ambiguity when there are multiple subscriptions active, which
seems like a situation that could easily arise at some point in
production, as it's determined by our SmartyStreets account, not
anything we deploy.  It'd be nice for it to already exist before its
needed to fix a broken system!


kschwabefry (approved):
Nice to have in advance

tsibley:
Deployed.

PR #241, opened 2021-07-20
Add verify/barcode-uses endpoint to API

davereinhart:
Add verify-barcode-use endpoint to API for use with Laboratory Information
Management System (LIMS) and/or other system integration, for automation of QC processes.

This endpoint takes a JSON array of objects as input, with each object containing a barcode
and its associated use type to verify (i.e. `sample`, `collection`, `clia`). The response
will be a JSON array of the same objects with boolean values for `barcode_found` to indicate
whether the barcode exists in ID3C and `use_match` to indicate whether the input use type
matches what is stored in ID3C.

### DIFF
--- a/lib/id3c/api/datastore.py
+++ b/lib/id3c/api/datastore.py
@@ -178,6 +178,38 @@ def store_fhir(session: DatabaseSession, document: str) -> None:
         except (DataError, IntegrityError) as error:
             raise BadRequestDatabaseError(error) from None
 
+@export
+@catch_permission_denied
+def verify_barcode_use_list(session: DatabaseSession, barcode_use_list: list) -> Any:
+    """
+    Check the given *barcode_use_list* containing objects with ``barcode`` and ``use``
+    keys and values to verify that each barcode exists in the backing database and that the
+    given use matches the stored use.
+
+    Returns a list of objects in the same order as the input, with each object including the
+    ``barcode`` (string) and ``use`` (string) being verified, ``barcode_found`` (boolean)
+    indicating whether the given barcode exists, and ``use_match`` (boolean) indicating whether
+    the given use matches the stored use. The ``use_match`` value will be `null` if the barcode
+    does not exist.
+    """
+    barcode_use_tuples = [(bu["barcode"],bu["use"]) for bu in barcode_use_list]
+    args_str = ','.join(['%s'] * len(barcode_use_tuples))
+    sql = "select q.barcode, q.use, \
+            case \
+                when identifier.barcode is not null then true else false \
+            end as barcode_found, \
+            case \
+                when identifier_set.use IS NULL then null \
+                when q.use::citext=identifier_set.use then true \
+                else false \
+            end as use_match \
+            from (values {}) as q (barcode, use) \
+            left join warehouse.identifier on q.barcode::citext = identifier.barcode \
+            left join warehouse.identifier_set using (identifier_set_id)".format(args_str)
+
+    result = session.fetch_all(sql, tuple(barcode_use_tuples))
+    return result
+    
 
 @export
 @catch_permission_denied
--- a/lib/id3c/api/routes.py
+++ b/lib/id3c/api/routes.py
@@ -2,6 +2,7 @@
 API route definitions.
 """
 import json
+from jsonschema import validate, ValidationError
 import logging
 import pkg_resources
 from flask import Blueprint, jsonify, request, send_file
@@ -165,6 +166,30 @@ def verify_barcode_uses(*, session):
 
     LOG.debug(f"Validating «{barcode_use_list}»")
 
+    schema = {
+        "type": "array",
+        "items": {
+            "type": "object",
+            "properties": {
+                "barcode": {
+                    "type": "string"
+                },
+                "use": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "barcode",
+                "use"
+            ]
+        }
+    }
+
+    try:
+        validate( instance = barcode_use_list, schema = schema )
+    except ValidationError as e:
+        return str(e), 400
+    
     result = datastore.verify_barcode_use_list(session, barcode_use_list)
     return jsonify([ row._asdict() for row in result ])
     
--- a/lib/id3c/api/routes.py
+++ b/lib/id3c/api/routes.py
@@ -144,7 +144,30 @@ def receive_fhir(*, session):
 
     return "", 204
 
+@api_v1.route("/verification/barcode-uses/verify", methods=['POST'])
+@content_types_accepted(["application/json"])
+@check_content_length
+@authenticated_datastore_session_required
+def verify_barcode_uses(*, session):
+    """
+    Validate a list of barcodes and use types.
+    
+    POST /barcode-uses with a JSON array of objects containing ``barcode`` and
+    ``use`` keys with string values.
 
+    Response will be a JSON array in the same order as input array, with objects containing
+    ``barcode`` (string) and ``use`` (string) from input array plus ``barcode_found`` (boolean)
+    to indicate if the barcode exists in the database, and ``use_match`` (boolean) to indicate
+    if the use from the input matches what is stored in the database (the ``use_match`` value
+    will be `null` if ``barcode_found`` is `false`).
+    """
+    barcode_use_list = request.get_json()
+
+    LOG.debug(f"Validating «{barcode_use_list}»")
+
+    result = datastore.verify_barcode_use_list(session, barcode_use_list)
+    return jsonify([ row._asdict() for row in result ])
+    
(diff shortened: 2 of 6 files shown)


davereinhart:
This PR is waiting on #234 to be merged, and will then be rebased prior to review.

lib/id3c/api/datastore.py
```diff
@@ -178,6 +178,39 @@ def store_fhir(session: DatabaseSession, document: str) -> None:
...
+    barcode_use_dict = {row[0]: row[1] for row in barcode_use_records}
+
+    result = []
+    for bu in barcode_use_list:
+        found = bu['barcode'] in barcode_use_dict
+        result_row = {
+            'barcode': bu['barcode'],
+            'use': bu['use'],
+            'barcode_found': found,
+            'use_match': bu['use']==barcode_use_dict[bu['barcode']] if found else None
```
tsibley, line 210: Quick note as I'm skimming: This should casefold the two sides of this comparison, otherwise this defeats the `ci` in `citext` by moving the comparison outside of the database and into Python.

lib/id3c/api/datastore.py
```diff
@@ -178,6 +178,39 @@ def store_fhir(session: DatabaseSession, document: str) -> None:
...
+    barcode_use_records = session.fetch_all("""
+        select identifier.barcode, identifier_set.use from warehouse.identifier 
+        join warehouse.identifier_set using (identifier_set_id)
+        where barcode = ANY(%s)""", (barcodes,))
+
+    barcode_use_dict = {row[0]: row[1] for row in barcode_use_records}
+
+    result = []
+    for bu in barcode_use_list:
+        found = bu['barcode'] in barcode_use_dict
```
tsibley, line 205: Oh, ditto this line too, actually, with `in`. Might be easier to just keep this matching in the database and produce more of the returned data structure there.

lib/id3c/api/routes.py
```diff
@@ -144,6 +144,30 @@ def receive_fhir(*, session):
 
     return "", 204
 
+@api_v1.route("/warehouse/verify-barcode-uses", methods=['POST'])
```
tsibley, line 147: It's not clear to me why this is nested under `/warehouse/…` when it's not a `warehouse` resource but an abstract RPC endpoint (what this [useful guide on naming RESTful resources](https://restfulapi.net/resource-naming/) calls the "controller" archetype).

lib/id3c/api/datastore.py
davereinhart: Good catch!  I refactored to have the matching occur in the query instead of in Python.

lib/id3c/api/routes.py
davereinhart: Ok, my only reason for putting it there is that it's interfacing with warehouse tables, but a new category makes sense. `/verify/…` might be a good option.

lib/id3c/api/datastore.py
davereinhart: Refactored to match as `citext` in SQL.

lib/id3c/api/routes.py
kschwabefry: I think typically if we want to use that controller pattern, we may want to name it with the verb at the end, something like `/identifier-use/verify` or `/barcode-use/verify`

Alternatively, we could restructure this enpoint a bit if we want it to be a more straightforward GET request for resources. Rather than doing the "verification" in the API, we could instead provide an endpoint that took the identifier, and returned the associated identifier use. That could be something like `/warehouse/identifiers/{id}/identifier-use` . I know you already discussed this with Jeremy though and I'm guessing that led to doing the verification within the API?

davereinhart: My thought process with `/verify/…` (or I could rename it to '/verification/…') was to have a grouping that we could add other endpoints to down the road. The [current endpoints](https://backoffice.seattleflu.org/production/api) are grouped into `/receiving/…` and `/warehouse/…` so I was trying to stick with a similar pattern.

Jeremy's original request was for one endpoint to verify a single barcode use and another endpoint to verify a list of barcodes and uses.  After chatting with him he agreed that for single barcodes he could just use the GET `/v1/warehouse/identifier/<id>` since that now also returns the `use` value. So this PR is just to add an endpoint for verifying a list, and he said his preferred response format is a list in the same order as the input, with a status for each barcode and use.

kschwabefry: Ok, that's fair if Jeremy wants this in addition to  `/v1/warehouse/identifier/<id>`, we can definitely add this then.

In regards to naming, I'm inclined to say that existing endpoints were grouped with `warehouse` and `receiving` prefixes to reuse the namespacing from the database. For this type of API, to follow naming conventions for RESTful APIs, I believe we'd want it to end in a verb to differentiate from endpoints that relate to documents or collections (see https://restfulapi.net/resource-naming/). If you'd like to namespace under `verification`, I think that would leave us with  `/verification/identifier-uses/verify` or `verification/barcode-uses/verify`.

davereinhart: That makes sense, I'll go ahead and rename it.  Jeremy is only going to be passing barcodes, but to keep things consistent with the `identifier` endpoints, I'll update it to accept full uuids or shortened barcodes and go with `/verification/identifier-uses/verify` for the name.

lib/id3c/api/routes.py
```diff
@@ -144,7 +144,30 @@ def receive_fhir(*, session):
...
+    to indicate if the barcode exists in the database, and ``use_match`` (boolean) to indicate
+    if the use from the input matches what is stored in the database (the ``use_match`` value
+    will be `null` if ``barcode_found`` is `false`).
+    """
+    barcode_use_list = request.get_json()
+
+    LOG.debug(f"Validating «{barcode_use_list}»")
 
+    result = datastore.verify_barcode_use_list(session, barcode_use_list)
+    return jsonify([ row._asdict() for row in result ])
```
kschwabefry, line 169: I know each row in the results has barcode_found and use_match. Do you think there would be benefit in inspecting the result and returning  an http response code accordingly here?

lib/id3c/api/routes.py
davereinhart: Tested a few different way but performance of the SQL query was much worse when trying to accommodate full UUIDs and shortened barcodes (since the input list could potentially contain either or both), so I left it as-is was to only check shortened barcodes. 

I renamed the endpoint `verification/barcode-uses/verify` as suggested and updated API documentation page accordingly.

lib/id3c/api/routes.py
davereinhart: Possibly, though I'm not sure what code we would want to apply since the result could always be a mix of verified and not verified. I can circle back to Jeremy on this so see what he prefers. 

I looked into possibility of implementing as a batch request, which could return a status code for each record, but that would require individual requests for each record combined into a single batch request, and most likely individual queries on the database for each record with results combined into a batch response. That seems much less efficient that this current method that uses a single select query.

kschwabefry: Fair point about batching. Sometimes a 207 is used for this (Multi-Status), indicating to look inside the response body for more detailed statuses about each operation or item. But, that's just one way of looking at it. I think the status could be interpretted in different ways. We could instead consider the response code indicative of the status of the verification operation, instead of pertaining to any individual item. We could keep a 200 response, as you have currently, if the verification operation completed successfully -- separating out the concern from whether barcodes were found/had matching uses. 

In that vein, it might be useful to the user to return a more specific 4XX type error, if the response is a malformatted array of json, or if doesn't contain the required keys, it would be ideal to return that to the user with a 400 response. I think currently it will just return a 5XX on any error.

davereinhart: Ok, 207 may be a good option here, with individual status codes for each identifier/use pair from the submitted list (200:OK or 404:Not Found). 404 could indicate either the barcode doesn't exist, or the use type doesn't match, so there'd still be a need to check the response details to determine which it is.

As far as the request body goes, if we want to implement as a batch API the standard format would be more like this:
```
[
    {
        "method" : "GET",
        "path"   : "https://backoffice.seattleflu.org/production/api/v1/verification/barcode-uses/verify",
        "body"   : "{\"barcode\":\"abcd1234\",\"use\":\"sample\"}"
    },
    ...
]
```
That's much more verbose than what I have right now, and would lend itself more toward checking each barcode/use individually and merging the results, which requires a lot more queries as noted above. So I'm not sure it makes sense to go that route unless we really do need to return individual status codes.

As far as returning 4XX error, I don't think it's possible for this to return a malformated array of json, but I can update it to return a 4XX if the request body is malformatted. The SQL query simply converts the incoming array into rows and left joins the necessary tables for verifying the values, so the response will always include everything from the passed array, even maintaining the order and capitalization despite the matching being case insensitive.

davereinhart:
Deployed.

PR #247, opened 2021-07-27
Consolidate upsert_sample function

davereinhart:
The `upsert_sample` function is implemented in three different places. This
consolidates those to a single standardized version of the function to reduce
redundancy. Also updates existing `upsert_sample` function calls to match the
new definition, as well as some minor code clean up.

This function is also being moved out of the `id3c.cli.command.etl` subpackage to
`id3c.db` in anticipation of also using it for a API development. Other "upsert"
functions are currently located in `etl/__init__.py` but are currently only used in
`etl` modules, whereas `db` contains more generic functions/classes used in `id3c.api`
and `id3c.cli`.

### DIFF
--- a/lib/id3c/cli/command/etl/__init__.py
+++ b/lib/id3c/cli/command/etl/__init__.py
@@ -384,42 +384,6 @@ def upsert_presence_absence(db: DatabaseSession,
     return presence_absence
 
 
-def upsert_sample(db: DatabaseSession,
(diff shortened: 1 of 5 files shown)


kschwabefry (approved):
LGTM! I did some dry runs for the ETLs against my local and didn't notice any issues.
All the questions I had were answered once I finished reading through all three initial versions that got consolidated here.

PR #248, opened 2021-07-28
Add Tiny Swabs support to presence-absence ETL

seidelma:
As the [Trello card](https://trello.com/c/9HDoLMez/1136-tinyswabs-update-presence-absence-etl-to-accept-tiny-swabs-collection-ids) states: 
> expand to accept collection ids specifically from the tiny swabs identifier set
Do we want to explicitly check that this ETL run is for tiny swabs before we say `collections-uw-tiny-swabs` is a valid value for `identifier.set_name`?


kschwabefry:
> Do we want to explicitly check that this ETL run is for tiny swabs before we say `collections-uw-tiny-swabs` is a valid value for `identifier.set_name`?

It may be helpful to back up a step here. Here's how the process generally goes. Samplify will put records into our `receiving.presence_absence` table, and then the presence absence ETL processes records from `receiving.presence_absence` into `warehouse.presence_absence`. So, that is to say, the ETL run isn't tied specifically to tiny swabs, it's tied to having new records in `receiving.presence_absence` and will process the whole bundle, which often includes samples from different studies and different collection identifier sets.  Individuals samples within that presence_absence record could be associated with tiny swabs, but I think the way to determine it is to look up what identifier set the barcode is tied to. (in this case, if it's in `collections-uw-tiny-swabs`). Does that help answer your question?

Also, I hadn't caught this originally, but looking at the ETL more closely, I think we'll also need to change `update_sample()`, so that it can match on collection id instead of sample id for the tiny swabs samples.

davereinhart:
To keep things DRY, would it make sense to merge the `update_sample` function here with the `upsert_sample` function being consolidated in PR #247?  That one will already match on collection id or sample id. The main difference seems to be that this version is update only, and will not insert the sample record if it doesn't exist, but we could potentially handle that with an additional argument.

kschwabefry:
Hi @seidelma did you want to incorporate the changes to `update_sample()` in this PR as well? It wasn't called out specifically in the original card but I think it will need to match on collection id instead of just sample id for these swabs.

lib/id3c/cli/command/etl/presence_absence.py
```diff
@@ -285,9 +285,15 @@ def sample_identifier(db: DatabaseSession, barcode: str) -> Optional[str]:
     """
     identifier = find_identifier(db, barcode)
 
+    valid_identifiers = [
+        "samples",
+        "collections-uw-tiny-swabs",
```
kschwabefry, line 290: I think we want to leave out `collections-uw-tiny-swabs` because it should only be used for the pilot, and marked experimental/ignored by this ETL.

seidelma: OK, I'll fix that so it doesn't get used.

davereinhart:
Looks good! Ran some tests on my local db, which required several steps and threw expected errors along the way:
 
- cleared processing log for last record in `receiving.presence_absence`, then ran `id3c etl presence-absence` with no issues (confirms it still works with sample identifiers)
- minted new identifiers in each of the collection-uw-tiny-swabs sets (including the one used in pilot study)
- inserted new sample records with those same identifiers in the `collection_identifier` column
- copied the `document` json from the last record in `receiving.presence_absence` and modified it by replacing the "investigatorId" values with the newly minted identifiers' barcodes, then inserted it as a new record in that table
- ran `id3c etl presence-absence` again and saw expected error for invalid set due to including identifiers from the pilot study
- moved those identifiers to the `home` set and ran ETL command again. Encountered error "Unable to determine target presence given" for the last object in the presence_absence document json, which was intentionally skipped during the initial ETL because it's original "investigatorId" value was "Hap1_Gblock" which is not a valid barcode.
- removed that object from the presence_absence record's json document and re-ran the ETL, at which point it succeeded

kschwabefry (approved):
Looks great!

PR #249, opened 2021-07-29
Retry failed REDCap API calls

davereinhart:
REDCap API has a bug that incorrectly returns 200 status code and HTML
response text in HTML format containing an "unknown error" and "This may
have occurred if you have opened multiple browser tabs of the same REDCap
page". Requests that receive this reponse often succeed with additional
attempts.

This change will search for that specific response and retry
the request up to 10 times before proceeding. If the limit of 10 retries
is reached and the same response is still being recieved, the function
will continue as it did before with the invalid response text. That will
result in the same exception that was occurring previously, which happens
when the response text is being processed within other parts of the code
that call this function. This error handling may be improved in the future,
but that behavior is being left in place to ensure alerts are still surfaced
in the same manner for now.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -486,7 +486,20 @@ class Project:
             'format': format,
         }
 
-        response = requests.post(self.api_url, data=data, headers=headers)
+        retry_count = 0
+        max_retry_count = 10
+
+        # Added as workaround for REDCap API bug which incorrectly returns 200 status code 
+        # and HTML response with "unknown error" message and substring included below, which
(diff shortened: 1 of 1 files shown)


davereinhart:
[Retry failed Redcap API calls](https://trello.com/c/ZS1YRZ2r/1157-retry-failed-redcap-api-calls)

davereinhart:
[Retry failed Redcap API calls](https://trello.com/c/ZS1YRZ2r/1157-retry-failed-redcap-api-calls)

kschwabefry (approved):

davereinhart:
Deployed.

PR #250, opened 2021-07-29
Add accessioning API to add/update sample records

davereinhart:
Adding new API endpoints for single and batch create and update
of sample records, to replace current process of ingestion from
aliquoting sheets.

### DIFF
--- a/lib/id3c/api/datastore.py
+++ b/lib/id3c/api/datastore.py
@@ -395,9 +395,9 @@ def store_sample(session: DatabaseSession, sample: dict) -> Any:
     failed validation.
     """
     with session:
-        sample_barcode = sample.pop("sample", None)
+        sample_barcode = sample.pop("sample_id", None)
         sample_identifier = find_identifier(session, sample_barcode) if sample_barcode else None
-        collection_barcode = sample.pop("collection", None) 
+        collection_barcode = sample.pop("collection_id", None) 
         collection_identifier = find_identifier(session, collection_barcode) if collection_barcode else None
         
         result = {
@@ -406,10 +406,7 @@ def store_sample(session: DatabaseSession, sample: dict) -> Any:
         }
 
         # validate barcodes
-        if not sample_barcode and not collection_barcode:
-            result["status"] = "validation_failed"
-            result["details"] = f"sample barcode or collection barcode required"
-        elif sample_barcode and not sample_identifier:
+        if sample_barcode and not sample_identifier:
             result["status"] = "validation_failed"
             result["details"] = f"sample barcode «{sample_barcode}» not found"
         elif sample_identifier and sample_identifier.set_use != 'sample':
@@ -425,12 +422,12 @@ def store_sample(session: DatabaseSession, sample: dict) -> Any:
         if result.get("status", None) == "validation_failed":
             return result
 
-        collected_date = sample.get("date", None)
+        collected_date = sample.get("collection_date", None)
 
         # When updating an existing row, update the identifiers only if the record has both
         # the 'sample_barcode' and 'collection_barcode' keys
         should_update_identifiers = True if (sample_identifier and collection_identifier) else False
-        
+
         sample, status = upsert_sample(session,
                 update_identifiers          = should_update_identifiers,
                 identifier                  = sample_identifier.uuid if sample_identifier else None,
--- a/lib/id3c/api/datastore.py
+++ b/lib/id3c/api/datastore.py
@@ -411,15 +411,16 @@ def store_sample(session: DatabaseSession, sample: dict) -> Any:
             result["details"] = f"sample barcode «{sample_barcode}» not found"
         elif sample_identifier and sample_identifier.set_use != 'sample':
             result["status"] = "validation_failed"
-            result["details"] = f"barcode «{sample_barcode}» has use type «{sample_identifier.set_use}»"
+            result["details"] = f"barcode «{sample_barcode}» has use type «{sample_identifier.set_use}» instead of expected use type «sample»"
         elif collection_barcode and not collection_identifier:
             result["status"] = "validation_failed"
             result["details"] = f"collection barcode «{collection_barcode}» not found"
         elif collection_identifier and collection_identifier.set_use != 'collection':
             result["status"] = "validation_failed"
-            result["details"] = f"barcode «{collection_barcode}» has use type «{collection_identifier.set_use}»"
+            result["details"] = f"barcode «{collection_barcode}» has use type «{collection_identifier.set_use}» instead of expected use type «collection»"
 
         if result.get("status", None) == "validation_failed":
+            LOG.debug(f"Validation failed for {sample} with details: {result.get('details')}")
             return result
 
         collected_date = sample.get("collection_date", None)
@@ -428,16 +429,22 @@ def store_sample(session: DatabaseSession, sample: dict) -> Any:
         # the 'sample_barcode' and 'collection_barcode' keys
         should_update_identifiers = True if (sample_identifier and collection_identifier) else False
 
-        sample, status = upsert_sample(session,
-                update_identifiers          = should_update_identifiers,
-                identifier                  = sample_identifier.uuid if sample_identifier else None,
-                collection_identifier       = collection_identifier.uuid if collection_identifier else None,
-                collection_date             = collected_date,
-                encounter_id                = None,
-                additional_details          = sample)
+        try:
+            sample, status = upsert_sample(session,
+                    update_identifiers          = should_update_identifiers,
+                    identifier                  = sample_identifier.uuid if sample_identifier else None,
+                    collection_identifier       = collection_identifier.uuid if collection_identifier else None,
+                    collection_date             = collected_date,
+                    encounter_id                = None,
+                    additional_details          = sample)
 
-        result["sample"] = sample
-        result["status"] = status
+            result["sample"] = sample
+            result["status"] = status
+        except Exception as e:
+            result["status"] = "upsert_error"
+            result["details"] = f"error upserting sample record: {str(e)}"
+            LOG.debug(f"Error on upsert_sample: {str(e)}")
+
         return result
 
 @export
--- a/lib/id3c/api/datastore.py
+++ b/lib/id3c/api/datastore.py
@@ -10,9 +10,11 @@ from typing import Any
 from uuid import UUID
 from werkzeug.exceptions import Forbidden, NotFound
 from .. import db
+from ..db import find_identifier
 from ..db.session import DatabaseSession
 from .exceptions import AuthenticationRequired, BadRequest
 from .utils import export
+from ..cli.command.etl.manifest import upsert_sample
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


davereinhart:
There's some additional work to do on the accessioning API, but wanted to get this PR up in draft mode for preliminary review to make sure we're in agreement on the approach. The two options we discussed during our call were asynchronous with insert/updates directly on the warehouse.sample table vs. synchronous with inserts on the receiving.manifest table. The branch currently does the former, which allows us to implement real-time data validation rather than accepting everything that is submitted as the current ingestion process does. 

I think the biggest advantage to this approach is that the LIMS users will get immediate data validation on each record that they can address right away instead of the status quo where we receive an alert and ask them to fix data issues later. The only downside I can see is that we won't have a historical record of every version of each record submitted, and we wouldn't be able to rerun ETL processes, because we'd no longer be using an ETL for the aliquoting sheets. That may not be a problem if we're only accepting validated data through the API.

There are 2 other ETLs that insert/update warehouse.sample records: FHIR and enrollments. Of those, only the FHIR ETL is in use, but both would continue to function as they currently do. The `upsert_sample` function that is used by all of these processes (including this new API) merges the document json field when updating existing records, so there should be no possibility of data being lost or overwritten.

kschwabefry:
@davereinhart thanks for sharing the thoughtful comments above. I agree that the benefits of synchronous feedback to the users, and the ability for users to self-correct data, might overwhelm the downsides here. I think we can additionally mitigate loss in ability to debug on our endby 1) providing thorough error messages and 2) logging request details (even if logs aren't retained indefinitely). We'd also definitely want to carry over the current validation happening in the ETL before inserting records, as you've started doing already.

(Secondary thought, do we want to create a database service account for the LIMS to use when its calling APIs?) 

Thinking about how data changes from multiple API calls or multiple processes might interact is interesting. Can you say a little more about this: 
> The upsert_sample function that is used by all of these processes (including this new API) merges the document json field when updating existing records, so there should be no possibility of data being lost or overwritten.

davereinhart:
The upsert function is designed to preserve data in the document field, by merging the new and existing json with `||`:
https://github.com/seattleflu/id3c/blob/1a4bcab4b72ef4bf182c2f6d40dc418ad56930f3/lib/id3c/cli/command/etl/manifest.py#L250-L255

This is noted in the comments on the `upsert_sample` function in `cli/command/etl/manifest.py`:
https://github.com/seattleflu/id3c/blob/1a4bcab4b72ef4bf182c2f6d40dc418ad56930f3/lib/id3c/cli/command/etl/manifest.py#L198-L200

I've maintained this behavior in the consolidated `upsert_sample` function in PR #247. As far as the other fields in the sample table, they are only overwritten if a non-null value is being passed into the function, which provides a good control mechanism for those. This behavior should be reviewed as part of PR #247, since that calls the consolidated `upsert_sample` function in different contexts where some fields are being updated and others are not.

davereinhart:
[Provide API for New LIMS: Submit an Aliquot Record for LIMS](https://trello.com/c/flb1DJG1/1142-provide-api-for-new-lims-submit-an-aliquot-record-for-lims)

lib/id3c/api/routes.py
```diff
@@ -283,6 +285,121 @@ def get_identifier_set_uses(*, session):
...
+    except ValidationError as e:
+        return str(e), 400
+
+    provenance = {
+        "source": request.referrer or request.remote_addr or "?",
+        "method": request.method,
+        "path": request.path,
+        "timestamp": datetime.now(),
+        "sha1sum": sha1(json.dumps(sample).encode()).hexdigest()
+    }
```
davereinhart, line 395: The `_provenance` for data coming from an API is going to look different from what is currently being used, which is specific to data coming from aliquoting worksheets:
```
"_provenance": {
        "row": 1234,
        "sheet": "aliquoting",
        "sha1sum": "a9993e364706816aba3e25717850c26c9cd0d89d",
        "workbook": "s3://bucket/folder/filename.xlsx"
}
```
Since we'll no longer have a row, sheet, or workbook, I replaces these with source (referrer or IP), method (i.e. "POST"), the endpoint path ("/v1/warehouse/sample") and current timestamp. The sha1sum will be generated per record instead of an entire file.

This seems like a relatively safe thing to change, but there could be some impacts downstream I'm not aware of. One example I found is in `shipping.reportable_condition_v1` view which parses workbook and sheet into columns. That view could be updated to accommodate this new provenance format, but wanted to get some input and finalize the format here before working on related changes.

lib/id3c/api/datastore.py
```diff
@@ -382,6 +383,61 @@ def fetch_identifier_set_uses(session: DatabaseSession) -> Any:
...
+def store_sample(session: DatabaseSession, sample: dict) -> Any:
+    """"
+    Validate the given *sample* and insert or update in the backing database.
+
+    Returns a list of in the same order as the input, with each object including
+    the ``sample_id`` (string), ``status`` (string) to indicate if inserted,
+    updated, or validation failed, and ``details`` to indicate reason for
+    failed validation.
+    """
+    with session:
```
seidelma, line 395: Won't this close the session when the context manager block exits? I'm curious about the interplay between that and `authenticated_datastore_session_required()` in lib/id3c/api/utils/routes.py, if any. It looks like the DatabaseSession `__exit__()` function calls the underlying psycopg2 connection `__exit__()` when the context manager is destroyed.

lib/id3c/api/routes.py
```diff
@@ -283,6 +285,147 @@ def get_identifier_set_uses(*, session):
...
+    JSON object with status (inserted, updated, or failed validation)
+    of the sample is returned.
+    """
+
+    sample = request.get_json()
+
+    LOG.debug(f"Creating/updating sample «{sample}»")
+
+    # everything except _provenance, which is auto-calculated and added below
+    schema = {
```
seidelma, line 308: Should we put this into some sort of config file so the API code doesn't need to change if the database does?

seidelma (commented):
Some questions about how the database session works, and where to put the schema definition.

lib/id3c/api/datastore.py
davereinhart: Yes, the session should close when the request is completed. The session is passed in from the corresponding `post_sample` function in `lib/id3c/api/routes.py`. I mostly followed the pattern established for the other endpoints, but just realized that I left out the `@export` and `@catch_permission_denied` decorators on this function. I'll fix that now!

lib/id3c/api/routes.py
davereinhart: Good idea! This JSON Schema doesn't match the database exactly, it's more to enforce how we want to receive the data (flat JSON) but then its structure gets modified before it's stored. It does clutter things up a bit to include the schema here though, so moving it to a new file would keep things more organized.  I'll add a new folder `lib/id3c/api/schemas` and load them from there.

davereinhart:
to test this branch locally:
```
PGSERVICE=seattleflu-local pipenv run flask run
```
Then in Postman, submit a POST request using valid sample identifier and collection identifier from your local database:
```
POST http://127.0.0.1:5000/v1/warehouse/sample
Content-Type:  application/json
Authorization:
    Basic Auth: [PostgreSQL username + pw]
```
With request body in this format:
```
{
    "sample_id": "[valid identifier from sample identifier-set]",
    "collection_id": "[valid identifier from collection identifier-set]",
    "sample_origin": "uw-reopening",
    "collection_date": "2021-06-23",
    "swab_site": "uw_club",
    "clia_id": "[clia barcode]",
    "received_date": "2021-06-23",
    "aliquot_a":"387743816",
    "aliquot_b":"387743843",
    "aliquoted_date": "2021-06-24",
    "rack_a":"TS01234567",
    "rack_b":"TS01234567",
    "rack_a_nickname":"COV2-123A",
    "rack_b_nickname":"COV2-123B",
    "notes": "testing",
    "swab_type": "ans",
    "collection_matrix": "dry"
}
```

lib/id3c/api/datastore.py
```diff
@@ -382,6 +383,63 @@ def fetch_identifier_set_uses(session: DatabaseSession) -> Any:
...
+            "collection_barcode": collection_barcode
+        }
+
+        # validate barcodes
+        if sample_barcode and not sample_identifier:
+            result["status"] = "validation_failed"
+            result["details"] = f"sample barcode «{sample_barcode}» not found"
+        elif sample_identifier and sample_identifier.set_use != 'sample':
+            result["status"] = "validation_failed"
+            result["details"] = f"barcode «{sample_barcode}» has use type «{sample_identifier.set_use}»"
```
kschwabefry, line 414: Not a blocker, but when validating the set use (for sample or collection ids), might be nice in the error message to include the expected set_use was, e.g.

`f"sample barcode «{sample_barcode}» has use type «{sample_identifier.set_use}»"`
or
`f"barcode «{sample_barcode}» has use type «{sample_identifier.set_use}» instead of expected type «sample»"`

lib/id3c/api/datastore.py
```diff
@@ -382,6 +383,63 @@ def fetch_identifier_set_uses(session: DatabaseSession) -> Any:
...
+        elif collection_barcode and not collection_identifier:
+            result["status"] = "validation_failed"
+            result["details"] = f"collection barcode «{collection_barcode}» not found"
+        elif collection_identifier and collection_identifier.set_use != 'collection':
```
kschwabefry, line 418: It looks like you dropped the validation of what collection identifier sets we expect here. Was that intentional to allow samples with any collection identifier set to be updated with this API?

lib/id3c/api/datastore.py
```diff
@@ -382,6 +383,63 @@ def fetch_identifier_set_uses(session: DatabaseSession) -> Any:
...
+            result["status"] = "validation_failed"
+            result["details"] = f"barcode «{collection_barcode}» has use type «{collection_identifier.set_use}»"
+
+        if result.get("status", None) == "validation_failed":
+            return result
```
kschwabefry, line 423: Adding in some logging on our end might be useful if we want to help debug issues.

lib/id3c/api/datastore.py
```diff
@@ -382,6 +383,63 @@ def fetch_identifier_set_uses(session: DatabaseSession) -> Any:
...
+@catch_permission_denied
+def store_sample(session: DatabaseSession, sample: dict) -> Any:
+    """"
+    Validate the given *sample* and insert or update in the backing database.
+
+    Returns a list of in the same order as the input, with each object including
+    the ``sample_id`` (string), ``status`` (string) to indicate if inserted,
+    updated, or validation failed, and ``details`` to indicate reason for
+    failed validation.
+    """
```
kschwabefry, line 396: Much of this code seems based on id3c's `manifest` command . I imagine we might keep both around for a while. I wonder if there's a way to structure this so that the two can share the common pieces -- it might take a bit of refactoring though, because of difference in the validation messaging when we want to set a specific response elements in an API response vs logging it/skipping DETs.
If the future vision is consolidation but its too big a piece of work at this time, maybe that could be added as a TODO

lib/id3c/api/routes.py
```diff
@@ -283,6 +266,71 @@ def get_identifier_set_uses(*, session):
...
+    # Transform aliquots data into array to match previously established format
+    aliquots = {key: value for key, value in sample.items() if key.startswith('aliquot_')}
+    if aliquots:
+        sample["aliquots"]=[]
+        for k, v in aliquots.items():
+            sample["aliquots"].append(sample.pop(k))
+
+    result = datastore.store_sample(session, sample)
+
+    return (jsonify(result), 200) if "sample" in result else (jsonify(result), 400)
```
kschwabefry, line 332: I wonder if we want to be a bit more nuanced here. We could return a 400 if we specifically see a user error like in validation, but we may want to fall back to a 500 in the more general case that we can't attribute to an issue with the user's request. https://restfulapi.net/http-status-codes/

kschwabefry (approved):
Really nice work here. A few questions below.

lib/id3c/api/datastore.py
davereinhart: Yes, definitely potential for consolidation here. Possibly creating another general function in the `db` subpackage (as was done recently with `upsert_sample`) to handle the validation process for this and the manifest ETL. I agree that would be a good improvement, but to meet the upcoming deadline I think we should keep it as-is and tackle that later.

lib/id3c/api/datastore.py
davereinhart: Yes, the idea here was to utilize the relatively new `use` column in the `identifier_set` table to avoid having to maintain a hardcoded list of `expected_identifier_sets` that needs to be updated every time a new set is added, as is being done in the manifest ETL function (it can also be replaced in that context but I didn't want to make any changes there that are unrelated to this branch)

lib/id3c/api/routes.py
davereinhart: Good suggestion. In testing it I wasn't able to get it to fail with anything that wasn't covered by JSON Schema or barcode validation, but will update it to return a 500 code in those cases.

lib/id3c/api/datastore.py
davereinhart: Added along with other suggestions in 896eb10effd21215b791305d919903c4414366b2

davereinhart:
Deployed.

lib/id3c/api/routes.py
kschwabefry: Thanks for adding the 500 response code. In case some ideas are helpful, here's a couple options for generating other errors:
* You can cause a `401` by supplying invalid credentials (or I think by shutting off your local database too)
* You can cause a `500` by modifying the application to make it throw an unhandled exception. It can be as basic as adding `raise Exception("Unknown error")` in `post_sample()`, or more nuanced -- like if you wanted to simulate a database issue, you could modify some of the SQL to reference a column that doesn't exist.

lib/id3c/api/datastore.py
kschwabefry: Makes sense! In inspecting more closely, it looks like the list in manifest.py does match perfectly with the identifier sets with use `collection`. I was thinking there might be additional filtering going on to identifier sets we expected to see come through the manifest, but since not, this makes a whole lot of sense. And will be one less piece to maintain. Sounds good to me!

lib/id3c/api/routes.py
davereinhart: Ok, thanks!  I realized that the `upsert_sample` function actually throws an exception in some cases, so I ended up wrapping that in a try/catch and sending back the error message in the API response. This is already deployed, but let me know if you other error handling suggestions other than what I added in c00de8098929e0d8230528baac58631ed5eb7ce9 and I can try to add them in when I implement a sample GET endpoint.

PR #252, opened 2021-08-05
Add label formats for UW tiny swabs home and observed sets

davereinhart:
Adding labelmaker layouts for two new identifier sets that were
added for UW tiny swabs: home and observed.

"UW TINY O" was used as label to indicate barcodes from observed set, and
"UW TINY U" was used for unobserved (home) set. This will also differentiate
them from the barcode labels from the pilot study, which used "UW TINY".

### DIFF
--- a/lib/id3c/labelmaker.py
+++ b/lib/id3c/labelmaker.py
@@ -339,6 +339,18 @@ class CollectionsUWTinySwabsLayout(LabelLayout):
     copies_per_barcode = 1
     reference = "seattleflu.org"
 
+class CollectionsUWTinySwabsHomeLayout(LabelLayout):
+    sku = "LCRY-2380"
+    barcode_type = 'UW TINY U'
+    copies_per_barcode = 1
+    reference = "seattleflu.org"
+
+class CollectionsUWTinySwabsObservedLayout(LabelLayout):
+    sku = "LCRY-2380"
+    barcode_type = 'UW TINY O'
+    copies_per_barcode = 1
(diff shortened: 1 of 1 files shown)


davereinhart:
[Tinyswabs: Create different identifer sets for home/kiosk](https://trello.com/c/eSKalpn7/1171-tinyswabs-create-different-identifer-sets-for-home-kiosk)

kschwabefry (approved):

lib/id3c/labelmaker.py
```diff
@@ -339,6 +339,18 @@ class CollectionsUWTinySwabsLayout(LabelLayout):
     copies_per_barcode = 1
     reference = "seattleflu.org"
 
+class CollectionsUWTinySwabsHomeLayout(LabelLayout):
+    sku = "LCRY-2380"
+    barcode_type = 'UW TINY U'
+    copies_per_barcode = 1
```
kschwabefry, line 345: Per Zack's slack message - should this be duplicates? https://seattle-flu-study.slack.com/archives/C023B12AZCK/p1628178216034000?thread_ts=1628106734.032600&cid=C023B12AZCK

(Not sure if there was another thread, this was the most recent I'm seeing)

davereinhart: Looked like it was going to be a mix for unobserved (singlets for dorms/rush, and duplicates home) so I set the default for singlets but can be overridden as needed.

davereinhart:
Deployed.

PR #253, opened 2021-08-13
Add sample editor role to database

davereinhart:
Initially for use with LIMS, this role will be used to insert and
update records in the `warehouse.sample` table as well as verify
barcodes and uses in `warehouse.identifier` and `identifier_set` tables
via ID3C API.

### DIFF
--- a/schema/revert/roles/sample-editor/grants.sql
+++ b/schema/revert/roles/sample-editor/grants.sql
@@ -1,0 +1,9 @@
+-- Revert seattleflu/schema:roles/sample-editor/grants from pg
+
+begin;
+
+revoke all on database :"DBNAME" from "sample-editor";
+revoke all on schema receiving, warehouse, shipping from "sample-editor";
+revoke all on all tables in schema receiving, warehouse, shipping from "sample-editor";
+
+commit;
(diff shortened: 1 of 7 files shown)


schema/revert/roles/sample-editor/grants.sql
```diff
@@ -0,0 +1,24 @@
+-- Revert seattleflu/schema:roles/sample-editor/grants from pg
+
+begin;
+
+revoke select, insert, update
+    on warehouse.sample
+  from "sample-editor";
```
kschwabefry, line 7: Typically we go a bit broader with the revert script, see this revert template: https://github.com/seattleflu/id3c/blob/master/schema/templates/revert/role/grants.tmpl . (Note that the template also can get used when you create the role with https://github.com/seattleflu/id3c/blob/master/dev/add-role )

davereinhart: Ok, that makes sense. I didn't realize there was a template, I was just following `role/manifest-uploader` as an example.

kschwabefry (approved):

davereinhart:
Sqitch deployment done on production db.

PR #256, opened 2021-09-10
Add date to sample API to be written to details

davereinhart:
Date should be included in `warehouse.sample` jsonb column `details`.
Adding here to match process in manifest ETL and ensure this value
is present for queries (and alerts) that rely on it.


kschwabefry (approved):

davereinhart:
Deployed.

PR #257, opened 2021-09-29
Rename CLIA ID in sample details to match existing data

davereinhart:
The manifest ETL renames the `clia_id` from worksheet to `clia_barcode` before storing
in sample details. Updating API endpoint here to match, so that data submitted
through the API matches format of data previously ingested from the manifest worksheets.

The `clia_barcode` key name is also used in shipping views so this will ensure it
is populating correctly there and will appear in related automated Slack alerts.


davereinhart:
- [ ] Run an update query on existing warehouse.sample records to rename sample -> details ->> `clia_id` to `clia_barcode`

davereinhart:
Deployed.

PR #258, opened 2021-10-05
Update samplenotfound logging

kschwabefry:
Now that we also sometimes look up samples that have no sample
identifier but do have a collection identifier (currently,
tiny swabs), include collection identifier in the error output
as well for easier debugging.


lib/id3c/cli/command/etl/__init__.py
```diff
@@ -244,7 +244,8 @@ def find_sample(db: DatabaseSession, identifier: str, for_update = True) -> Any:
         """ + query_ending, (identifier,identifier,))
 
     if not sample:
-        LOG.info(f"No sample with identifier «{identifier}» found")
+        LOG.info(f"No sample found with identifier «{identifier}» or collection identifier «{collection_identifier}»")
```
davereinhart, line 247: There is no local variable collection_identifier here.

davereinhart (changes requested):

lib/id3c/cli/command/etl/__init__.py
kschwabefry: Fixing!

PR #260, opened 2021-10-11
Rename properties in sample details

davereinhart:
Renaming properties in sample.details JSON to match data that had been
previously ingested through ETL from aliquoting sheets, to provide more
consistency in the data. `collection_date` property is being removed
because it is redundant with `date` property and did not exist in record
ingested through the ETL.

After deploying this change, records that had been submitted through the API
prior to this will be updated via a query to have the correct property names.


davereinhart:
Deployed.

PR #269, opened 2021-11-10
Lookup address without secondary address field (appartment number)

cooperqmarshall:
Setting address['secondary'] field to a blank string will sometimes
return a valid address when the original lookup failed. These changes
preform address lookups via the us_street_lookup() and extract_address()
with address['secondary'] = '' after a failed lookups for both the
above methods with the original `secondary` field.

### DIFF
--- a/lib/id3c/cli/command/geocode.py
+++ b/lib/id3c/cli/command/geocode.py
@@ -294,6 +294,15 @@ def geocode_address(address: dict) -> dict:
 
     if not result:
         LOG.info(f"Invalid address: no response from SmartyStreets.")
+        '''
+        Incorrect user input in the secondary address field can cause lookups to fail.
+        Setting this field to a empty string and running the lookups again can fix
+        this issue.
+        '''
+        if address.get('secondary'):
+            LOG.info('Looking up address with empty secondary address field')
+            address['secondary'] = ''
+            return geocode_address(address)
 
     return parse_first_smartystreets_result(result)
 


cooperqmarshall:
Added some more logging to help with testing

truong97:
Hey Cooper, I touched base with Dave for his input and feedback. He recommended, for simplicity and reduced runtime/lookups, to:
1. first, check if there is a value in secondary or not

- right now, your changes creates a copy and makes secondary an empty string, regardless if it was empty originally. I  would remove those lines as well as storing the original and copy as an array. Instead create an if statement of some sort to check as a pre-filter of sort. 

2. Within that check (if it has a value in secondary), go head and run against it against us_street_lookup and send lookup

- I would remove the for loop
- make sure that the first lookup doesn't return none - you still want it to attempt another lookup but with secondary as empty string (i.e line 286)

3. if result (lookup.result) is empty and secondary is not null/empty, remove the secondary (empty string) as you intended and run against us_street_lookup and send_lookup 
4. Continue process (logging warnings if secondary as empty still invalid result, etc) and return `parse_first_smartystreets_result(result)`

Let me know if that makes sense.

cooperqmarshall:
> Let me know if that makes sense.

@truong97 Makes sense. I added a commit with the changes. Let me know if this is the correct interpretation of your feedback.

davereinhart:
@cooperqmarshall See if the approach here makes sense. It may be easier to have the function call itself after removing the secondary address: https://gist.github.com/davereinhart/42dd93fd4c31d9dc77fc9091d91ca6ad

cooperqmarshall:
@davereinhart Thank you for the Gists! I like your implementation with geocode_address_without_secondary(). The second time it gets called in recursion (if it gets that for), secondary will be blank so it will return None out of call stack. I will update the PR and test with this tomorrow.

cooperqmarshall:
@davereinhart Why did you call ``geocode_address_without_secondary()`` twice in your Gist? The first call after ``if lookup.street is None or not lookup.street.strip():`` seems like it would not be useful since it would never reach a lookup and return None since the street part of the address is not changed. The first call can be removed I think and running the lookups can happen after the US Street and free text lookups: https://gist.github.com/cooperqmarshall/3b2ce619e27957c6912ad48ae889bfab

davereinhart:
> @davereinhart Why did you call `geocode_address_without_secondary()` twice in your Gist? The first call after `if lookup.street is None or not lookup.street.strip():` seems like it would never reach a lookup and return None since the street part of the address is not changed. The first call can be removed I think and running the lookups can happen after the US Street and free text lookups: https://gist.github.com/cooperqmarshall/3b2ce619e27957c6912ad48ae889bfab

Good catch!  I wasn't 100% sure how that streets lookup worked, but if it's not using the secondary value it makes sense to remove that call to the new function. It would be useful to add some comments too, to describe the new function and why it's being called.

cooperqmarshall:
@davereinhart I pushed the changes with code comments but opted to not include a new function because I could not figure out a good return value for the case that `address.get('secondary') == false`. As it was, the function did not return anything for this case which was not what the definition `def geocode_address_without_secondary(address: dict) -> dict:` implied.

davereinhart:
@cooperqmarshall Looks good! There's actually no real value in adding a new function since you are only calling it from one place now. This approach makes sense and is even less code!

davereinhart (approved):

cooperqmarshall:
@davereinhart Thanks for the approval! Could we wait to merge until next Monday? Also, It's allowing me to merge this but should that be done by someone on the dev team when the time comes?

truong97:
Deployed as of Dec 1 2021.

PR #270, opened 2021-11-22
Prevent local smartystreets usage

davereinhart:
Changes to make it more difficult to use SmartyStreets credits during local development. This has occurred multiple times while testing REDCap DET ETL processes locally, causing unnecessary usage.

The geocoding cache file will now be required, and will fail if the specified file is not found. The other change is that the user will be prompted prior to any requests being sent to the SmartyStreets API, if geocoding in a non-production environment. This will help ensure that use of the geocoding service during local development is more visible and intentional.

### DIFF
--- a/lib/id3c/cli/command/__init__.py
+++ b/lib/id3c/cli/command/__init__.py
@@ -169,9 +169,9 @@ def pickled_cache(filename: str = None) -> Iterator[TTLCache]:
         try:
             with open(filename, "rb") as file:
                 cache = pickle.load(file)
-        except FileNotFoundError:
-            LOG.warning(f"Cache file «{filename}» does not exist; starting with empty cache.")
-            cache = empty_cache
+        except FileNotFoundError as error:
+            LOG.error(f"Cache file «{filename}» does not exist, please provide a valid cache.")
+            raise error from None
         else:
             assert isinstance(cache, TTLCache), \
                 f"Cache file contains a {cache!r}, not a TTLCache"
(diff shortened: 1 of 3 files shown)


davereinhart:
There may be a better method to check whether we're running in production or not, but the PGSERVICE environment variable seemed like a reliable indicator that easy to check.

kschwabefry (commented):
Thanks for adding these protections. Limiting usage outside of production except for when its done intentionally should really help keep costs down there. 

I think my hesitation is that usage of PG_SERVICE to check whether the environment is production feels indirect. It should cover current use cases in production. But it feels unintuitive, and I think we'd benefit from protecting against future use cases not defaulting to allow geocoding lookups in case they have a different PG service. What would you think about adding an environment variable 'ENVIRONMENT' or similar?

davereinhart:
> Thanks for adding these protections. Limiting usage outside of production except for when its done intentionally should really help keep costs down there.
> 
> I think my hesitation is that usage of PG_SERVICE to check whether the environment is production feels indirect. It should cover current use cases in production. But it feels unintuitive, and I think we'd benefit from protecting against future use cases not defaulting to allow geocoding lookups in case they have a different PG service. What would you think about adding an environment variable 'ENVIRONMENT' or similar?

That makes a lot more sense! I'll work on this after the holiday. Thanks for the review!

kschwabefry:
[Make it harder to use smartystreets lookups locally](https://trello.com/c/IaQAtCbh/1285-make-it-harder-to-use-smartystreets-lookups-locally)

davereinhart:
Changed it to check value of environment variable `GEOCODING_ENV` = "production". 

- [ ] `GEOCODING_ENV=production` will need to be added to each of the cron files that include geocoding, prior to this being deployed.

kschwabefry (approved):

davereinhart:
Deployed.

PR #271, opened 2021-11-29
Adding Github action to automatically publish to pypi

davereinhart:
Publishing id3c to pypi will help standardize how this package is being
installed compared to other dependencies. This will make the package simpler
include in multiple applications and pin to a specific version when needed.

One workflow change this will require is that commits on master branch will need to
be tagged to be published to pypi.org (whereas all commits on master branch will be
published to test.pypi.org). A version naming convention will need to be
decided on and integrated into the deployment workflow, but the specific format
chosen will not impact this PR.

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -1,0 +1,37 @@
+name: CD
+
+on: push
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python distributions to PyPI and TestPyPI
+    runs-on: ubuntu-18.04
+    steps:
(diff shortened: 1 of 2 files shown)


davereinhart:
A simple version naming convention was already established, so one option is to stick with that:
https://github.com/seattleflu/id3c/commit/538ba22f676922845be39bab1025acb9c67ae87d

kschwabefry (approved):

davereinhart:
Note to self: this needs to be rebased and updated to use Python 3.9

PR #272, opened 2021-11-30
Add timeout to REDCap requests

davereinhart:
A 300 second timeout is being added to all REDCap requests, which ultimately
all use the _fetch function. It has been confirmed there is 120 second timeout on the
server side so this should be plenty of extra time before giving up on a request.

The goal of this change is to prevent requests from hanging indefinitely, and to
throw an exception when a request exceeds 5 minutes.

### DIFF
--- a/lib/id3c/cli/redcap.py
+++ b/lib/id3c/cli/redcap.py
@@ -494,7 +494,7 @@ class Project:
         # in many cases succeeds with additional attempts.
         # -drr, 7/28/2021
         while retry_count <= max_retry_count:
-            response = requests.post(self.api_url, data=data, headers=headers)
+            response = requests.post(self.api_url, data=data, headers=headers, timeout=300)
             if response.status_code==200 and 'multiple browser tabs of the same REDCap page. If that is not the case' in response.text:
                 retry_count += 1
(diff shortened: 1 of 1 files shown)


davereinhart:
[Enable request timeouts across the board](https://trello.com/c/FlyMRvxm/1095-enable-request-timeouts-across-the-board)

lib/id3c/cli/redcap.py
```diff
@@ -494,7 +494,7 @@ def _fetch(self, content: str, parameters: Dict[str, str] = {}, *, format: str =
         # in many cases succeeds with additional attempts.
         # -drr, 7/28/2021
         while retry_count <= max_retry_count:
-            response = requests.post(self.api_url, data=data, headers=headers)
+            response = requests.post(self.api_url, data=data, headers=headers, timeout=300)
```
kschwabefry, line 497: Thanks for adding this in. Excited to see if this helps alleviate some of the persistent issues. One thought - with a timeout, would it be appropriate to retry the request?

davereinhart: Based on our discussion at Standup this morning, we'll implement it as-is for now and see how often it happens before deciding whether or not to retry on timeout.

kschwabefry (approved):

davereinhart:
Deployed.

PR #273, opened 2021-12-09
Adding ValueError classes for COVID vaccination status and manufacturer

davereinhart:
These will be used in `id3c-customizations` as part of the clinical ETL.
New data was recieved for SCH that includes COVID vaccination status and
manufacturer which will be ingested into ID3C. These errors will be raised
if unreconized values are encountered during the ETL process.


davereinhart:
[Reprocess Y1, Y2, Y3 SCH data](https://trello.com/c/BRuEXAKV/1205-reprocess-y1-y2-y3-sch-data)

kschwabefry (approved):

PR #276, opened 2022-01-07
Update FHIR ETL to include Immunization entries

davereinhart:
New entries are being added to the FHIR bundle for Immunization as part
of the updated UW Retrospectives ETL seattleflu/id3c-customizations#267.
    
These entries are using standard FHIR format and coding, and are linked to
a FHIR Patient, so in ID3C they will be ingested into the `individual.details`
column rather than `encounter.details`.
    
The `individual.details` jsonb column has not been utilized yet in ID3C, so this
will be the only data present until other ETLs are updated to include Immunization
entries or other resource entries related to Patients are are added to the FHIR
bundle and this ETL.

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -176,8 +176,9 @@ def process_encounter_bundle_entry(db: DatabaseSession, bundle: Bundle, entry: B
     LOG.debug(f"Processing Encounter Resource «{entry.fullUrl}».")
 
     related_resources = extract_related_resources(bundle, entry)
-
-    encounter = process_encounter(db, resource, related_resources)
+    immunization_resources = extract_immunization_resources(bundle)
+    
+    encounter = process_encounter(db, resource, related_resources, immunization_resources)
 
     if not encounter:
         raise SkipBundleError("Insufficient information in Bundle to create an encounter")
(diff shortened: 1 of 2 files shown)


davereinhart:
Related to https://github.com/seattleflu/id3c-customizations/pull/267

davereinhart:
[Reprocess DETs for SFS Clinical Data Pulls - UW Retros & Add in additional fields for ingestion](https://trello.com/c/IF9NYjpM/1226-reprocess-dets-for-sfs-clinical-data-pulls-uw-retros-add-in-additional-fields-for-ingestion)

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -176,8 +176,9 @@ def process_encounter_bundle_entry(db: DatabaseSession, bundle: Bundle, entry: B
     LOG.debug(f"Processing Encounter Resource «{entry.fullUrl}».")
 
     related_resources = extract_related_resources(bundle, entry)
-
-    encounter = process_encounter(db, resource, related_resources)
+    immunization_resources = extract_immunization_resources(bundle)
+    
+    encounter = process_encounter(db, resource, related_resources, immunization_resources)
```
kschwabefry, line 181: Is the processing of immunization resources split out from "related_resources" here because related resources is supposed to be encounter level, and immunization is supposed to be patient level?

davereinhart: > Is the processing of immunization resources split out from "related_resources" here because related resources is supposed to be encounter level, and immunization is supposed to be patient level?

Exactly, immunization resource entries are only linked to patients in this case. The would only be linked to an encounter if the encounter was the actual immunization event. In our case these are "historical immunizations" which do not reference a specific encounter: https://www.hl7.org/fhir/immunization-example-historical.json.html

kschwabefry (approved):
Changes look good to me. The second commit looks like its missing a message and description - can you go back and add one, or combine into one commit? 

It would be great to have others on the team take a look at this as well and https://github.com/seattleflu/id3c-customizations/pull/267 as well.

davereinhart:
Deployed

PR #279, opened 2022-01-14
Unpinning google-api-python-client version

davereinhart:
This package was previously pinned to v2.24.0 because a bug was
introduced in v2.25.0. That bug was fixed in version 2.26.1:
https://github.com/googleapis/google-api-python-client/releases/tag/v2.26.1, so it
is now safe to upgrade to the latest version (tested with v2.35.0).


davereinhart:
[Investigate and if possible unpin google-api-client dependency](https://trello.com/c/MP16u3Un/1219-investigate-and-if-possible-unpin-google-api-client-dependency)

kschwabefry (approved):

PR #280, opened 2022-02-01
Add Cascadia Identifiers Set

bencap:
In anticipation of the Cascadia study, this change adds the
identifier set `collections-cascadia-tiny-swabs-home`. Because
this set uses duplicate tiny barcodes, this collection set
inserts a blank barcode before each 4th barcode to keep the
last column empty and duplicate barcodes next to each other.

This change was tested locally by doing a dry run of minting barcodes for the new identifier, which output barcodes in the correct format (51 pairs of duplicates per sheet, with the last column blank, for a total of 1,020 barcodes per 20 sheets).

Update: Barcodes were tested and printed and look good to the requesters.

### DIFF
--- a/lib/id3c/labelmaker.py
+++ b/lib/id3c/labelmaker.py
@@ -356,7 +356,7 @@ class CollectionsScanTinySwabsLayout(LabelLayout):
     sku = "LCRY-2380"
     barcode_type = 'SCAN TINY'
     copies_per_barcode = 1
-    reference = "scanpublichealth.org" 
+    reference = "scanpublichealth.org"
 
 
 class CollectionsAdultFamilyHomeOutbreakTinySwabsLayout(LabelLayout):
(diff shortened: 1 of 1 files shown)


lib/id3c/labelmaker.py
```diff
@@ -378,6 +378,12 @@ class CollectionsAirsLayout(LabelLayout):
     copies_per_barcode = 2
     reference = "fredhutch.org"
 
+class CollectionsCascadiaTinySwabsHome(SamplesLayout):
```
bencap, line 381: Inheriting from SamplesLayout may not be the best solution here. Looking at the code, it may be better to add a `LCRY2380DuplicateLayout` class and then have `SamplesLayout` and `CollectionsCascadiaTinySwabsHome` inherit from that. This way, we can avoid duplication of the `blanks_before` function that this new collection inherits from `SamplesLayout` but not have this class inherit from another layout that is in use. Didn't add that at first since it is a bit more additional code, but is a simple change if whoever reviews this agrees it is a better solution.

bencap:
[Cascadia Study - configure new collection ID set](https://trello.com/c/IGso0e04/1242-cascadia-study-configure-new-collection-id-set)

lib/id3c/labelmaker.py
truong97: Hey @bencap, I agree that is a good idea, especially with the direction that a lot of studies may be going towards (tiny duplicates).

truong97 (approved):
Looks good! If you haven't done so, add the new identifier set to the database (use `id3c identifier set create` command)

bencap:
Deployed

PR #285, opened 2022-02-24
Rework gitignore to function as allowlist

davereinhart:
Reworking of .gitignore files to exclude everything up front, with
exceptions for files and folder that should be included. This is to
reduce the chance of pushing any files unintentionally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,29 +1,27 @@
-*.pyc
-__pycache__/
-/.mypy_cache/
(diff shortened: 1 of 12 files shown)


davereinhart:
[RNI followup from linelist on github](https://trello.com/c/DpFsHAGb/1326-rni-followup-from-linelist-on-github)

kschwabefry (approved):

PR #286, opened 2022-02-26
Silence Unwanted Smarty Streets Retry Output

bencap:
Occasionally, Smarty Streets will output an error to `stdout` that is not actionable, but since it sends to `stderr` instead of logging, it still shows up in our alerts channel. This change silences the message: `There was an error processing the request. Retrying in <n> seconds.`.

Dave sent [this stackoverflow answer](https://stackoverflow.com/questions/22822267/how-to-capture-print-output-of-another-module) which allows us to capture the output of a module and send it to `stdout`. I modified the answer slightly to capture from `stderr` instead, since this is where Smarty Streets prints this message. I was worried there could be the potential that we would miss some legitamate errors with this method, so I send any printed output from Smarty Streets to logging. If the message starts with the same pattern as our unwanted error, it is send to `logging.info`, whereas all other `stderr` outputs are surfaced to `logging.error`. This way, we should still get all other Smarty Streets alerts besides this unwanted one.

Because I didn't want to incur credit charges (and because this specific message seems like it would be really difficult to reproduce locally), I tested this by adding print statements to the location where we query the cache for addresses. These messages printed to `stderr` just like Smarty Streets would and say the same thing. I also confirmed other printed output gets surfaced to `logging.error`. Once I was confident that it captured the unwanted error, I wrapped both calls to the Smarty Streets API in the geocoding file with the `stderr` redirector. Lmk if there are ideas for additional/more robust testing of the change.

### DIFF
--- a/lib/id3c/cli/command/geocode.py
+++ b/lib/id3c/cli/command/geocode.py
@@ -22,6 +22,8 @@ import pandas as pd
 import sys
 import json
 import yaml
+import io
+import contextlib
 from os import environ, chdir
 from os.path import dirname
 from textwrap import dedent
@@ -275,7 +277,7 @@ def geocode_address(address: dict) -> dict:
     returns a dict containing a canonicalized address and lat/long coordinates
     from SmartyStreet's US Street geocoding API.
     """
-    
+
     # if not running in production, then prompt for SmartyStreets lookup
     # to prevent unintentional use of credits during local development and testing
     global GEOCODE_IN_NON_PROD
@@ -293,9 +295,9 @@ def geocode_address(address: dict) -> dict:
                 continue
             else:
                 break
-        
+
         GEOCODE_IN_NON_PROD = geocode_addresses
-        
+
         if geocode_addresses in ['yes', 'y']:
             LOG.debug("Proceeding with geocoding. Smartystreet credits will be used.")
             pass
@@ -305,7 +307,7 @@ def geocode_address(address: dict) -> dict:
         elif geocode_addresses == 'none':
             LOG.debug("Skipping geocoding.")
             return None
-        else: 
+        else:
             LOG.debug("Skipping geocoding.")
             return None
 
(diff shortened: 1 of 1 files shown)


davereinhart:
> Was thinking about this more over the weekend, do logging errors get surfaced as alerts in the alerts channel, or is it only stderr output?

Yes, warning- and error-level logging in Python does surface in the alerts channel. In `backoffice/crontabs` files you will each has `LOG_LEVEL` environment variable set to `warning`.

bencap:
[Silence smarty streets retry output](https://trello.com/c/yyp1x4U7/1272-silence-smarty-streets-retry-output)

lib/id3c/cli/command/geocode.py
```diff
@@ -320,7 +322,20 @@ def geocode_address(address: dict) -> dict:
...
+    smarty_stderr = io.StringIO()
+    with contextlib.redirect_stderr(smarty_stderr):
+        STREET_CLIENT.send_lookup(lookup)
+
+    for smarty_err in smarty_stderr.getvalue().splitlines():
+        if smarty_err.startswith('There was an error processing the request. Retrying in'):
+            LOG.info(smarty_err)
+        else:
+            LOG.error(smarty_err)
+    smarty_stderr.close()
```
kschwabefry, line 337: Nonblocker, but this rerouting of stderr could be nice to put into a short function which could be reused below as well.

kschwabefry (approved):
Looks good Ben! 
Per your commit comment: 

> Lmk if there are ideas for additional/more robust testing of the change.

Your testing process looks good for development. It could be helpful to run it once as well with a real smarty street API call. We probably can't trigger the error we're seeing sporadically in prod, but with a change here, its worth confirming in the general case that this still works as expected.

One other nice-to-have would be to update the commit message to include a reference to 
https://github.com/smartystreets/smartystreets-python-sdk/blob/master/smartystreets_python_sdk/retry_sender.py#L29 to show where this output is coming from upstream.

lib/id3c/cli/command/geocode.py
bencap: Ah yeah for sure I will do that, I was looking at it wanting to put some of the code duplication in a function but getting pause from the two different smarty streets calls. Not sure why I didn't just think to put some of the duplication in a function...

bencap:
>  We probably can't trigger the error we're seeing sporadically in prod, but with a change here, its worth confirming in the general case that this still works as expected.

@kschwabefry While testing I accidentally supplied incorrect credentials, which actually does trigger the error we are trying to silence. When it triggered it was actually printing, which eventually made me realize that the error messages are sent to `stdout` and not `stderr`, despite what the upstream print statement indicates. I fixed the requisite redirects and added the method to surface messages per your comment.

kschwabefry:
Thanks for letting me know about testing! Glad that helped find the stdout/stderr issue. Changes look good Ben.

bencap:
Deployed.

PR #287, opened 2022-03-02
Update pandas to 1.4.1

seidelma:
Update pandas to 1.4.1. 

This also requires some changes in the specimen manifest parsing code, because pandas has changed the way it handles missing values in strings. I almost suspect this is a bug in pandas, because of the way it breaks things. Consider the following code:

```
import pandas as pd

print(pd.__version__)

d = {'c1': ['foo', 'bar', None]}
df = pd.DataFrame(data=d, dtype='string')
print(df.dtypes)

print(df.apply(lambda c: c.str.strip().replace({pd.NA: None}).replace({"": None, "na": None})))
```
This mirrors the way the specimen manifest-parsing code does a replacement of missing values around [line 334](https://github.com/seattleflu/id3c/compare/seidelma/bugfix/pandas14?expand=1#diff-4f914815f7f16734ccfb03bedcc1104bba4dfe3cef528b2e28022fd4109ad1f5L334).

With Pandas 1.1.5, that code yields:
```
1.1.5
c1    string
dtype: object
     c1
0   foo
1   bar
2  None
```

With Pandas 1.4.1:
```
1.4.1
c1    string
dtype: object
     c1
0   foo
1   bar
2  <NA>
```

This is kind of an expedient (read: 'ugly') hack and I hate it, so if anybody more familiar with Pandas has a better way of doing this I'm all ears.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,10 +18,10 @@
     "default": {
         "aiobotocore": {
             "hashes": [
-                "sha256:5fd4d7aefa0896fe4dd32815ead8a52ed5ccb8016c7c5743fe8ff71c3c074120"
+                "sha256:dbe9ab997851c2458b07a85f682be2ccf35d679d5de0f8f571229f740ad7ba71"
             ],
             "markers": "python_version >= '3.6'",
-            "version": "==2.1.0"
+            "version": "==2.1.1"
         },
         "aiohttp": {
             "hashes": [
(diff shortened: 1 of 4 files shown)


seidelma:
I've been looking for a root cause (and will keep searching through the Pandas code) but I wanted to get this PR out there so we can move forward with Pandas 1.4 on our end.

davereinhart:
@seidelma I came across this method to replace the pd.NA values:

`df = df.astype(object).where(pd.notnull(df), None)`

After that, the dictionary generated with `to_dict` has `None` values.

seidelma:
Thanks Dave! That looks like it'll do the job for both Pandas 1.4 and 1.1.5.

seidelma:
Per discussion in standup I'm going to leave the openpyxl-warning-silencing code as-is, since it's a spurious error message and anybody who runs the `id3c.cli.io.pandas.read_excel()` function on an affected Excel file is going to encounter it.

PR #288, opened 2022-03-07
Update pandas to 1.4.1

seidelma:


seidelma:
Tested this thing against all the files specified in `season-3-AWS-S3.yaml` instead of just one of them. It no longer throws errors when it encounters `<NA>` values in the `rack_[ab]*` fields during the parse.

PR #289, opened 2022-03-09
Add aliquot C and rack C fields to sample JSON schema

davereinhart:
To be able to accept aliquot and rack C values in sample POST requests,
the JSON schema is being updated to include the `aliquot_c`, `rack_c` and
`rack_c_nickname` attributes. These will then be included in arrays with
the corresponding A and B values to be stored in `sample.details`.

### DIFF
--- a/lib/id3c/api/schemas.py
+++ b/lib/id3c/api/schemas.py
@@ -55,6 +55,9 @@ POST_SAMPLE_SCHEMA = {
         "aliquot_b": {
             "type": "string"
         },
+        "aliquot_c": {
+            "type": "string"
+        },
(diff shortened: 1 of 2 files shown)


kschwabefry (approved):
Great how easily extensible this is.
Can you please add the new attributes into the docs as well in `lib/id3c/api/static/index.html`

davereinhart:
> Great how easily extensible this is. Can you please add the new attributes into the docs as well in `lib/id3c/api/static/index.html`

Even easier than I thought it would be! Good catch on the docs page, now updated.

davereinhart:
Deployed.

PR #291, opened 2022-03-15
Upsert encounter improvements

davereinhart:
Attempting to address two separate but related issues with this PR. 

The first issue addressed are the large gaps in the `encounter_id` sequence, due to the the use of `insert ... on conflict update ...` syntax. In the cases of an `update` (which most of our transactions are) the sequence is still incremented. To avoid this issue, the `upsert_encounter` function will now use a `select` followed by `insert` or `update` based on whether the record exists or not. This matches the approach used in `upsert_sample`.

The second issue being addressed is the bloat caused by updating records unnecessarily. Currently, regardless of whether a record's data is changing or not, the update query runs and results in dead tuples which negatively impact performance. To prevent updating of records with unchanged data, a comparison of a hash of the old and new values is being added. If the values are identical, the `update` query will be skipped.


davereinhart:
To test this branch, you'll want to have unprocessed records in `receiving.fhir` of your local db, then run the FHIR ETL. To verify that changed records are being updated, you can manually edit `fhir.document` json to change any value that gets included in the encounter record, then clear the `processing_log` for that record and run the FHIR ETL again.

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -159,6 +160,14 @@ def upsert_encounter(db: DatabaseSession,
 
         LOG.info(f"Updating existing encounter {encounter.id}")
 
+        new_row_hash = db.fetch_row("""
+            select md5(row ( %(individual_id)s::integer, %(site_id)s::integer, %(encountered)s::timestamp with time zone, %(age)s::interval, %(details)s::jsonb)::text)
+        """, data)
```
kschwabefry, line 165: Is the casting here necessary here but not in the initial select because of a difference in how execute and fetch_row are interacting with the data types?

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -123,7 +123,8 @@ def upsert_encounter(db: DatabaseSession,
     with db.cursor() as cursor:
         cursor.execute("""
             select encounter_id as id,
-                identifier
+                identifier,
+                md5(row (encounter.individual_id, encounter.site_id, encounter.encountered, encounter.age, encounter.details)::text) as row_hash
```
kschwabefry, line 127: I like this hashing approach overall, and agree it should reduce updates in the table. If we added more columns to the table, I imagine we'd have to update here as well? 

I'm assuming you didn't include all columns in the hash because you wanted to exclude created/modified columns. I wonder how often we'd get "updates" due to a difference in just those columns if all columns were included in the hash. From a maintenance perspective, it might be preferable to include all the columns, but I'm not sure how much the tradeoff would end up being.

lib/id3c/cli/command/etl/__init__.py
davereinhart: The casting here is to ensure that the incoming values are the same data type and format as the stored values prior to hashing. For example the incoming `encountered` value may be "2022-03-16" but its stored as "2022-03-16 00:00:00-07"

lib/id3c/cli/command/etl/__init__.py
davereinhart: Right, I only included the columns that the `upsert_encounter` function it attempting to set. The created and modified dates don't need to be considered when determining if the data has changed. If we add more columns that would be set via `upsert_encounter` then yes, they would need to be added to the hash.

kschwabefry: Understood about created/modified not *needing* to be considered. I do still have some concerns that this creates non-obvious implications for schema changes down the line and makes the code more fragile. If this is currently the only place programatically modifying encounters, there should only be nominal functional difference in the amount of updates if it includes all columns. 
As an alternative to including all columns in the hash, we could explicilty ignore the created/modified columns with an approach like https://blog.jooq.org/selecting-all-columns-except-one-in-postgresql/

davereinhart: I see what you're getting at, but any schema changes would have already required multiple changes to `upsert_encounter`, including the parameters and `data` dict. Agree that updating the hashed columns would be less obvious, but the benefits to hardcoding them this way is that the incoming `data` values can be cast to match how they would appear if they were already in the table, so when hashed they will match the value coming from the `select` query. The hash values are different without that casting.

If you see a way that can work, let me know. Otherwise I can add some comments to the code explaining the need to update those lines if there are schema changes.

kschwabefry: Ok! Thanks for the discussion here Dave. I think adding a comment would be helpful. I think thats all of the feedback I have, so I'll go ahead and mark approved from my end.

kschwabefry:
I did some local testing with the fhir etl and looks good on my end. confirmed: 

- new encounter got created
- existing but unmodified encounter got skipped
- existing but modified encounter got updated

davereinhart:
Deployed.

davereinhart:

PR #292, opened 2022-03-18
Ignore FHIR resource reference IDs when determining if data changed

davereinhart:
The current ETL processes for creating FHIR bundles include the generation
of random UUIDs which are used as reference identifiers within each bundle.
Since these values are not indicitive of the actual data changing, they
are being removed before comparing the existing and incoming data.

Also being removed is the md5 hashing, which was not necessary for the comparison.

### DIFF
--- a/lib/id3c/cli/command/etl/__init__.py
+++ b/lib/id3c/cli/command/etl/__init__.py
@@ -122,29 +122,32 @@ def upsert_encounter(db: DatabaseSession,
 
     # Select on identifier to determine if the encounter record already exists.
     #
-    # This query also includes a hash comparison to determine if the data has changed and
-    # an update is necessary. If a record is found with the given identifier, a hash is
-    # calculated on the columns that would be set on update, and compared to a hash calculated
-    # on the incoming values.
+    # This query also compares the existing vs. incoming row to determine if an update
+    # is necessary. If a record is found with the given identifier, the columns that would be
+    # set on update are compared to the incoming values, ignoring any values in details starting
+    # with `urn:uuid:` (FHIR resource reference IDs that are randomly generated each time a FHIR
+    # bundle is constructed, and are not indicitive of data having changed).
     #
-    # Note: if there are schema changes to `warehouse.encounter` and this upsert function, the
-    # list of columns and data types included in the hash comparison should be updated to include
-    # all columns set on update.
+    # Note: if there are schema changes to `warehouse.encounter` resulting in changes to the columns
+    # that are set on update, the list of columns being compared should be updated to match.
+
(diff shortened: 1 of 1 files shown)


davereinhart:

bencap (approved):
This is working as intended when testing locally and skipping the record when only the UUIDs are different, I just want to make sure I have an understanding of the original rationale for these UUIDs. These fhir records are created and all the UUIDs are generated randomly every time, regardless of whether the data is actually new or not. Then when they are upserted into the database, the UUIDs are used to link different parts of the record together (like the `encounter` aspects), correct? So basically if that underlying data isn't changing, there is no reason why we should be updating the UUIDs as well - which is what this is changing.

davereinhart:
> This is working as intended when testing locally and skipping the record when only the UUIDs are different, I just want to make sure I have an understanding of the original rationale for these UUIDs. These fhir records are created and all the UUIDs are generated randomly every time, regardless of whether the data is actually new or not. Then when they are upserted into the database, the UUIDs are used to link different parts of the record together (like the `encounter` aspects), correct? So basically if that underlying data isn't changing, there is no reason why we should be updating the UUIDs as well - which is what this is changing.

That's exactly right. When we reprocess records (which we do in high volume for HCT) the FHIR bundles get regenerated in `receiving.fhir` for old encounters, and re-ingested into `warehouse.encounter` with no changes other than these FHIR reference IDs, but it was still seeing this as a data change.

Most of them are used as `fullUrl` or 'reference' ids:
https://www.hl7.org/fhir/bundle-definitions.html#Bundle.entry.fullUrl

It feels like a bit of a workaround to handle it this way, but far less code to strip them out here instead of trying to generate them consistently, and there's very little overhead.

kschwabefry (approved):
At first I was hoping we might be able to do this with jsonb operators (https://www.postgresql.org/docs/14/functions-json.html), but because of the way the keys are nested, it seems that would be a lot more convoluted. The regex is a good solution for these purposes. In the longer term, this seems like another argument for splitting things out from the details column into a more structured table.

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -125,7 +125,9 @@ def upsert_encounter(db: DatabaseSession,
     # This query also includes a hash comparison to determine if the data has changed and
     # an update is necessary. If a record is found with the given identifier, a hash is
     # calculated on the columns that would be set on update, and compared to a hash calculated
```
kschwabefry, line 127: Nit - update comment to reflect dropping the hashing

davereinhart:
Deployed.

PR #293, opened 2022-03-18
Upsert `encounter_location` Improvements

bencap:
Similar to what was just fixed by Dave in the `warehouse.encounter` upsert function (see #291), this change compares old values in the `warehouse.encounter_location` table with new incoming values from the `receiving.fhir` table. If the values are the same, then this update is skipped, thereby reducing bloat in this table.


davereinhart:

davereinhart (changes requested):

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -378,17 +378,86 @@ def upsert_encounter_location(db: DatabaseSession,
...
+                row(
+                   encounter_location.encounter_id,
+                   encounter_location.relation,
+                   encounter_location.location_id
+                ) !=
+                row(
+                   %(encounter_id)s::integer,
+                   %(relation)s::citext,
+                   %(location_id)s::integer
+                ) as data_changed
```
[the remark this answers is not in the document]
bencap: Ah of course, the where statement takes care of the other too, thanks!

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -378,17 +378,76 @@ def upsert_encounter_location(db: DatabaseSession,
...
+        )
+
+    else:
+        raise Exception(f"More than one encounter matching. encounter_id: {encounter_id}, relation: {relation}")
+
+    if encounter_loc:
+        LOG.info(f"Upserted encounter with location '{relation}'")
 
-        assert cursor.rowcount == 1, "Upsert affected no rows!"
+    return encounter_loc
```
davereinhart, line 450: To keep the behavior consistent, this function doesn't need to return anything, but we should try to keep the assert check in place, but have it check `len(encounter_loc)==1` instead.

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -378,17 +378,76 @@ def upsert_encounter_location(db: DatabaseSession,
...
+        )
+
+    elif len(encounter_locs) == 1:
+        encounter_loc = encounter_locs[0]
+
+        LOG.info("Updating existing encounter location")
+
+        if encounter_loc.data_changed == False:
+            LOG.info(f"Skipping upsert for encounter with location '{relation}' (no change).")
+            return encounter_loc
```
davereinhart, line 431: Similar to the comment below, no need to return anything if we're matching the prior implementation.

davereinhart (changes requested):

lib/id3c/cli/command/etl/__init__.py
bencap: For sure on the matching prior implementation. Do you know why this upsert in particular has no return value? Seems like all the other upserts in this file do return the update they made.

davereinhart: Yeah, I noticed that too. My guess is that the returned value isn't needed anywhere in the process, just confirmation that it was upserted.

davereinhart (approved):

bencap:
Deployed

PR #294, opened 2022-03-23
Cast rows to text for data comparison in upsert_encounter

davereinhart:
When comparing two items with data type `row`, the presence of a null element
within either row will return null rather than the expected boolean. This was
resulting in old encounter records continuing to be updated even though the data was not changing.

To avoid this behavior, the rows are being cast to text before comparing them.


bencap (commented):
Working as intended for me, records on master with null age and null incoming age are getting updated while records on this branch are not.

bencap (approved):

davereinhart:
Deployed.

davereinhart:

PR #297, opened 2022-04-04
Prevent Updates to Unchanged Location Records

bencap:
This change compares incoming values with existing values in the `warehouse.location` table
to decide if there is a need to perform an update. Unnecessary updates add bloat to the
table, which this change will cut down on.

### DIFF
--- a/lib/id3c/cli/command/etl/__init__.py
+++ b/lib/id3c/cli/command/etl/__init__.py
@@ -357,19 +357,70 @@ def upsert_location(db: DatabaseSession,
     else:
         hierarchy = hierarchy + "," + location_hierarchy
 
-    location = db.fetch_row("""
-        insert into warehouse.location (scale, identifier, hierarchy)
-        values (%s, %s, %s)
+    data = {
+        "scale": scale,
+        "identifier": identifier,
+        "hierarchy": hierarchy
+    }
 
-        on conflict (scale, identifier) do update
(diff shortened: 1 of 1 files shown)


bencap:

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -357,19 +357,70 @@ def upsert_location(db: DatabaseSession,
...
+    }
+
+    with db.cursor() as cursor:
+        cursor.execute("""
+            select
+                location_id as id,
+                scale,
+                identifier,
+                hierarchy,
+                (location.hierarchy) != (%(hierarchy)s) as data_changed
```
davereinhart, line 373: We'll want to cast the incoming value to `hstore` data type for the comparison. The extra parenthesis aren't necessary.

davereinhart (changes requested):

davereinhart (commented):
~~@bencap I noticed some interesting behavior when testing with a FHIR bundle containing multiple encounters. An update on the location record (when data has changed) happens for every encounter. I suspect that's because the changes aren't committed until the entire bundle is processed, but I haven't confirmed that yet.~~ Disregard, appears be to specific to the process I was using to test it, subsequent tests look good.

davereinhart (approved):

bencap:
Deployed.

PR #298, opened 2022-04-13
Log changes to encounter_id when updating sample records

davereinhart:
Changes to the encounter_id linked to each sample may indicate an issue with the source
data, ETL, or the result of corrections being made on the source. This results in encounter
records that had previously been linked to samples to no longer be, which could indicate they
are no longer valid and can be deleted.

Adding logging at debug level to monitor how often this occurs during normal operations.
If it's reasonably low, this may be elevated to warning level. If it's occurring more
often than expected, it will require some additional investigation.


bencap (approved):

davereinhart:
Deployed.

PR #299, opened 2022-04-21
Use `cache.keys()` to cache addresses rather than Try/Catch

bencap:
Using try/catch will surface a key error if an address is not in the cache.
Normally this isn't a problem, but if another exception occurs while in
the `catch` segement of an address not being in the cache the key will get
logged out. Since this key is an address, there is therefore the potential
that this address is logged on geocoding failure. This change will prevent
the actual address from being logged if there is a geocoding failure.


kschwabefry (approved):
Kudos on the commit message too, you did a great job of capturing why this change is being made. 👍

bencap:
Deployed.

PR #301, opened 2022-05-16
Fix misnamed key in `upsert_location`

bencap:
When I deployed this previously I deployed with a misnamed key in one of the SQL commands.


bencap:
Deployed

PR #302, opened 2022-05-17
Ingest Inconclusive P/A Results as Null

bencap:
Previously, `None` type presence absence results encoded within FHIR bundles were dropped by this pipeline. This change ingests any such results with `presence = null` so that inconclusive results may be ingested as part of the PHSKC ETL.

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -892,7 +892,7 @@ def process_presence_absence_tests(db: DatabaseSession, report: DiagnosticReport
         Return the boolean value of a presence/absence result observation.
 
         Expects the observation value to be within valueBoolean or
-        valueCodeableConcept. Raises Exception if both are None.
+        valueCodeableConcept (or None in the case of valueBoolean).
 
         Also raises Exception if valueCodeableConcept contains an unknown code.
         """
@@ -912,7 +912,9 @@ def process_presence_absence_tests(db: DatabaseSession, report: DiagnosticReport
 
             return code_map[code]
 
-        raise Exception("Could not find presence/absence observation value in valueBoolean or valueCodeableConcept")
+        elif observation.valueBoolean is None:
+            return observation.valueBoolean
+
 
     if not report.result:
         raise Exception("An empty value for `result` violates the FHIR docs.")
@@ -923,7 +925,7 @@ def process_presence_absence_tests(db: DatabaseSession, report: DiagnosticReport
         snomed_code = matching_system_code(observation.code, SNOMED_SYSTEM)
         assert snomed_code, "No SNOMED code found"
 
-        # Skip results that are Inconclusive
+        # Skip SNOMED based results that are Inconclusive
         if snomed_code == '911000124104':
             continue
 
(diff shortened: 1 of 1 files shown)


bencap:
[For PHSKC samples, add result_value to the columns we ingest.](https://trello.com/c/yj5B2B32/1371-for-phskc-samples-add-resultvalue-to-the-columns-we-ingest)

davereinhart (commented):
I noticed that there are two older pipelines that might be impacted by this change. Other than UW-Retrospectives ETL (which is already dropping inconclusive results as you noted) we have presence/absence records ingested via FHIR for the Kiosk project (16691) and through the FHIR API endpoint for Ellume test kits (https://github.com/seattleflu/documentation/blob/32ddcb31bac9e4725701a775e5daa43c3ca93c4f/Data-flow.md#ellume)

Both of these projects are over, but it's worth checking to see if these changes would have any impact if we ever needed to reprocess that data or received any new data through the FHIR API endpoint.

bencap:
@davereinhart I had looked at the kiosk one and it looks like we are only setting the `valueBoolean` field as `True` or `False`, so altering the logic to also ingest None type results shouldn't have an effect (if anything it is a bit odd that inconclusive results are ingested as `False` in that pipeline https://github.com/seattleflu/id3c-customizations/blob/8b1763133a9ffa966ce3103cd58d2a4bd748939a/lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py#L290-L309). 

For the Ellume kits I'm not sure of the possible values provided with those FHIR bundles. If they post them directly to the endpoint I can try looking back through the receiving table for what they've sent in the past.

davereinhart:
> @davereinhart I had looked at the kiosk one and it looks like we are only setting the `valueBoolean` field as `True` or `False`, so altering the logic to also ingest None type results shouldn't have an effect (if anything it is a bit odd that inconclusive results are ingested as `False` in that pipeline https://github.com/seattleflu/id3c-customizations/blob/8b1763133a9ffa966ce3103cd58d2a4bd748939a/lib/seattleflu/id3c/cli/command/etl/redcap_det_kiosk.py#L290-L309).
> 
> For the Ellume kits I'm not sure of the possible values provided with those FHIR bundles. If they post them directly to the endpoint I can try looking back through the receiving table for what they've sent in the past.

That logic in the Kiosk ETL does seem odd. It's marking inconclusive as `True` though, isn't it?

davereinhart:
@bencap Looks like there are at least some "Indeterminate" results in the FHIR bundles from Ellume. (examples: receiving.fhir records 5640, 5687, 10756). I haven't tracked them down in warehouse to see how they were ingested yet. We could re-process those FHIR bundles to see if they're handled any differently with your changes.

bencap:
@davereinhart Thanks for taking a look I'll look into those records and run them through the appropriate pipeline again to observe the behavior. What was the query you used to find those out of curiosity?

davereinhart:
> @davereinhart Thanks for taking a look I'll look into those records and run them through the appropriate pipeline again to observe the behavior. What was the query you used to find those out of curiosity?

I had to look through a larger set to see what value they were using, but this query will pull them up for you (added a limit because it's a huge table and we only need a few examples):
```
select fhir_id, document -> 'entry' -> 0 -> 'resource' -> 'contained' as result
from receiving.fhir 
where document::text like '%Ellume%'
and document::text like '%Indeterminate%'
limit 3
```

davereinhart (approved):

bencap:
Deployed.

PR #303, opened 2022-05-20
Add index to identifier.barcode column

davereinhart:
To improve performace of sort, filters, and joins that use identifier.barcode column, an
index is being added. One of the drivers for this change is to address recurring issue of
`shipping.uw_reopening_ehs_reporting_v1` view hanging indefinitely, which includes a join
with `shipping.return_results_v3` on the barcode column.


davereinhart:
After further review, this doesn't seem necessary since there is a unique constraint resulting in an index being added by postgres (https://www.postgresql.org/docs/13/indexes-unique.html).

PR #304, opened 2022-05-23
Rework upsert individual function

davereinhart:
To prevent gaps in the `individual_id` sequence and unnecessary updates to records
in `warehouse.individual` when there are no changes, the `upsert_individual` function
is being updated to first check if a record exists and then only perform the update
if the relevant columns have different data. Preventing unnecessary updates will help
reduce bloat and provide a more accurate modified date on the record.

### DIFF
--- a/lib/id3c/cli/command/etl/__init__.py
+++ b/lib/id3c/cli/command/etl/__init__.py
@@ -78,23 +78,64 @@ def upsert_individual(db: DatabaseSession, identifier: str, sex: str = None, det
     data = {
         "identifier": identifier,
         "sex": sex,
-        "details": Json(details),
+        "details": Json(details) if details else None,
     }
 
-    individual = db.fetch_row("""
-        insert into warehouse.individual (identifier, sex, details)
-            values (%(identifier)s, %(sex)s, %(details)s)
+    with db.cursor() as cursor:
+        cursor.execute("""
+            select individual_id as id,
+                identifier,
+                (
+                    individual.sex != %(sex)s
+                    or
+                    (coalesce(individual.details,'{}'::jsonb) != (coalesce(individual.details, '{}'::jsonb) || coalesce(%(details)s, '{}')::jsonb))
+                ) as data_changed
+            from warehouse.individual
+            where identifier = %(identifier)s
+            for update
+            """, data)
 
-        on conflict (identifier) do update
-            set sex     = excluded.sex,
-                details = coalesce(individual.details, '{}'::jsonb) || excluded.details
+        individuals = list(cursor)
 
+    if not individuals:
+        LOG.info("Creating new individual")
+
+        individual = db.fetch_row("""
+            insert into warehouse.individual (
+                    identifier,
+                    sex,
+                    details)
+                values (
+                    %(identifier)s,
+                    %(sex)s,
+                    %(details)s)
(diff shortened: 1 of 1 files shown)


davereinhart:
While testing this function, I discovered that a previous update of this function resulted in some extraneous null values in `individual.details` records coming. That issue is resolved with this update, but will require cleaning up those records on prod after deployment:
```
--update warehouse.individual set details = null
select * from warehouse.individual
where jsonb_strip_nulls(individual.details->0) = '{}'::jsonb
```

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -78,23 +78,64 @@ def upsert_individual(db: DatabaseSession, identifier: str, sex: str = None, det
...
+                returning individual_id as id, identifier
+        """, { **data, "individual_id": individual.id })
+
+    # More than one found → error
+    else:
+        raise Exception(f"More than one individual found matching identifier: {identifier}")
 
-    LOG.info(f"Upserted individual {individual.id} «{individual.identifier}»")
+    if individual:
+        LOG.info(f"Upserted encounter {individual.id} «{identifier}»")
```
bencap, line 138:
```suggestion
        LOG.info(f"Upserted individual {individual.id} «{identifier}»")
```

lib/id3c/cli/command/etl/__init__.py
```diff
@@ -78,23 +78,64 @@ def upsert_individual(db: DatabaseSession, identifier: str, sex: str = None, det
...
-        insert into warehouse.individual (identifier, sex, details)
-            values (%(identifier)s, %(sex)s, %(details)s)
+    with db.cursor() as cursor:
+        cursor.execute("""
+            select individual_id as id,
+                identifier,
+                (
+                    individual.sex != %(sex)s
+                    or
+                    (coalesce(individual.details,'{}'::jsonb) != (coalesce(individual.details, '{}'::jsonb) || coalesce(%(details)s, '{}'::jsonb)))
```
bencap, line 91: Curious about this line, in what circumstances would `(coalesce(individual.details,'{}'::jsonb) != (coalesce(individual.details, '{}'::jsonb)` these two things not be equal to each other? Does it have something to do with the null values you noted in your comment above?

bencap (commented):
Looked good to me based on local testing, just a minor typo in one of the log statements.

lib/id3c/cli/command/etl/__init__.py
davereinhart: The two values being compared there are:

`coalesce(individual.details,'{}'::jsonb)` 
and
`(coalesce(individual.details, '{}'::jsonb) || coalesce(%(details)s, '{}'::jsonb))`

The latter is the merged value that would be used if an update is performed. And yes, the coalesces are there to make sure it handles nulls correctly. The current version is resulting in jsonb arrays where a null is appended each time there's an update but this change should prevent that.

bencap: Ah got it thanks, I was misreading the statement.

bencap (approved):
Looked good to me based on local testing, just a minor typo in one of the log statements.

kschwabefry (approved):

davereinhart:
Deployed.

PR #305, opened 2022-05-24
Rework upsert presence_absence function

davereinhart:
The `upsert_presence_absence` function is being updated in line with the other
upsert functions, to prevent gaps in the sequence used for `presence_absence_id` and
avoid unneccessary updates when data has not changed.

Also adding warning-level logs for non-typical changes to `presence_absence` records which
may require further investigation (changing `present`, `sample_id`, or `target_id` values).

### DIFF
--- a/lib/id3c/cli/command/etl/__init__.py
+++ b/lib/id3c/cli/command/etl/__init__.py
@@ -564,36 +564,76 @@ def upsert_presence_absence(db: DatabaseSession,
         "sample_id": sample_id,
         "target_id": target_id,
         "present": present,
-        "details": Json(details)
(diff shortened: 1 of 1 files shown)


davereinhart:
One thing I noticed while reworking this function is that it was merging details on updates. I kept that behavior the same for now, but it seems like we may want it to overwrite rather than merge there.

kschwabefry:
[QC alerts for sample updates](https://trello.com/c/c7XqM7Vv/1370-qc-alerts-for-sample-updates)

davereinhart:

davereinhart:
Deployed.

PR #306, opened 2022-05-25
Warn when sample identifiers are changed during upsert

kschwabefry:
upsert_sample can allow identifiers to change for a given sample.
It should be rare that a sample identifier or collection identifier
is changed from one non-null value to another. When this happens,
log a warning so that we can confirm with the data source (typically,
the lab) that this change was intended. And if so, we can do any
manual cleanup as necessary for participants results or linelist
reporting, for example.

### DIFF
--- a/lib/id3c/db/__init__.py
+++ b/lib/id3c/db/__init__.py
@@ -251,9 +251,15 @@ def upsert_sample(db: DatabaseSession,
 
         LOG.info(f"Updating existing sample {sample.id}")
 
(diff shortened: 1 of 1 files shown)


kschwabefry:
[QC alerts for sample updates](https://trello.com/c/c7XqM7Vv/1370-qc-alerts-for-sample-updates)

kschwabefry:

davereinhart (approved):

kschwabefry:
Deployed.

PR #308, opened 2022-05-31
Handle nanoseconds in timestamp values

davereinhart:
The presence-absence ETL failed due to an incorrectly formatted timestamp
which had 9 decimal places instead of the 6 that it was expecting. Using
parser from dateutil handles varying number of decimal places on the seconds
value.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,10 +18,10 @@
     "default": {
         "aiobotocore": {
             "hashes": [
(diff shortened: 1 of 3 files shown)


kschwabefry (approved):
Code lgtm, thanks for the quick fix.

I wonder if this was a change to how samplify stores the timestamps internally or just how its transmitted to us. If the former, that could impact us seeing re-transferred old data as changed if we got more specific timestamps for the same presence absence data.

davereinhart:
Deployed.

PR #309, opened 2022-06-15
Add API Endpoint for `/receiving/manifest`

bencap:
This change adds an API endpoint for resources intended to be inserted
into the `receiving.manifest` table. Inserted resources can then be run
through the existing manifest ETL routine to create/update samples.


bencap:
[Build API for incident report ingest](https://trello.com/c/xBUGE1O3/1249-build-api-for-incident-report-ingest)

lib/id3c/api/routes.py
```diff
@@ -149,14 +150,56 @@ def receive_fhir(*, session):
...
+    Receive JSON representation of manifest resources.
+    """
+    document = request.get_data(as_text = True)
+
+    LOG.debug(f"Received Manifest Document")
+
+    add_provenance(document, request)
+    datastore.store_manifest(session, document)
+
+    return "", 204
```
davereinhart, line 169: If there's no immediate use case for this endpoint, may be better to leave it out.

lib/id3c/api/schemas.py
```diff
@@ -100,3 +100,45 @@
...
+            "type": "string",
+            "enum": ["discarded", "continued processing"]
+        }
+    },
+    "anyOf": [
+        { "required":
+            [ "sample_id" ] },
+        { "required":
+            [ "collection_id" ] }
+    ],
```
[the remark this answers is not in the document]
bencap: Yeah I think you are right about this. I only see the `collection_id` filled in on both the LIMS incident report and the google sheet one.

lib/id3c/api/schemas.py
```diff
@@ -100,3 +100,43 @@
...
+        },
+        "corrective_action": {
+            "type": "string",
+            "enum": ["discarded", "continued processing"]
+        }
+    },
+    "allOf": [
+        { "required":
+            [ "collection_id" ] }
+    ],
```
davereinhart, line 140: Not a blocker, but I think `"required": ["collection_id"],` will accomplish the same thing.

lib/id3c/api/schemas.py
```diff
@@ -100,3 +100,43 @@
...
+        "sample_id": {
+            "type": "string",
+            "minLength": 8,
+            "maxLength": 8
+        },
+        "collection_id": {
+            "type": "string",
+            "minLength": 8,
+            "maxLength": 8
+        },
```
davereinhart, line 116: To be processed by the manifest ETL, these keys should be `sample` and `collection`.

davereinhart (changes requested):

davereinhart (commented):
One detail about the manifest ETL that may be important here, is that if `collection` and `sample` barcodes are both present, it can update `identifier` or `collection_identifier` in the sample record. This was the desired behavior for AQ sheets, but probably not what we want for incident reports, so it may be best to only have `collection` barcode, or only allow `collection` _or_ `sample`, but not both.
https://github.com/seattleflu/id3c/blob/673a83107da95477c166a6a350e7b19994d998db/lib/id3c/cli/command/etl/manifest.py#L118-L122

bencap:
@davereinhart Sure that makes sense. Given what I have seen from the incident reports (always have collection barcode but rarely have sample barcode) I would say working only with collection barcode would make sense. We could have either/or though just to keep the option for later.

davereinhart (approved):

davereinhart:
[Coordinate with Jeremy and Melissa to test Incident Endpoint.](https://trello.com/c/63AGkmaO/1396-coordinate-with-jeremy-and-melissa-to-test-incident-endpoint)

bencap:
Deployed.

PR #310, opened 2022-07-12
Prevent Updates to Unchanged Sample Records

bencap:
This change checks for identifier and metadata changes in sample records prior to upserting
them into the warehouse table. If no changes are detected we can safely skip updates and
thereby make less requests to our table. Note we check identifiers separately because we
often upsert with one or the other.

This change also adds a warning if the function is called and requested to update identifiers
but is only passed 1 identifier. This would be useful if we have a sample with both identifiers
filled out and we are updating it with only one identifier passed in (this would erase the
existing identifier).

### DIFF
--- a/lib/id3c/db/__init__.py
+++ b/lib/id3c/db/__init__.py
@@ -214,15 +214,34 @@ def upsert_sample(db: DatabaseSession,
         "collection_identifier": collection_identifier,
         "collection_date": collection_date,
         "encounter_id": encounter_id,
-        "additional_details": Json(additional_details),
+        "additional_details": Json(additional_details) if additional_details else None,
     }
 
     # Look for existing sample(s)
     with db.cursor() as cursor:
         cursor.execute("""
-            select sample_id as id, identifier, collection_identifier, encounter_id
-              from warehouse.sample
-             where identifier = %(identifier)s
+            select
+                sample_id as id, identifier, collection_identifier, encounter_id, details,
+                row (
+                    identifier,
+                    collection_identifier
+                )::text !=
+                row (
+                    %(identifier)s,
+                    %(collection_identifier)s
+                )::text as identifiers_changed,
+                row(
+                    collected::timestamp,
+                    encounter_id,
+                    details
+                )::text !=
+                row(
+                    coalesce(%(collection_date)s, collected)::timestamp,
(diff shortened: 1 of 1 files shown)


bencap:

lib/id3c/db/__init__.py
```diff
@@ -214,15 +214,34 @@ def upsert_sample(db: DatabaseSession,
...
+                row(
+                    collected::timestamp,
+                    encounter_id,
+                    details
+                ) !=
+                row(
+                    %(collection_date)s::timestamp,
+                    %(encounter_id)s::integer,
+                    coalesce(details, '{}'::jsonb) || coalesce(%(additional_details)s, '{}')::jsonb
+                ) as metadata_changed
```
davereinhart, line 242: To ensure `identifer_changed` and `metadata_changed` return true or false, you can cast these rows to text before the comparison to handle null values.

lib/id3c/db/__init__.py
```diff
@@ -261,6 +287,9 @@ def upsert_sample(db: DatabaseSession,
         if collection_identifier and sample.collection_identifier and collection_identifier != sample.collection_identifier:
             LOG.warning(f"upsert_sample: collection_identifier is changing on sample {sample.id} from {sample.collection_identifier} to {collection_identifier}")
 
+        if sample.identifiers_changed is None and update_identifiers:
```
davereinhart, line 290: If the change suggested above is made, this should be able to be changed to check for `False` instead of `None`.

lib/id3c/db/__init__.py
```diff
@@ -214,15 +214,34 @@ def upsert_sample(db: DatabaseSession,
...
+                    %(collection_identifier)s
+                )::text as identifiers_changed,
+                row(
+                    collected::timestamp,
+                    encounter_id,
+                    details
+                )::text !=
+                row(
+                    %(collection_date)s::timestamp,
+                    %(encounter_id)s::integer,
```
davereinhart, line 240: Still seeing records with `metadata_changed = True`, resulting in an update, when data is not changing. These two lines should also coalesce with the existing values, to match how the record would appear after the update if it occurs.

davereinhart (approved):

bencap:
Deployed.

PR #311, opened 2022-07-15
Add `mint_identifiers` function to improve performance

davereinhart:
A `mint_identifiers` function is being added to improve performance of identifier
minting by handling more of the process in postgres and less in python. The function
will generate uuids and check the hamming distance of the last 8 characters against
all existing barcodes. It will continue trying new identifiers until the distance check
passes, at which point it will insert a record into the identifier table. The function will repeat
this process until it has reached the indicated `number_to_mint` and returns the records that
were inserted.

### DIFF
--- a/lib/id3c/db/__init__.py
+++ b/lib/id3c/db/__init__.py
@@ -41,7 +41,6 @@ def mint_identifiers(session: DatabaseSession, name: str, n: int) -> Any:
     `permission denied` error.
     """
     minted: List[Any] = []
-    failures: Dict[int, int] = {}
 
     # Lookup identifier set by name
     identifier_set = session.fetch_row("""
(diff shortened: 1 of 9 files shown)


davereinhart:
[Barcode minting performance improvements](https://trello.com/c/aY0byUxH/1395-barcode-minting-performance-improvements)

davereinhart:
To test, I stood up the RDS testing instance and deployed the sqitch changes.  Using this command to test performance vs. master branch:
`PGSERVICE=seattleflu-testing LOG_LEVEL=info pipenv run id3c identifier mint samples 100 --dry-run`

lib/id3c/db/__init__.py
```diff
@@ -56,37 +56,24 @@ def mint_identifiers(session: DatabaseSession, name: str, n: int) -> Any:
 
     started = datetime.now()
 
-    while len(minted) < n:
-        m = len(minted) + 1
+    minted = session.fetch_all("""
+        select uuid, barcode, identifier_set_id, generated
+        from mint_identifiers(%s, %s)
+        """, (identifier_set.id, n))
```
bencap, line 62: You mentioned this to me but wanted to note here too, does it make sense to wrap this inside a savepoint? Not sure if it matters or not since everything is handled in the Postgres side now but maybe it would protect from some edge cases with the DB connection that aren't immediately apparent.

lib/id3c/db/__init__.py
```diff
@@ -56,37 +56,24 @@ def mint_identifiers(session: DatabaseSession, name: str, n: int) -> Any:
 
     started = datetime.now()
```
bencap, line 57:
```suggestion

```
Don't believe this line is needed anymore with stats being calculated in postgres.

lib/id3c/db/__init__.py
davereinhart: I don't believe the savepoint is needed anymore since the iteration is being handled in PostgreSQL function, and the entire transaction will either succeed or be rolled back if it fails. The previous implementation was using them to remove inserts that failed from from the transaction, but I'm not sure that's actually necessary.

davereinhart: Confirmed we shouldn't need savepoints here, and should expect the same behavior because postgres automatically rolls back any changes since the begin statement when there's an exception: [42.13.2.1. Implicit Rollback After Exceptions](https://www.postgresql.org/docs/13/plpgsql-porting.html#PLPGSQL-PORTING-OTHER)

bencap: Thanks for looking into it @davereinhart

bencap (approved):

kschwabefry (approved):

davereinhart:
Deployed.

PR #314, opened 2022-08-01
Ignore Empty NDJSON Lines

bencap:
Pandas fixed a bug that omitted trailing new lines from NDJSON dataframe output, however this caused parsing of NDJSON's to fail on the final empty line. We were originally going to handle this while dumping the NDJSON's, however it is better to use the format as intended (see #313). This change will ignore empty lines while loading NDJSON files, allowing us to use the fixed format.

### DIFF
--- a/lib/id3c/json.py
+++ b/lib/id3c/json.py
@@ -23,7 +23,7 @@ def load_json(value):
     messaging, compared to :exc:`json.JSONDecodeError`, when stringified.
     """
     try:
-        return json.loads(value)
+        return json.loads(value) if value else None
     except json.JSONDecodeError as e:
         raise JSONDecodeError(e) from e
 
@@ -40,7 +40,7 @@ def load_ndjson(file: Iterable[str]) -> Iterable:
     """
     Load newline-delimited JSON records from *file*.
     """
-    yield from (load_json(line) for line in file)
+    yield from (load_json(line.strip()) for line in file)
 
 
 class JsonEncoder(json.JSONEncoder):


bencap:
@seattleflu/core-devs Interested to get some thoughts on this. The pandas change does add a trailing newline (so the end of the file looks like `\n\n`). Based on this spec (https://github.com/ndjson/ndjson-spec#32-parsing), it seems fine to ignore empty lines, but I'm also a bit annoyed Pandas made that change since I prefer the old behavior. One of the benefits of NDJSON to me is that each line is valid JSON, which that Pandas change breaks.

I'd be interested to know other thoughts, but I personally prefer the other implementation (which I may have closed prematurely). It doesn't seem like the spec necessitates files ending in `\n\n` and doing so hurts some of the aspects of NDJSON files that make them nice to work with (one valid record per line, easily concatenated or subset-able files, really simple loading and dumping interface).

davereinhart:
> @seattleflu/core-devs Interested to get some thoughts on this. The pandas change does add a trailing newline (so the end of the file looks like `\n\n`). Based on this spec (https://github.com/ndjson/ndjson-spec#32-parsing), it seems fine to ignore empty lines, but I'm also a bit annoyed Pandas made that change since I prefer the old behavior. One of the benefits of NDJSON to me is that each line is valid JSON, which that Pandas change breaks.
> 
> I'd be interested to know other thoughts, but I personally prefer the other implementation (which I may have closed prematurely). It doesn't seem like the spec necessitates files ending in `\n\n` and doing so hurts some of the aspects of NDJSON files that make them nice to work with (one valid record per line, easily concatenated or subset-able files, really simple loading and dumping interface).

I agree, I think it's a little cleaner to not have the extra newline at the end, but based on the documentation you linked to we should allow it and handle it during parsing (not just at the end, but anywhere in the file). It also says it should be configurable, so I'd suggest adding another argument and logic to `load_ndjson` function for `ignore_empty_lines` (with default `True`).

kschwabefry (approved):

davereinhart (approved):

bencap:
Deployed.

PR #315, opened 2022-09-15
Unpin xlrd and upgrade to latest version

bencap:
Xlrd was previously pinned to version = 1.2.0 because of security vulnerabilities that prevented it from parsing new xlsx files. Openpyxl now handles these files and xlrd handles .xls files. Pandas requires xlrd >= 2.1.0 now, which we didn't have because it could cause issues parsing certain barcodes. Now that we've transitioned away from aliquoting sheets, we only parse excel files for retrospective data.

This package upgrade was tested on SCH, PHSKC, and UW retros data. For SCH data, problematic barcodes parsed the same on 1.2.0 and 2.1.0. For PHSKC data, there was no diff between the production ndjson output on xlrd 1.2.0 and my local output on xlrd 2.1.0. Likewise, there was no diff between the prod uw retro ndjson generated with xlrd 1.2.0 and my local ndjson generated with xlrd 2.1.0.

I think that covers the jobs still using manifest excel files, but let me know if there are others and I can look into those as well.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,11 +18,11 @@
     "default": {
         "aiobotocore": {
             "hashes": [
-                "sha256:6554ebea5764f66f4be544a4fcaa0953ee80e600dd7bd818ba4893d72bf12bfb",
-                "sha256:eae059eb51726cee4de2027cfc72bfccc76cf0c229d6b2b08f640e53a568f657"
+                "sha256:6c25381d31b712652bc2f6008683949351c240c56d24b1d8ae252c1034a50f63",
+                "sha256:f9fe0698cc497861bdb54cd16161c804031f758ada9480c35540f20c0c078385"
             ],
-            "markers": "python_version >= '3.6'",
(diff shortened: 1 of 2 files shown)


davereinhart:
@bencap Thank for doing some thorough testing! I think we are still parsing AQ sheets as part of the UW retros job, so may be worth testing on those but they're all XLSX format so may not be impacted.

bencap:
@davereinhart I did test that job as well. Ran locally with 2.1.0 and diffed against what was already in the S3 receiving folder (generated from prod with 1.2.0) gave me no new records, so that one seems fine too.

davereinhart (approved):

bencap:
Deployed.

bencap:
[ID3C: Look at other library (not xlrd) to read XLSX files](https://trello.com/c/VQguko8S/946-id3c-look-at-other-library-not-xlrd-to-read-xlsx-files)

PR #318, opened 2023-02-02
Add a new `singlet` Format for Sample Barcodes

bencap:
Lab will now occasionally ask for sample barcodes in single formation rather than in duplicate formation. This change formalizes this request and allows minting users to specify the single format when minting sample barcodes. The duplicate format remains the default sample barcode format.


davereinhart (approved):
Looks good, and saves paper! :)

bencap:
Deployed.

PR #321, opened 2023-02-24
Add a Case Insensitive Option to Manifest Diff

bencap:
Adds an option `--ignore-case` to the manifest diff ID3C command. Running the command without this option retains the current behavior, which is to do a deep diff of manifest files while remaining sensitive to string case. Running the command with this option will perform the same diff, but will treat upper and lowercase strings the same.

### DIFF
--- a/lib/id3c/cli/command/manifest.py
+++ b/lib/id3c/cli/command/manifest.py
@@ -182,7 +182,7 @@ def parse_using_config(config_file):
 
     A YAML document can also contain a list of workbooks that share the same
     format:
-    
+
     \b
         ---
         workbooks:
@@ -198,7 +198,7 @@ def parse_using_config(config_file):
           accession_no: accession
           sample_origin: sample_origin
         ...
-    
+
     The sample_column entry specifies the name of the column
     containing the sample barcode. The collection_column entry specifies
     the name of the column containing the collection barcode. You must supply one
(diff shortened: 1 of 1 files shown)


bencap:
Should be merged alongside both https://github.com/seattleflu/backoffice/pull/362 and https://github.com/seattleflu/id3c-customizations/pull/357

lib/id3c/cli/command/manifest.py
```diff
@@ -403,9 +403,13 @@ def _parse(*,
     metavar = "<manifest-b.ndjson>",
     type = LocalOrRemoteFile("r"))
 
+@click.option("--ignore-case",
+    metavar = "<ignore-case>",
+    is_flag = True)
```
davereinhart, line 408: Could also add `default=False` here.

lib/id3c/cli/command/manifest.py
```diff
@@ -403,9 +403,13 @@ def _parse(*,
...
+
 @format_doc(PROVENANCE_KEY = PROVENANCE_KEY)
 
-def diff(manifest_a, manifest_b):
+def diff(manifest_a, manifest_b, ignore_case):
```
davereinhart, line 412: A default `ignore_case=False` could also be added here, but may not be necessary if this function is only being called via click command.

davereinhart (approved):

lib/id3c/cli/command/manifest.py
bencap: Ah true, let me add it in here just in case.

davereinhart (approved):
Looks good!

bencap:
Deployed

PR #322, opened 2023-03-10
Use logging to print promts for input

davereinhart:
Use of the input function with message string, and/or print function causes issues when geocoding function is used as part of a process that is using dump_ndjson and redirecting stdout to a file. To avoid having the input prompt text message included in stdout, it's being moved to a logging message.


bencap (approved):
Thanks for this Dave!

bencap:
Deployed.

PR #323, opened 2023-05-02
Synchronize REDCap Record Deletions with ID3C

bencap:
Adds a command `id3c redcap-sync delete` that pulls REDCap log data to determine any deleted REDCap records, then synchronizes those deletions with ID3C by deleting any encounters and related data associate with that REDCap record. Deletions don't necessarily need to be handled automatically, and running this new command with the `--post-to-slack` flag will surface an alert to ID3C maintainers in the #id3c-alerts channel if there are records that might require manual intervention.

Adds a new `logs` method to ID3C's notion of a REDCap project, which allows us to fetch log records from a REDCap project filtered by the APIs allowed fields.

Adds a new function decorator `with_redcap_project` that allows us to easily tag Click commands that should be provided the minimal command line options to build a REDCap project.


bencap:
It might be worth considering running this job on all past deletion events in core projects once this is merged. I ran it against most of our SCAN projects and didn't find any encounters in need to deletion (phew since I did that QC), but there are some HCT records (both Y3 and prior) that get pulled in.

lib/id3c/cli/command/redcap_sync.py
```diff
@@ -0,0 +1,315 @@
...
+    - `warehouse.location`
+    - `warehouse.sample`
+    - `warehouse.presence_absence`
+    """
+    LOG.info(f"Deleting all relational encounter data for encounter {encounter_id}")
+
+    # Take a snapshot of relational location data before we delete our encounter.
+    location_ids_associated_with_this_encounter = db.fetch_all(
+        f"SELECT encounter_id, location_id FROM warehouse.encounter_location WHERE encounter_id = {encounter_id}"
+    )
```
davereinhart, line 257: An alternative is to this and the other queries (that might simplify some logic below) is to select only the records that are safe to delete. e.g.
```
SELECT location_id FROM warehouse.encounter_location t1 WHERE encounter_id = {encounter_id}
AND location_id NOT IN (SELECT location_id FROM warehouse.encounter_location 
					   WHERE location_id = t1.location_id AND encounter_id <> {encounter_id})
```

lib/id3c/cli/command/redcap_sync.py
```diff
@@ -0,0 +1,327 @@
...
+    sample_ids_associated_with_this_encounter = db.fetch_all(
+        f"SELECT encounter_id, sample_id FROM warehouse.sample WHERE encounter_id = {encounter_id}"
+    )
+    for sample in sample_ids_associated_with_this_encounter:
+        deleted_pa_rows = delete_presence_absence_by_sample(db, sample.sample_id)
+        deleted_sample_rows = delete_sample(db, sample.sample_id)
+        LOG.debug(
+            f"Deleted {deleted_pa_rows} rows from `presence_absence` and {deleted_sample_rows} from `sample` associated "
+            + f"with `encounter_id` = {sample.encounter_id} and `sample_id` = {sample.sample_id}"
+        )
```
davereinhart, line 291: We probably want to avoid deleting sample records if they were also received from LIMS, but it might make sense to scrub the record and re-sync from LIMS to make sure we don't have anything from REDCap included.

lib/id3c/cli/command/redcap_sync.py
```diff
@@ -0,0 +1,327 @@
...
+
+    # All relational encounter data with FK constraints has now been deleted, so it's safe to delete the encounter.
+    deleted_encounter_rows = delete_encounter(db, encounter_id)
+    LOG.debug(
+        f"Deleted {deleted_encounter_rows} rows from `encounter` associated "
+        + f"with `encounter_id` = {encounter_id}"
+    )
```
davereinhart, line 298: This should always return 1, right? Could add a check for that here instead of the log message.

lib/id3c/cli/command/redcap_sync.py
```diff
@@ -0,0 +1,327 @@
...
+        with db.savepoint(f"individual {individual_id}"):
+            deleted_individual_rows = delete_individual(db, individual_id)
+            LOG.debug(
+                f"Deleted {deleted_individual_rows} rows from `individual` associated with "
+                + f"`encounter_id` = {encounter_id} and `individual_id` = {individual_id}"
+            )
+    except ForeignKeyViolation:
+        LOG.debug(
+            f"Couldn't delete `individual_id` = {individual_id}. This individual_id is still referenced by other encounters in ID3C."
+        )
```
davereinhart, line 312: Similar to locations, you could write a query for "individuals_unique_to_this_encounter" and only attempt to delete those.

lib/id3c/db/__init__.py
```diff
@@ -312,3 +312,81 @@ def upsert_sample(db: DatabaseSession,
...
+    Delete an individual by its `encounter_id`.
+    Will return the number of rows affected by this operation
+    """
+    with db.cursor() as cursor:
+        cursor.execute(f"DELETE FROM warehouse.individual WHERE individual_id = {individual_id}")
+
+    return cursor.rowcount
+
+
+def delete_location(db: DatabaseSession, location_id: int):
```
davereinhart, line 351: It might be nice to be able to pass in a list of `location_ids` and then delete them with a single query.

lib/id3c/db/__init__.py
```diff
@@ -312,3 +312,81 @@ def upsert_sample(db: DatabaseSession,
...
+    Will return the number of rows affected by this operation
+    """
+    with db.cursor() as cursor:
+        cursor.execute(f"DELETE FROM warehouse.encounter WHERE encounter_id = {encounter_id}")
+        assert cursor.rowcount < 2, f"More than one encounter associated with {encounter_id}"
+
+    return cursor.rowcount
+
+
+def delete_encounter_location(db: DatabaseSession, encounter_id: int):
```
davereinhart, line 329: Should this be renamed `delete_encounter_locations_by_encounter` for consistency (since we're not passing PK)?

lib/id3c/cli/command/redcap_sync.py
bencap: Yeah good thought. Going to follow up with this on Slack

bencap: Handling this by deleting records with only `note` and `coding` keys in their details. This indicates they have a provenance of REDCap and were not included on any LIMS / AQ sheet incident reports.

davereinhart (approved):

bencap:
Deployed.

PR #324, opened 2023-05-18
Only Send Slack Message if Deletions Exist

bencap:
Fix for #323 to only send slack messages if there are records which might need to be deleted.


bencap:
Deployed.

PR #325, opened 2023-07-12
Use psycopg2-binary, upgrade click to 8.x, pin click to <8.1.4

sallybg:
Upgrade click for compatibility with M2-compatible Fiona, and temporarily pin click to <8.1.4 to avoid mypy type check failures (pallets/click#2558).


davereinhart (approved):
Nice work! You may just want to add a minimum version for click "click >=8.0,<8.1.4".

davereinhart (approved):
Looks good! Thanks for getting these additional packages pinned to match what's been done in production. We can deal with unpinning those separately.

PR #327, opened 2023-07-25
Create New Viral Evolution Identifier Set

bencap:
Creates a new identifier set for the upcoming viral evolution study modeled after the Cascadia identifier set layout. Can be merged alongside seattleflu/id3c-customizations#367


bencap:
We can run `pipenv run id3c identifier set create collections-viral-evolution collection 'Collection tubes from the Viral Evolution Study'` against ID3C once we are ready to make this change.

bencap:
Viral Evo / SILVER study no longer planned for BBI.

PR #328, opened 2023-07-27
Adding row level security to sample and genomics tables

davereinhart:
Adds row-level security to the following tables in warehouse schema:
- sample
- genomic_sequence
- consensus_genome
- sequence_read_set

An `access_role` column has been added to each table to store an optional role name that the data is accessible to. The default behavior (with access_role=NULL) will remain unchanged, with no row-level security enforced.

Note: I did not add verify scripts for each table, as some of the them have foreign key constraints that can cause those fail if things aren't populated yet. I may revisit this if I have more time, but for now, testing sqitch deploy and revert should be sufficient.


bencap (commented):
This looks good Dave and is working for me on my local (upgraded to 15.3). Unless I'm mistaken on the SSH tunnel, we will also need to update `sqitch.conf` to point to localhost and access ID3C via SSH to deploy this via Sqitch.

Just making sure I understand the Metabase side: We would have add a new Metabase user that has access to the marked Cascadia data (the default metabase user would not have that access). We'd then create a new Metabase database connection using that user to hit ID3C. 
Then Metabase users with Cascadia privileges would be inside a Cascadia group, which would then be the only group with access to that new Cascadia database connection.

bencap (approved):
Looks good to me, we will need to ensure that no additional users are members of the `dumper` role before deploying to ensure that no one retains undesirable access to the receiving tables.

davereinhart:
Deployed via sqitch.

PR #329, opened 2023-08-24
Sample API update

davereinhart:
Updates to warehouse/sample POST endpoint to allow `access_role` value to be passed. A database constraint will ensure this value is set correctly based on the `sample_origin` which can be used to identify data from a project that should have restricted access.

### DIFF
--- a/lib/id3c/db/__init__.py
+++ b/lib/id3c/db/__init__.py
@@ -186,7 +186,8 @@ def upsert_sample(db: DatabaseSession,
                   collection_identifier: Optional[str],
                   collection_date: Optional[str],
                   encounter_id: Optional[int],
-                  additional_details: dict) -> Tuple[Any, str]:
+                  additional_details: dict,
+                  access_role: Optional[str] = None) -> Tuple[Any, str]:
     """
     Upsert sample by its *identifier* and/or *collection_identifier*.
     An existing sample has its *identifier*, *collection_identifier*,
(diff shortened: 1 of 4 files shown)


lib/id3c/db/__init__.py
```diff
@@ -200,13 +201,16 @@ def upsert_sample(db: DatabaseSession,
         "collection_date": collection_date,
         "encounter_id": encounter_id,
         "additional_details": Json(additional_details) if additional_details else None,
+        "additional_details_without_prov": Json({k: additional_details[k] for k in additional_details if k != '_provenance'}) if additional_details else None,
+        "access_role": access_role,
     }
 
+    LOG.info(data)
```
bencap, line 208: Did you mean to leave this?

bencap (approved):
Looks good Dave. Thanks for the provenance changes too btw, although I was curious: Would there not be cases where there might be a valid provenance change that we would now miss out on? It might be worth the extra logic to look exclusively at the timestamp.

davereinhart:
> Looks good Dave. Thanks for the provenance changes too btw, although I was curious: Would there not be cases where there might be a valid provenance change that we would now miss out on? It might be worth the extra logic to look exclusively at the timestamp.

My thought was that if the only thing changing is the provenance, then the record's actual data is not changing, so the update isn't needed and the new provenance isn't really relevant.

bencap:
Ah ok, that seems fair wrt provenance- Thanks!

davereinhart:
Deployed.

PR #331, opened 2023-11-28
Add option to process tract without address in FHIR ETL

sallybg:
Previously, the process_location function, called as part of the etl fhir function, assumed that the encounter location was an address with a tract as a parent location. Sample metadata received as part of KP2023 do not contain addresses, only census tracts, so process_location now handles encounter locations that are just tracts with no address and no parent location.

### DIFF
--- a/lib/id3c/cli/command/etl/fhir.py
+++ b/lib/id3c/cli/command/etl/fhir.py
@@ -798,15 +798,17 @@ def process_locations(db: DatabaseSession, encounter_id: int, encounter: Encount
 def process_location(db: DatabaseSession, encounter_id: int, location: Location):
     """
     Process a FHIR *location* and attach it to an *encounter_id*.
+    First looks for a parent location's tract hierarchy;
+    if no parent location is available, assumes that the location entry itself
+    is a tract and looks for a tract hierarchy.
     """
-    # XXX FIXME: This function assumes we always have an address, and never
-    # just a tract.  It also assumes the scales tract and address and their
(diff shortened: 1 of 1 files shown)


lib/id3c/cli/command/etl/fhir.py
```diff
@@ -844,12 +846,19 @@ def process_address(location: Location, tract_hierarchy: str) -> Optional[Any]:
...
 
     except AttributeError:
         LOG.info(f"No parent location found for Location with code «{code}», " + \
             f"relation «{relation}».")
-        tract_hierarchy = None
+        try: # handle encounter locations that are only tracts
+            tract = get_tract(location)
+            tract_hierarchy = tract.hierarchy
+            tract_id = tract.id
+        except:
```
bencap: We should generally try to avoid blanket `except` statements. Are we trying to catch the `AssertionError` here in the case that the tract is unknown? Maybe we are also trying to catch the `AttributeError` in the case of `tract.hierarchy` not being defined?

lib/id3c/cli/command/etl/fhir.py
```diff
@@ -860,7 +869,8 @@ def process_address(location: Location, tract_hierarchy: str) -> Optional[Any]:
     upsert_encounter_location(db,
         encounter_id    = encounter_id,
         relation        = relation,
-        location_id     = address.id)
+        location_id     = address.id if address else tract_id)
```
bencap: You could just use `tract.id` directly here instead of creating a new `tract_id` variable.

